PR #1, opened 2019-09-06
Add a link to the projects page to the nav bar

wbthomason:
I'd find it convenient to link to the projects page from the global site nav. This PR adds such a link.


sampsyo:
Seems like a good idea; thanks!

PR #28, opened 2019-09-24
Support multiple authors for a post

rachitnigam:


sampsyo:
Thanks! I changed the TOML format to use a sequence of tables instead of a single table so that the ordering is deterministic.

Issue #25, opened 2019-09-12
Project 1 Proposal: Add record types to Bril


liuhenry4428:
**What will you do?**
We will add support for mutable record types. A record will have a set of named fields with types, and those types can include record types. We are also thinking about adding support for constant record types (by adding a const keyword) to designate a record as immutable. @sampsyo what do you think about this? 

**How will you do it?**
We will edit the Bril interpreter, and also bril2json for tooling.

**How will you empirically measure success?**
We will consider this project successful if we add working record types that integrate completely with the existing types, such that any valid type can be used in a record.

Working with Katy Voor (kev29)

sampsyo:
Sounds good! I'd probably just pick mutable or immutable and stick with it—that's probably enough for this project.

I think a big choice you'll want to make is whether record types will need to be declared up front and then used by name or just listed inline, without a central declaration. In the former case, you can decide whether the type system should be nominal or structural.

sampsyo:
Closed in #31.

PR #31, opened 2019-09-25
Record Types!

liuhenry4428:


kavoor:
closes #25

content/blog/2019-09-25-RecordTypes.md
```diff
@@ -0,0 +1,131 @@
...
+v0: int = const 3;
+v1: bool = const false;
+Milo: Dog = record {age: v0; isAsleep: v1}
+v2: int = const 4120;
+AndrewMyers: Person = record {class: v2; dog: Milo}
+```
+
+When defining a nested record, it is required to first define the nested record and assign it to a variable. Then, you can define the outer record with the field initialized to the variable holding the nested record. When type-checking these nested records, we need to lookup the type of the outer record from our type environment, and step through the fields, one by one, comparing the signature with the type returned from the lookup of the initializing variable. 
+
+Consider the case for checking the variable initializing a nested record, i.e. `Milo`. With nominal typing, if we expected a type, Dog, and we looked up the variable to have type ‘Dog’, we know this must have been previously typechecked when it was defined and added to our environment. Therefore it still must typecheck, due to immutability. This sort of _shallow type-checking_ falls out of the immutability property.
```
kavoor: I think this should say "... falls out of the nominal subtyping".

content/blog/2019-09-25-RecordTypes.md
```diff
@@ -0,0 +1,131 @@
...
+
+### *With* Syntax
+While immutable data structures allow you to more easily reason about how values flow through your program, these currently immutable records are cumbersome to change. 
+For example, to change one field, you must recreate the entire record. While we do want a new record to be created, copying variables between two records is overly tedious. 
+As such, we decided to implement a with syntax, similar to OCaml in which the user just needs to specify a record name as the base, and then the fields desired to be changed, along with the new value. This maintains immutability, without so much tedium.
+
+We use the following syntax:
+
+    <new record>: <record type> = 
+        <old record> with {<field1 name>: <field1 value>; <field2 name>: <field2 value> … };
```
sampsyo: As above—you don’t have to invent new syntax here; you could instead try to make it fit as a “normal” instruction.

content/blog/2019-09-25-RecordTypes.md
```diff
@@ -0,0 +1,131 @@
...
+
+Where:
+`<new record>` is the name of the new record
+`<record type>` is the same type as `<old record>`
+`with` is a new keyword
+`<field# name>` must match with a field in `<type>`
+`<field# value>` must be a variable with a type that matches its field name.
```
sampsyo: Commas and conjunctions, as above.

content/blog/2019-09-25-RecordTypes.md
```diff
@@ -0,0 +1,131 @@
...
+Within the braces, the user may specify 0 to n field name and value pairs, where n is the total number of fields in the record type.
+
+This syntax was designed to have a similar format as record instantiation. 
+
+### Immutability
+Immutable data structures are easier to reason about when developing compiler optimizations. However, forcing developers to recreate the entire record every time they need to change a value poses usability challenges. We decided on sticking to immutability in part for optimizations' sake. The ability to make guarantees that a value will not change lends itself well to constant-folding, even after a function call that takes in your record as an argument.
```
sampsyo: Because this is a top-level design concern, it might be wise to put it at the top of your writeup. Also, I’d argue that usability is not a paramount concern for a compiler IR—but the ability to compile higher-level languages into it is.

content/blog/2019-09-25-RecordTypes.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Evaluation
+We set out to implement record types and we successfully implemented immutable nominal record types. Record types do not have any restrictions except that two fields cannot have the same name and that a record type cannot be recursive. Besides that, everything is allowed. Record types play nicely with existing types and new syntactical additions follow previous precedents set by Bril or otherwise follow precedents set by other languages that have record types. Record declaration, instantiation, and access are all (we think) intuitive and straightforward, and this means our record types provide good value as an addition to the Bril language. 
+
+We found that creating new records was a tedious process if the record was large, so we implemented *with* statements in addition to the features mentioned above for situations where one wanted to duplicate a record with a few changes. It should be noted that it is bad form to use a with statement with no fields because that would be identical to referencing the old record with `… = id oldRecordName.`
```
sampsyo:
```suggestion
We found that creating new records was a tedious process if the record was large, so we implemented *with* statements in addition to the features mentioned above for situations where one wanted to duplicate a record with a few changes. It should be noted that it is bad form to use a with statement with no fields because that would be identical to referencing the old record with `… = id oldRecordName`.
```

content/blog/2019-09-25-RecordTypes.md
```diff
@@ -0,0 +1,131 @@
...
+We were successful in this aspect as creating a new record from an existing one but changing one field is nearly the same amount of code as mutating directly.
+```
+type Person = {age: int; isAsleep: bool}
+v0: int = const 21;
+v1: bool = const false;
+Henry: Person = record {age: v0; isAsleep: v1}
+v2: bool = const true;
+AwakeHenry: Person = Henry with {isAsleep: v2}
+```
+Possible Mutable Record Syntax:
```
sampsyo: Could use more discussion. (Also no need for title case.)

content/blog/2019-09-25-RecordTypes.md
```diff
@@ -0,0 +1,131 @@
...
+We found that creating new records was a tedious process if the record was large, so we implemented *with* statements in addition to the features mentioned above for situations where one wanted to duplicate a record with a few changes. It should be noted that it is bad form to use a with statement with no fields because that would be identical to referencing the old record with `… = id oldRecordName.`
+
+We were successful in this aspect as creating a new record from an existing one but changing one field is nearly the same amount of code as mutating directly.
+```
+type Person = {age: int; isAsleep: bool}
+v0: int = const 21;
+v1: bool = const false;
+Henry: Person = record {age: v0; isAsleep: v1}
+v2: bool = const true;
+AwakeHenry: Person = Henry with {isAsleep: v2}
```
sampsyo: Missing semicolons.

sampsyo (changes requested):
Nice work! Enclosed are some low-level suggestions, and here are some high-level ones:

- What exactly did you implement? No need to go into detail, but say what you modified---the spec, interpreter, parser, pretty-printer, etc.
- When you discuss the extensions, giving the text format as examples is fine, but remember that the canonical form of Bril is the JSON AST. So please also discuss what that looks like. (I'd argue that the concrete syntax you use in the text format is less important than the underlying abstract syntax.) 
- Your evaluation should be empirical, so that section should discuss how you can tell your implementation does what it is supposed to do.

sampsyo (approved):
Fantastic; this looks great!!

Issue #11, opened 2019-09-11
Project 1 Proposal: Compiling Bril IR into LLVM 


Hecmay:
**What will you do?**
Compiling Bril IR into LLVM 

**How will you do it?**
Write a parser in C and use LLVM IRBuilder to build the JIT compilation flow. Our system should be flexible enough to support future features such as floating points and other arithmetic operations.

**How will you empirically measure success?**
Develop some test cases. Run through the LLVM and assert final results are the same with Typescript.

**Team members**
@seanlatias
@zhouyuan1119
@Hecmay

sampsyo:
Cool! Using IRBuilder does seem like a good way to go about it. See also #4 and #6, where there are other people also interested in generating LLVM. I provided some advice about how to approach it there. And you might consider coordinating with other people trying to things along similar lines.

> Write a parser in C

Please don't bother writing your own parser. You can use an off-the-shelf JSON parser to load in Bril programs. In fact, [LLVM has one built in](http://llvm.org/doxygen/JSON_8h.html)!

PR #33, opened 2019-09-25
Project 1: LLVM JIT Compiler for Bril

hecmay:
Resolves proposal #11 with @zhouyuan1119 @seanlatias

This project aims to extend the reachability of Bril IR to different backend devices by compiling Bril programs to LLVM IR. The blog post consists of three parts: 
1. Methodology: brief introduction to the overall workflow 
2. Implementation details: how basic blocks and instruction are built using LLVM
3. Evaluation: analysis of the generated result and comparison of performance

Our program is available in [the forked Bril repo](https://github.com/seanlatias/bril/tree/master/codegen-llvm) under `codegen-llvm` folder.

### DIFF
--- a/content/blog/2019-09-25-llvm-backend-for-bril.md
+++ b/content/blog/2019-09-25-llvm-backend-for-bril.md
@@ -2,14 +2,17 @@
 +++
 title = "Project Report 1: LLVM IR transformer"
 extra.bio = """
-
+In this project, we aim to extend the reachability of Bril IR to different backend devices by compiling Bril programs to LLVM IR. We execute the generated LLVM IR via LLVM execution engine to verify its functional correctness. Finally, we compare the runtime between LLVM JIT compilation and Bril interpreter
 """
 [[extra.authors]]
 name = "Shaojie Xiang"
+link = "https://github.com/Hecmay"
 [[extra.authors]]
 name = "Yi-Hsiang (Sean) Lai"
+link = "https://github.com/seanlatias"
 [[extra.authors]]
 name = "Yuan Zhou"
+link = "https://github.com/zhouyuan1119"
 +++
 
 ## Project Report 1: LLVM JIT Compiler for Bril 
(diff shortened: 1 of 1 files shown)


sampsyo (changes requested):
Looks awesome! Nice work. Here are some pre-publication suggestions.

content/blog/2019-09-25-llvm-backend-for-bril.md
```diff
@@ -0,0 +1,127 @@
...
+  b2:                                               ; preds = %b1, %0
+    br label %end
+  end:                                              ; preds = %b2
+    %16 = load i64, i64* %1
+    %17 = call i64 @printf(i8* getelementptr inbounds ([4 x i8], [4 x i8]* @1, i32 0, i32 0), i64 %16)
+    ret i32 0
+  }
+```
+
+After verifying the correctness of the code generator, we also compare the performance of LLVM simulation and Bril Interpreter. The performance is measured with profiling tool in Linux and C++. We run the same program for 10 times and take the average runtime. For the test program with a regular for loop iteratively computing one multiply operation for 1 billion times, the LLVM interpreter run about 10 times faster than Bril interpreter. The average runtime is 0.47 seconds and 0.05 seconds for Bril and LLVM interpreter respectively. The LLVM execution engine achieves approximately 1000x speedup over the Bril interpreter without optimizing the loop inside. We can expect higher speedup if the loop is optimized away.
```
sampsyo:
```suggestion
After verifying the correctness of the code generator, we also compare the performance of LLVM simulation and the Bril interpreter. The performance is measured with profiling tool in Linux and C++. We run the same program for 10 times and take the average runtime. For the test program with a regular for loop iteratively computing one multiply operation for 1 billion times, the LLVM interpreter runs about 10 times faster than the Bril interpreter. The average runtime is 0.47 seconds and 0.05 seconds for Bril and LLVM interpreter respectively. The LLVM execution engine achieves approximately 1000x speedup over the Bril interpreter without optimizing the loop inside. We can expect higher speedup if the loop is optimized away. 
```

PR #34, opened 2019-09-25
Project 1 report: Static type checking

### LINKED ISSUE #3: Project 1 Proposal: Static type checking for Bril
**What will you do?**
We will implement a static type checker for Bril. It can be used to find type errors and uses of undefined variables.

**How will you do it?**
We will first go through the Bril language to find all types existing and the corresponding typing rules. Then we will implement the type checker in ts based on the rules designed.

There are two general cases for type errors. Either no type is defined before use, or the use case violates the predefined type. Thus, we test the type checker also in two folds. A bunch of test files would to be written covering non-predefined types, violation of typing rules for all types, and well-typed code as well.

Project with @tissue3 and @neiladit

neiladit:
Resolves #3


sampsyo:
I've removed the code in your post that included MathJax and switched your post over to use the built-in KaTeX stuff that I just added for all posts. Can y'all please check that everything still looks OK?

cc: @neiladit, @ElaineJiang, @tissue3

sampsyo (changes requested):
Hello! This is a good start, but this blog post could be a lot better with two main changes:

- Reconsider the formalism. I'm actually not sure the typing rules add too much to the discussion, and there are notational problems that make them somewhat imprecise. Maybe you can instead just enumerate the types of each instruction kind with a table or something, or perhaps just give a couple of rules. But if you keep the typing rules, you'll need to make them look more like typing rules—using a proper typing context and notation like `\Gamma \vdash e : \tau`. Which may be harder than it looks.
- Give more detail on your implementation and the challenges. You briefly allude to giving line-number errors, for example. Describe how that actually works and perhaps even show an example of the output of your tool. And be more illustrative about programs that you decided were "valid" or "invalid," especially when the right choice was not obvious.

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,108 @@
...
+$$ \frac{<n,\sigma>\Downarrow int}{<var:int = const\ n ,\sigma>\Downarrow \sigma'}\\ \frac{a,\sigma \Downarrow_a var_{int}}{<id\ a,\sigma>\Downarrow \sigma'}\\ \frac{b,\sigma \Downarrow_b var_{bool}}{<id\ b,\sigma>\Downarrow \sigma'}\\ \frac{}{<ret,\sigma>\Downarrow End}\\ \frac{\forall v_i \in \sigma[var]}{<print\ v_1,v_2,\dots,v_i,\dots,v_n,\sigma> \Downarrow \sigma'} $$
+
+
+
+### Implementation
+
+Once we had the rules setup for all types of instructions, we check each instruction against them. We have a first pass in the algorithm which collects a list of labels and ensures that each label is unique. This also helps to check for control instructions if the label in the argument is valid.  In the second pass, we go over each instruction and check for various type errors - 
+
+1. For const instructions, check if assigned value is of the same type as the destiantion type. We also check if the variable has already been assigned to a different type in that context, in which case reassignment would be an error.
+2. For arithmetic and logical operations, we check if the two source operands are valid and are of the same type as the destination type.
```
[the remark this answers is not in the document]
neiladit: I am a bit confused on this one. If the assignment comes after the use as an argument, isn't the variable undefined (which should be an error)?

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+
+The error raised by the type checker outputs the line at which we find the first operation breaking the type check rule and mentions the type of error found.
+
+
+### Implementation
+
+We have a first pass in the algorithm which collects a list of labels and ensures that each label is unique. If there are multiple labels with the same name, we throw an error at this point. This also helps us create a set of label names which is used to check for valid strings (of labels) during the second pass for control flow operations.  In the second pass, we go over each instruction and check for various type errors. Some fundamental checks in this process are:
+
+1. Invalid instructions- This ensures that all the arguments and destination (if applicable) of an instruction are available, which is given by the length of instruction.
```
sampsyo: Use a colon, not a hyphen, after the introductory phrase in each list item.

sampsyo: > length of instruction

Do you mean the number of arguments?

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+2. Argument and destination type- For various operations we check if arguments and destination variables have the correct type. Something like `bool d = a + b` where `a,b` are integers would raise an error for the destination variable.
```
sampsyo: `bool d = a + b` is not Bril syntax—do you mean `d: bool = add a b`?

sampsyo: Use "`a` and `b`" instead of "`a,b`" (the comma isn't syntax).

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+3. Redefined variables- We check if the destination variable has already been assigned to a different type in that context. We do allow const instructions on the same type as it can be treated as an assignment operation rather than a definition. So `int a =2; bool a = true` is not allowed but `int a =2; int a =5` is allowed. We do this by keeping a set of variables of each type (int and bool) defined in the function. This helps while checking existing definitions and possible redefinition errors.
```
sampsyo: > We do allow const instructions on the same type as it can be treated as an assignment operation rather than a definition.

Is this something special about "const," or is it true of all instructions? It seems to be implying the former, but I think it's the latter.

sampsyo: Again, it would be clearer if you used real Bril syntax here instead of this alternative C-like syntax.

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+4. Undefined variables- We check if the arguments to the instruction have defined variables using the set of variables mentioned before.
```
sampsyo: What if the variable is defined "later" in program order, but it's guaranteed to be defined because of the way jumps work? Is that an error? (It's fine if so, but you just have to make it clear that this is a limitation.)

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+4. Undefined variables- We check if the arguments to the instruction have defined variables using the set of variables mentioned before. 
+
+
+
+## Hardest parts during the implementation
+
+The type checker implementation, though straightforward, had a few challenges.
+
+1. The typing rules need to be designed carefully for each operation, ranging from arithmetic, boolean to control instructions. Take the branch instruction as an example, say we have "br b left right". The type checker needs to go through all the existing boolean variables to ensure that b is predefined and left and right are valid labels in the code snippet. Also, though not currently supported in Bril, the instructions could be nested. Thus, it is important to maintain modularity of arithmetic/boolean and control flow checking so that future updates could be made easily to support recursive checking.
```
sampsyo: > ranging from arithmetic, boolean to control instructions.

That's not quite grammatical—maybe you mean "ranting from arithmetic and boolean to control instructions."

sampsyo: > Take the branch instruction as an example, say we have "br b left right".

This is a comma splice—consider a colon instead of a comma. And use Markdown code backticks instead of quotes.

sampsyo: What do you mean by "nested" instructions? I'm not quite sure I get it—to be clear, something like `add a (mul b c)` as an alternative to `x: int = mul b c; y: int = add a x` will never be supported in an instruction-based IR.

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+2. To ensure that the type checker works properly with labels, a first pass of the labels is designed. Through the first pass, all existing labels are saved and no duplicate ones are allowed to avoid conflicts. Though this almost doubles the type checking overhead, it further ensures the correctness of the program by static analysis.
+3. Keeping track of the line number and returning proper error message when encountered with an type checking error. When scanning through the code line by line after the first pass, a line number is maintained during type checking to help the programmer debug. In order to maintain the line number with blank lines and comments, we take in the original code snippet and ignore these lines while updating it during type checking.
```
sampsyo: I'm still a little confused about how you did this—in particular, are you working on the Bril text representation? Or is this working on the JSON representation? If it's on the text representation, did you write your own parser or hack up the existing one?

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+
+## Evaluation and results
+
+We wrote a set of benchmarking programs to test our implementation for various types of rules defined in the design section above. 
+
+The general test cases are classified into two sub-directories: [should-fail](https://github.com/tissue3/bril/tree/master/test/type-check/should-fail) and [should-pass](https://github.com/tissue3/bril/tree/master/test/type-check/should-pass). It has expected output named as "\*.out" corresponding to the input file "\*.bril". User can simply run all test cases in a directory by running turnt directory/\*.bril. For example, to run should-fail benchmark, one can just run turnt test/type-check/should-pass/\*.bril at the main directory of bril.
```
sampsyo: Consider using Markdown code backticks for filenames.

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+
+Because should-pass cases are trivial, where we enumerate list existing operations, in the following table we only list tests cases that our type checker will report error message. The second column of the table aims to help one understand why an error would be reported by our type checker.
```
sampsyo: > report error message

report *an* error message

sampsyo: > where we enumerate list existing operations

Maybe either "enumerate" *or* "list," but not both?

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+| Instruction | Type Checking Rule |Testing Code Snippet  |
+| ----------- | :-----------: | :-------------: |
+| Conflict Definition       |   Variables cannot be redefined with a different type.   |      v: int = 5;<br/> v :bool = true;      |
+| Arithmetic         |          Adding an integer and a boolean type.           |                a: int = 4;<br/>b: bool = true;<br/>c: int = add a b;               |
+| Boolean         |      Cannot assign the output of boolean to an integer.      |    b: bool = true;<br/> a:int = not b;    |
+| Const       |    Cannot assign an integer const to a bool variable.    |            a: int =  const true;             |
+| Cond Branch |    Only takes bool variable as input (and 2 labels).     |     a: int = 1;<br/> br a here there;<br/> here: print a;<br/>there: jmp here; |
+
+
+Finally, though not required, we also implemented other checking passes as long as an input stream is parsable by bril2json, an interpret for bril. That includes type existence checking, number checking, argument existence checking, label existence checking and label repentance checking.
```
sampsyo: > bril2json, an interpret for bril.

To be clear, `bril2json` is not an interpreter. It's just a parser.

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+
+| Instruction |               Type Checking Rule               |          Testing Code Snippet          |
+| ----------- | :-----------------------------------------------------: | :-----------------------------------: |
+| Type existence       | The destination type is undefined.          |  boolean: v0= const true; |
+| Argument Number       | The expected argument is 2 but only 1 is given.          |  int: v0= const 5;<br/> int: v1 = add v0;  |
+| Argument Existence | The argument is never defined. | int: v1 = add v0; |
+| Label Existence      | Label argument in control operation not present in code. |            jmp its_a_trap; (Consider this as is a full program)     |
+| Repeated Label |      A label should be unique and not be repeated.       | jmp label;<br/> label: a:int=1;<br/> label: a:int =2 |
+
+By and large, we have implemented the checker satisfying all of our defined behaviors. But we don't know if that's exhaustive for all possible errors (not necessarily type error). We would be very happy if someone comes up with more cases and reaches out us by mail or github issues.
```
sampsyo:
```suggestion
By and large, we have implemented the checker satisfying all of our defined behaviors. But we don't know if that's exhaustive for all possible errors (not necessarily type errors). We would be very happy if someone comes up with more cases and reaches out us by mail or GitHub issues.
```

content/blog/2019-09-25-type_checker.md
sampsyo: I'm not quite sure what this paragraph is saying. What's the difference between the original version of the type checker described above and this new one? Do they not work on the same input format?

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+
+| Instruction |               Type Checking Rule               |          Testing Code Snippet          |
+| ----------- | :-----------------------------------------------------: | :-----------------------------------: |
+| Type existence       | The destination type is undefined.          |  boolean: v0= const true; |
+| Argument Number       | The expected argument is 2 but only 1 is given.          |  int: v0= const 5;<br/> int: v1 = add v0;  |
+| Argument Existence | The argument is never defined. | int: v1 = add v0; |
+| Label Existence      | Label argument in control operation not present in code. |            jmp its_a_trap; (Consider this as is a full program)     |
+| Repeated Label |      A label should be unique and not be repeated.       | jmp label;<br/> label: a:int=1;<br/> label: a:int =2 |
```
sampsyo: Seems like these code examples have some syntax backward. For example, it's not `boolean: v0= const true;`. Maybe you mean `v0: bool = const true;`.

sampsyo (changes requested):
Good work on the improvements! Things are headed in the right direction. I see you've deleted the inference rules for the type system, which is fine. I do still think you have some work to do before this is a publishable writeup, however—there are two main things to work on:

- Try to be as clear as possible in the "Design" section about what is allowed and disallowed by your type checker. I had trouble understanding precisely what was and was not allowed. I think some examples of "good" and "bad" programs would help—not the obvious ones that _everyone_ would agree are good or bad, but more subtle ones that reveal the precise decisions you made.
- Instead of using pseudo-syntax for a C-like language in your writeup, I think it would be much easier to understand if you use "real" Bril syntax throughout.

content/blog/2019-09-25-type_checker.md
tissue3: This is not syntax backward. Because boolean is an undefined type, our type checker will report that the destination variable type is undefined.

sampsyo: OK but still, don't you mean to have the type go _after_ the variable name? (Or is the point that `v0` is not a type?)

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+
+## Design and Implementation
+
+### Design
+
+Bril currently supports 2 types - **int** and **bool** which makes type checking relatively easy. Our type checker is defined such that an arithmetic operation  `a = b+c` would raise an error if either the source or destination arguments are *not* int. Similarly, boolean operations only accept all bool arguments and comparison operations have integer arguments but a boolean destination. These type definitions have been nicely defined [here](https://capra.cs.cornell.edu/bril/langref.html). Currently Bril only supports operations on variables and not constants which implies `a = b+5` is an invalid operation. During type checking, we also ensure that the variable has been defined in the function before. Conversely, we make sure that there isn't a redefinition of the variable. For control flow operation, we ensure that labels are present in the code and are uniquely defined. This allows us to treat label strings as a separate type, invisible to the user.
```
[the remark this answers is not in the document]
neiladit: > Similarly, it's not quite clear what you mean by "invalid operation" in this:
> 
> > Currently Bril only supports operations on variables and not constants which implies `a = b+5` is an invalid operation.
> 
> Because it's not even clear how you'd write that down in Bril abstract syntax. In particular, this makes it hard to understand whether "invalid operation" means "this is something our type checker must catch" or "this is something that can't even be expressed, so it's something we don't have to worry about."

I thought this is for detecting an operation like:
`a: int = add b 5`
Do you think `5` will act as an undefined variable (and hence won't be detected as a constant integer)?

content/blog/2019-09-25-type_checker.md
ElaineJiang: Changed to "including"

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+
+## Evaluation and results
+
+We wrote a set of benchmarking programs to test our implementation for various types of rules defined in the design section above.
+The general test cases are classified into two sub-directories: [should-fail](https://github.com/tissue3/bril/tree/master/test/type-check/should-fail) and [should-pass](https://github.com/tissue3/bril/tree/master/test/type-check/should-pass). It has expected output named as `*.out` corresponding to the input file `*.bril`. User can simply run all test cases in a directory by running `turnt directory/*.bril`. For example, to run should-fail benchmark, one can just run `turnt test/type-check/should-pass/*.bril` at the main directory of bril.
```
sampsyo:
```suggestion
The general test cases are classified into two sub-directories: [should-fail](https://github.com/tissue3/bril/tree/master/test/type-check/should-fail) and [should-pass](https://github.com/tissue3/bril/tree/master/test/type-check/should-pass). It has expected output named as `*.out` corresponding to the input file `*.bril`. The user can simply run all test cases in a directory by running `turnt directory/*.bril`. For example, to run should-fail benchmark, one can just run `turnt test/type-check/should-pass/*.bril` at the main directory of bril.
```

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,100 @@
...
+| Conflict Definition  |   Variables cannot be redefined with a different type.   |      v: int = 5;<br/> v :bool = true;      |
+| Arithmetic         |          Adding an integer and a boolean type.           |                a: int = 4;<br/>b: bool = true;<br/>c: int = add a b;               |
+| Boolean         |      Cannot assign the output of boolean to an integer.      |    b: bool = true;<br/> a:int = not b;    |
+| Const       |    Cannot assign an integer const to a bool variable.    |            a: int =  const true;             |
+| Cond Branch |    Only takes bool variable as input (and 2 labels).     |     a: int = 1;<br/> br a here there;<br/> here: print a;<br/>there: jmp here; |
+
+Finally, though not required, we also implemented other checking passes as long as an input stream is parsable by bril2json. For example, `v0: boolean =const true;` is a legal statement for bril2json, but boolean is not an existing type in bril, so our type checker will report the error. Such kind of checking pass includes type existence checking, number checking, argument existence checking, label existence checking and label repentance checking.
```
sampsyo:
```suggestion
Finally, though not required, we also implemented other checking passes as long as an input stream is parsable by bril2json. For example, `v0: boolean =const true;` is a legal statement for bril2json, but boolean is not an existing type in bril, so our type checker will report the error. This kind of checking pass includes type existence checking, number checking, argument existence checking, label existence checking and label repentance checking.
```

sampsyo: What are "number checking" and "label repentance checking"?

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,100 @@
...
+
+| Instruction |               Type Checking Rule               |          Testing Code Snippet          |
+| ----------- | :-----------------------------------------------------: | :-----------------------------------: |
+| Type existence       | The destination type is undefined.          | v0: boolean = const true; |
+| Argument Number       | The expected argument is 2 but only 1 is given.          |  v0: int = const 5;<br/> v1: int = add v0;  |
+| Argument Existence | The argument is never defined. | v1: int = add v0; |
+| Label Existence      | Label argument in control operation not present in code. |            jmp its_a_trap; (Consider this as is a full program)     |
+| Repeated Label |      A label should be unique and not be repeated.       | jmp label;<br/> label: a:int=1;<br/> label: a:int =2 |
+
+By and large, we have implemented the checker satisfying all of our defined behaviors. But we don't know if that's exhaustive for all possible errors (not necessarily type errors). We would be very happy if someone comes up with more cases and reaches out us by mail or GitHub issues.
```
sampsyo:
```suggestion
By and large, we have implemented the checker satisfying all of our defined behaviors. But we don't know if that's exhaustive for all possible errors (not necessarily type errors). We would be very happy if someone comes up with more cases and reaches out to us by mail or GitHub issues.
```

content/blog/2019-09-25-type_checker.md
sampsyo: Again, no need to fix this now, but I don't quite understand why there are two categories of tests. As far as I can tell from your implementation, all of these depend on being parseable with my Lark parser ("bril2json"), so it doesn't seem right to say that these ones are the ones that are relevant in that case.

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,100 @@
...
+2. To ensure that the type checker works properly with labels, a first pass of the labels is designed. Through the first pass, all existing labels are saved and no duplicate ones are allowed to avoid conflicts. Though this almost doubles the type checking overhead, it further ensures the correctness of the program by static analysis.
+3. Keeping track of the line number and returning proper error message when encountered with an type checking error. To track the line number, the type checker works on the Bril text representation and is developed with its own parser. When scanning through the code line by line after the first pass, a line number is maintained during type checking to help the programmer debug. In order to maintain the line number with blank lines and comments, we take in the original code snippet and ignore these lines while updating it during type checking.
```
sampsyo: > developed with its own parser.

This seems to be wrong. I don't see a parser in your code:
https://github.com/sampsyo/bril/pull/9/files#diff-1b34e9c92d3d0993e1e975a194da566a

It's reusing the `briltxt` parser. But there's some funky logic in there that attempts to keep track of which line in the original text representation corresponds to which parsed Bril instruction.

content/blog/2019-09-25-type_checker.md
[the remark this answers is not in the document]
sampsyo: That's up to you to determine. But 5 clearly won't parse as a variable.

content/blog/2019-09-25-type_checker.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+
+## Evaluation and results
+
+We wrote a set of benchmarking programs to test our implementation for various types of rules defined in the design section above.
+The general test cases are classified into two sub-directories: [should-fail](https://github.com/tissue3/bril/tree/master/test/type-check/should-fail) and [should-pass](https://github.com/tissue3/bril/tree/master/test/type-check/should-pass). It has expected output named as `*.out` corresponding to the input file `*.bril`. User can simply run all test cases in a directory by running `turnt directory/*.bril`. For example, to run should-fail benchmark, one can just run `turnt test/type-check/should-pass/*.bril` at the main directory of bril.
+Because should-pass cases are trivial, where we enumerate list existing operations, in the following table we only list tests cases that our type checker will report an error message. The second column of the table aims to help one understand why an error would be reported by our type checker.
```
sampsyo: Looks like you missed this comment about "enumerate list"? https://github.com/sampsyo/cs6120/pull/34#discussion_r331818904

sampsyo (changes requested):
You're getting closer here! I found another batch of writing errors (happy to keep doing this, or you can proofread your own writing :smiley:), and there's a different, bigger thing: I asked last time whether you were using your own parser, and this current draft says you are.

But I was still confused about the distinction between the "main" tests and the "bril2json-depending" tests, so I took a look at your code from your PR, and I don't see a parser there. You are reusing the existing parser. (There is custom logic for tracking "ignored" lines to try to guess which line of the pre-parsed input corresponds to which parsed instruction.) If nothing else, it's important for your writeup to accurately reflect what you did for your project.

content/blog/2019-09-25-type_checker.md
tissue3: There are two categories because I am trying to make distinction between type check and anything should be checked. For example, I feel label check has nothing to do with type, but it's parsable and an error should be reported.

tissue3: Number checking checks the required arguments to an operator. For example,`int: v1 = add v0;` is parsable, but it's wrong.
Label refers to branch/jmp target. Such target should not appear twice. For example, `jmp label;<br/> label: a:int=1;<br/> label: a:int =2; ` makes jmp don't know where to go.

sampsyo: Sounds reasonable! I was confused by the reference to bril2json, but that distinction makes sense to me. :)

sampsyo: Both sound like good explanations!

sampsyo:
Nice work!!

PR #36, opened 2019-09-26
Project 1: Integer SIMD Support

### LINKED ISSUE #7: Project 1 Proposal: Integer SIMD support
I plan to add bril support for SIMD instructions and simple auto-vectorization in the Typescript frontend.

I'll modify the brili interpreter to support vector types in addition to int types. Vector instructions will be executed using Typescript's vectorization support. To support auto-verctorizaton, I will modify ts2bril. When a loop is encountered, it will need to be partially unrolled and SIMD bril instructions must be emitted.

I expect slight program speedups due to the use of vector instructions (assuming there are enough loop iterations!).

pbb59:
Resolves #7. Couldn't get the image to work in Zola.


sampsyo:
Fixed the image for you.

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Vector Instructions
+
+Single-Instruction Multiple-Data (SIMD) allows computer hardware to obtain signifigant speedups over the conventional execution paradigm, Multiple-Instruction Multiple-Data (MIMD). SIMD executes fewer instructions than MIMD to accomplish the same amount of work. In SIMD, multiple arithmetic operations are grouped together under one instruction. Generally, this aritmetic is executed over multiple execution units at the same instant (spatial SIMD), although some implementations allow the same instruction to use the same functional unit over multiple cycles (temporal SIMD). Various hardware architectures allow the programmer to support SIMD by exposing vector instructions in their ISA. Typically, there is a vector register file which holds multiple elements per register and can participate in vector arithmetic.
```
sampsyo:
```suggestion
Single-Instruction Multiple-Data (SIMD) allows computer hardware to obtain signifigant speedups over the conventional execution paradigm, Multiple-Instruction Multiple-Data (MIMD). SIMD executes fewer instructions than MIMD to accomplish the same amount of work. In SIMD, multiple arithmetic operations are grouped together under one instruction. Generally, this arithmetic is executed over multiple execution units at the same instant (spatial SIMD), although some implementations allow the same instruction to use the same functional unit over multiple cycles (temporal SIMD). Various hardware architectures allow the programmer to support SIMD by exposing vector instructions in their ISA. Typically, there is a vector register file which holds multiple elements per register and can participate in vector arithmetic.
```

Please run spell check. :smiley:

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Vector Support in the Bril Interpreter
+
+### Memory Support
+
+Vector instructions are only useful when operating on large chunks of data. Conversely, registers are not designed to hold this amount of data. A prerequisite for vector instructions is support for data memory in the interpreter. The interpreter memory is implemented like a stack. The program is allowed to access any location in the fixed size memory. Stack frames were not implemented because the interpreter only supports a single function.
+
+Memory access requires both loads and store operations. To emulate arrays, we can loop over multiple addresses and perform a load and/or store at each location. Load and store instructions in Bril are implemented similarly to their assembly counterparts. A load takes in a memory address (an `int` register in Bril) and writes to a destination register. A store is an effect operation and does not write to a register. It uses two source registers: a register containing a value and a register containing an address. The Bril syntax is given below.
```
sampsyo:
```suggestion
Memory access requires both load and store operations. To emulate arrays, we can loop over multiple addresses and perform a load and/or store at each location. Load and store instructions in Bril are implemented similarly to their assembly counterparts. A load takes in a memory address (an `int` register in Bril) and writes to a destination register. A store is an effect operation and does not write to a register. It uses two source registers: a register containing a value and a register containing an address. The Bril syntax is given below.
```

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+```C
+// load value from address 2
```
sampsyo: The text format uses `#` for comments, not `//`.

content/blog/2019-09-25-interpreter-vector-support/index.md
sampsyo: Maybe it's worth mentioning the names of the instructions, `lw` and `sw`?

Also, FWIW, you might be interested in this other design with arrays, loads, and stores:
https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/manually-managed-memory/

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Vector Support in the Bril Interpreter
+
+### Memory Support
+
+Vector instructions are only useful when operating on large chunks of data. Conversely, registers are not designed to hold this amount of data. A prerequisite for vector instructions is support for data memory in the interpreter. The interpreter memory is implemented like a stack. The program is allowed to access any location in the fixed size memory. Stack frames were not implemented because the interpreter only supports a single function.
```
sampsyo: I'm not quite sure what you mean by this:

> The interpreter memory is implemented like a stack.

It sounds like this is actually like a big C heap, i.e., a flat address space you can load and store from? It doesn't seem much like a stack, even ignoring the fact that there are no function calls.

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+addr: int = const 2;
+data: int = lw addr;
+
+// store value to address 2
+sw data addr;
+```
+
+### Interpreted Vector Instructions
+
+Vector instructions were added to the Bril interpreter. We specifically implement fixed-sized vector instructions of length four (akin to native Intel `__m128` SSE instructions). Typescript and Javascript (Typescript always compiles to Javascript) do not have support for vector instrinsics in the current standard. Thus, we implement the Bril vector instructions as a loop over four values. Additionally, we add vector registers in Bril which must be used in the vector instructions. We target a vector-vector add (vvadd) program, so we include interpreter support for `vadd`, `vload`, and `vstore` instructions. The `vload` and `vstore` instructions communicate data between vector registers and the interpreter stack. The `vadd` instructions adds two vector registers and writes to a destination vector register. An example vvadd program is shown below.
```
sampsyo:
```suggestion
Vector instructions were added to the Bril interpreter. We specifically implement fixed-sized vector instructions of length four (akin to native Intel `__m128` SSE instructions). TypeScript and JavaScript (TypeScript always compiles to JavaScript) do not have support for vector instrinsics in the current standard. Thus, we implement the Bril vector instructions as a loop over four values. Additionally, we add vector registers in Bril which must be used in the vector instructions. We target a vector-vector add (vvadd) program, so we include interpreter support for `vadd`, `vload`, and `vstore` instructions. The `vload` and `vstore` instructions communicate data between vector registers and the interpreter stack. The `vadd` instructions adds two vector registers and writes to a destination vector register. An example vvadd program is shown below.
```

TS and JS both have two capital letters in them. (Please find/replace throughout.)

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+// do the vector add
+vc: vector = vadd va vb;
+
+// store from vector register to memory
+vstore vc c;
+```
+
+### Intrinsic Vector Support
+
+It's awkward that the interpreter supports vector instructions, but doesn't actually use a native vector assembly instruction to perform the computation. We expect that the performance of the interpreted vector instructions will be poor. To explore this hypothesis, we create a version of an interpreted vector add instruction using three methods: 1) Typescript, 2) serial c++, and 3) vectorized c++. We run each test for 10,000 iterations and average the execution time over five runs. We assumed 10,000 iterations was enough time for the Typescript/Javascript JIT to warmup. The average time per loop iteration for each implementation is given below.
```
sampsyo:
```suggestion
It's awkward that the interpreter supports vector instructions but doesn't actually use a native vector assembly instruction to perform the computation. We expect that the performance of the interpreted vector instructions will be poor. To explore this hypothesis, we create a version of an interpreted vector add instruction using three methods: 1) TypeScript, 2) serial C++, and 3) vectorized C++. We run each test for 10,000 iterations and average the execution time over five runs. We assumed 10,000 iterations was enough time for the TypeScript/JavaScript JIT to warmup. The average time per loop iteration for each implementation is given below.
```

Dependent clauses => no comma. Also C++ is capitalized.

sampsyo: To be specific about the JIT, you're using [Node](https://nodejs.org/en/), which uses the [Google's V8 JIT](https://v8.dev) to execute JS.

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+|     Method    | Time per iteration (ns) |
+| ------------- | ------------- |
+| Typescript  | 317  |
```
sampsyo:
```suggestion
| TypeScript  | 317  |
```

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+### C++ binding for Typescript
+
+In order to utilize vector intrinsics, we need to call the C++ implementations in the Bril interpreter. We use [nbind](https://github.com/charto/nbind) to allow Typescript to execute binaries generated from C++. These sorts of calls will add potentially significant overhead to the execution. We quantify this overhead to see if it is practical in the intepreter. Note that each time we run a single vector instruction we must make a call to the binding. We run a vector add program with various iterations. Each iteration does two vector loads, a vector add, and a vector store along with instructions to facilitate the iteration. We run five configurations (128, 1024, 2048, 4196, 8192) and average the execution time over five runs. We compare the execution time with and without calls to the binding (literally just comment the line out). Note that the calls include passing arguments to the C++ binary, which incurs some additional overhead. On average there is a 10% overhead in the program due to the binding call. This overhead is expected to be offset from the substantial speedups offered by the C++ implementation.
```
sampsyo:
```suggestion
In order to utilize vector intrinsics, we need to call the C++ implementations in the Bril interpreter. We use [nbind](https://github.com/charto/nbind) to allow TypeScript to execute binaries generated from C++. These sorts of calls will add potentially significant overhead to the execution. We quantify this overhead to see if it is practical in the intepreter. Note that each time we run a single vector instruction we must make a call to the binding. We run a vector add program with various iterations. Each iteration does two vector loads, a vector add, and a vector store along with instructions to facilitate the iteration. We run five configurations (128, 1024, 2048, 4196, 8192) and average the execution time over five runs. We compare the execution time with and without calls to the binding (literally just comment the line out). Note that the calls include passing arguments to the C++ binary, which incurs some additional overhead. On average there is a 10% overhead in the program due to the binding call. This overhead is expected to be offset from the substantial speedups offered by the C++ implementation.
```

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+|     Method    | Time per iteration (ns) |
+| ------------- | ------------- |
+| Typescript  | 317  |
+| Serial C++  | 16  |
+| Vector C++  | 9  |
+
+There is a fair benefit to using native vector instruction from C (35x) speedup over the Typescript version. A large portion of this benefit comes from just using C in the first place. The serial C version achieves 20x better performance while the vector version further improves the performance by a more modest 1.8x.
+
+### C++ binding for Typescript
```
sampsyo:
```suggestion
### C++ binding for TypeScript
```

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+// iterations (increment by vector length)
+i: int = add i vecSize;
+done: bool = ge i size;
+br done vvadd_done vvadd_loop;
+
+vvadd_done:
+```
+
+Notice that the program does not initialize values in memory or print  the results. We do not want to include the initialization and cleanup time in the measured runtime. In the future, we could implemenet a Bril timer on/off instruction that could solve this problem. We time the execution of the interpreter using Typescript's `console.time` and `console.timeEnd` functions. We take care not to time the file I/O part of the interpreter as this would dominate the runtime for the program sizes that we run.
```
sampsyo: Those are really JS's standard library functions.

content/blog/2019-09-25-interpreter-vector-support/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Evaluation
+
+### Correctness
+
+We write multiple test programs to verify that the memory and vector instructions functioned as expected. [Turnt](https://github.com/cucapra/turnt) is used to test the expected output of the program from `print` instructions. The vector programs could not be verified with Turnt, however, because it needs to be executed in the interpreter directory to find the location of the C++ binaries. These programs were verified by manually inspecting the output. We test a simple store and then load, multiple stores and then multiple loads, and vvadd with both the Typescript implementation and the C++ vector implementation. All functioned as specified in this document.
```
sampsyo: I think you should be able to make Turnt work—you could, like, `cd /path/to/interp ; ...` to hack it?

sampsyo (changes requested):
Looks great! Here are some low-level writing suggestions.

pbb59:
Added changes, but not sure what you meant about the '#' for comments instead of '//' in the pseudocode examples?

sampsyo:
Awesome! I just meant that the Bril syntax actually does have comments, and they start with `#` rather than `//`. But if it's just pseudocode, I guess it doesn't matter.

PR #39, opened 2019-09-26
Project 1 Report: Type Inference for Bril

### LINKED ISSUE #13: Project 1 Proposal: Type Inference for Bril
**What will you do?**
I will create a type inference system for Bril. Specifically, you should be able
to take a Bril program that specifies only some or none of the types on the
variables and produces a new Bril program that has all the type annotations
(or throws an error).
For example, one will be able to write this bril program:
```
main {
  v0 = const 1;
  v1 = const 2;
  v2 = add v0 v1;
  print v2;
}
```
and transform it into this bril program:
```
main {
  v0: int = const 1;
  v1: int = const 2;
  v2: int = add v0 v1;
  print v2;
}
```

This is useful in cases where the frontend isn't typed. For example, someone
building a Python frontend would need to specify the types of variables, which
may be annoying. This takes some burden off the frontend implementation.
From this, we should get typechecking for free.

**How will you do it?**
First I'll ignore all existing types to the program, and use the language reference
to determine which operations determine the types of each variable. Then for free
type checking, I will run a pass over the original program and make sure the types
we've inferred match what was already given.
I will also change the grammar for `bril2txt` to allow for optional typing on
variables for bril programs.

First I will take the existing bril programs, remove the type annotations, and
run my type inference program. I will make sure that the output matches the
existing bril programs to verify correctness. Then I will hand curate some
bril programs that shouldn't type check and verify that they don't. I'll also
write bril programs that only explicitly type some variables to make sure the
type inferrer can handle these cases.

chrisroman:
Closes #13

### DIFF
--- a/content/blog/2019-09-25-bril-type-inference.md
+++ b/content/blog/2019-09-25-bril-type-inference.md
@@ -1,0 +1,194 @@
++++
+title = "Type Inference for Bril"
+extra.bio = """
+  Christopher Roman is a second semester MEng student in computer science. He is interested in compilers and distributed systems. He is also the best [Melee](https://en.wikipedia.org/wiki/Super_Smash_Bros._Melee) player at Cornell :)
+"""
+[[extra.authors]]
+name = "Christopher Roman"
++++
+
+## Overview
+The goal of this is to build a type inferrer for Bril. Ultimately, we want to
+take a Bril program that has some (or none) of the types specified on variables
+and produce a new Bril program that has all the correct type annotations.
+For example, this program:
+```
+main {
+  v0 = const 1;
+  v1 = const 2;
+  v2 = add v0 v1;
+  print v2;
+}
+```
+should be transformed into the program:
+```
+main {
+  v0: int = const 1;
+  v1: int = const 2;
+  v2: int = add v0 v1;
+  print v2;
+}
+```
+This is useful in cases where the frontend isn't typed. For example, someone
+building a Python frontend would need to specify the types of variables, which
+may be annoying. This takes some burden off the frontend implementation.
+Additionally, it makes it quicker to write Bril programs because we don't have
+to worry about correctly typing the variables.
+
+We also will implement type checking through this type inference.
+
+## Design
+For our workflow, we'd like to take a semi-typed bril program and convert it
+between JSON and text representations. This requires changing the grammar so
+that type annotations are optional.
+
+Because Bril only supports two types, `int` and `bool`, the type inference is
+very straight forward. For example, consider arithmetic operations, i.e.
+`add`, `mul`, `sub`, and `div`. The arguments to this operations *must* be ints,
+and the result is an int. Therefore, if we see a statement like `x = add a b`,
+we know `x`, `a`, and `b` are all ints. If at any point we inferred that one
+of these variables are not ints, then we have a type unification error, and the
+program is not well-typed.
+
+## Implementation
+
+### Modifying the Grammar
+We want to make type annotations optional. So first I modified the grammar to
+have a rule for type annotations:
+```
+type_decl.5: ":" type
+const.4: IDENT [type_decl] "=" "const" lit ";"
+...
+```
+This is an issue however because there is an ambiguity between labels and
+assignments. Consider these two Bril programs:
+```
+main {
+l:
+  x = const 5;
+}
+```
+and
+```
+main {
+  l: x = const 5;
+}
+```
+These programs consist of the same tokens. However, the first one intuitively
+means that we have a label `l` and some int `x` with a value of 5. The second
+one means that we have a variable called `l` of type `x` with a value of 5.
+It is incorrect to simply have one rule as a higher priority, because
+semantically, both of these should be allowed as separate programs. So we have
+two options:
+1. Force labels to be on its own line
+2. Only allow fixed type names, like "int" and "bool"
+
+I decided to go with (2) because I didn't know how to easily do (1), but in
+retrospect (1) may have been better, because some other students' projects
+allow for user-defined types.
+
+### Type Inference
+As noted in the design, inferring the types of variables for a single statement
+is straightforward. Here is an example snippet showing how types for comparison
+ops are inferred:
+```
+...
+elif instr["op"] in COMPARISON_OPS:
+    for arg in instr["args"]:
+        type_var(gamma, arg, "int", i)
+    type_var(gamma, instr["dest"], "bool", i)
+...
+```
+Here, we keep track of a typing context `gamma` which maps variables to their
+type. Then, we check that each argument is either untyped or already has the
+type `int`; otherwise we'll throw a type unification error. Finally, we do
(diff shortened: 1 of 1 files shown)


content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
++++
+title = "Type Inference for Bril"
+extra.bio = """
+  Christopher Roman is a second semester MEng student in computer science. He is interested in compilers and distributed systems. He is also the best [Melee](https://en.wikipedia.org/wiki/Super_Smash_Bros._Melee) player at Cornell :)
```
sampsyo:
```suggestion
  Christopher Roman is a second semester MEng student in computer science. He is interested in compilers and distributed systems. He is also the best [Melee](https://en.wikipedia.org/wiki/Super_Smash_Bros._Melee) player at Cornell. :)
```

Gotta complete those sentences, even when they're jokes. :smiley:

content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
...
+"""
+[[extra.authors]]
+name = "Christopher Roman"
++++
+
+## Overview
+The goal of this is to build a type inferrer for Bril. Ultimately, we want to
```
sampsyo:
```suggestion
The goal of this project is to build a type inferrer for Bril. Ultimately, we want to
```

content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
...
+  v1: int = const 2;
+  v2: int = add v0 v1;
+  print v2;
+}
+```
+This is useful in cases where the frontend isn't typed. For example, someone
+building a Python frontend would need to specify the types of variables, which
+may be annoying. This takes some burden off the frontend implementation.
+Additionally, it makes it quicker to write Bril programs because we don't have
+to worry about correctly typing the variables.
```
sampsyo: FWIW, I think the Python motivation is somewhat imperfect—for arbitrary Python programs, it may be *impossible* to come up with a static typing with the same semantics. That's just the nature of a dynamic language. So we'd still be stuck compiling a statically-typed *subset* of Python, even if the backend infers the types.

But making it easier to write the IR makes sense!

content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
...
+
+We also will implement type checking through this type inference.
+
+## Design
+For our workflow, we'd like to take a semi-typed bril program and convert it
```
sampsyo:
```suggestion
For our workflow, we'd like to take a semi-typed Bril program and convert it
```

content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
...
+between JSON and text representations. This requires changing the grammar so
+that type annotations are optional.
+
+Because Bril only supports two types, `int` and `bool`, the type inference is
+very straight forward. For example, consider arithmetic operations, i.e.
```
sampsyo:
```suggestion
very straightforward. For example, consider arithmetic operations, i.e.,
```

content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
...
+one means that we have a variable called `l` of type `x` with a value of 5.
+It is incorrect to simply have one rule as a higher priority, because
+semantically, both of these should be allowed as separate programs. So we have
+two options:
+1. Force labels to be on its own line
+2. Only allow fixed type names, like "int" and "bool"
+
+I decided to go with (2) because I didn't know how to easily do (1), but in
+retrospect (1) may have been better, because some other students' projects
+allow for user-defined types.
```
sampsyo: Yeah, probably so. :smiley:

content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
...
+}
+```
+This is a valid program that should typecheck. Specifically, `x` and `y` are
+both ints. However, if we naively go sequentially through each instruction, we
+don't know the type of `y` until we have `y = const 5`, at which point it is too
+late to type `x`. To resolve this, we have two options:
+1. Rerun the type inference algorithm, stopping when no more variables have been
+inferred
+2. Keep track of which variables must have the same type, and after 1 pass,
+setting their types to be the same.
```
sampsyo: It's worth noting that this is basically constraint-based type inference (i.e., how type inference works for "real" languages).

content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
...
+To properly evaluate that everything presented here is correct, we have to test
+the parser and the type inferrer. To do these simultaneously, we can run tests
+as follows:
+```
+command = "cat {filename} | bril2json | python ../../infer.py | bril2txt"
+```
+We are taking Bril text programs, turning them into JSON, generating a new
+equivalent typed program, and turning it back to text. First I started with
+simple Bril programs that already existed to build some confidence. Then to gain
+full confidence, I wrote tests that use every kind of operations, e.g.
```
sampsyo:
```suggestion
full confidence, I wrote tests that use every kind of operations, e.g.,
```

content/blog/2019-09-25-bril-type-inference.md
```diff
@@ -0,0 +1,204 @@
...
+that *shouldn't* typecheck in fact could not be type-inferred. For example:
+```
+main {
+  F1 = const false;
+  T1 = const 1;
+  b1 = and F1 T1;
+}
+```
+Unfortunately I couldn't get Turnt to validate the error message that was
+output, so instead I manually checked to see that the error was what I expected.
```
sampsyo: FWIW, this capability was recently added: https://github.com/cucapra/turnt/pull/9

sampsyo (changes requested):
Looks excellent! This seems like a legitimately useful tool. Here are a few writing suggestions.

content/blog/2019-09-25-bril-type-inference.md
chrisroman: I see, I never thought about that. To give some clearer motivation, I've just removed this example in 10c884b

content/blog/2019-09-25-bril-type-inference.md
chrisroman: Awesome! This is a pretty useful feature.

content/blog/2019-09-25-bril-type-inference.md
chrisroman: :O did not know this. TIL

chrisroman:
I've fixed all suggested changes. Hopefully it's ready for merging :)

sampsyo (approved):
Awesome; looks great!

PR #43, opened 2019-09-26
Project 1: Run Bril on Raspberry Pi Natively

### LINKED ISSUE #9: Project 1 Proposal: Generate ARM64 assembly from Bril
**What will you do?**
Write a code generation program that takes a JSON representation of a Bril program and generates aarch64 assembly code. The generated assembly code should be able to run on aarch64 processor, such as Raspberry Pi 3 (or later versions, which have 64 bit ARM processor)

**How will you do it?**
The code generation program should not only support current Bril instructions, but it also should be easy to expand for more features, such as interesting features implemented by other teams.
Therefore, I will first build a AST from JSON Bril, then generate assembly code for each operations. Most of current Bril operation are easy to do, there are almost one-to-one translations between Bril and ARM assembly. The only relatively tricky part is print, which could be solved by calling printf from C.

1. Build a benchmark set of different types of workload
2. Comparing the performance between ARM64 binary and brili (maybe and some interpreter or binaries from other team) on a Raspberry Pi 3B+ board.

0x486F626F:
Closes #9


sampsyo (changes requested):
This is great! Nice work, and thanks for generating visualizations. Here are some writing suggestions.

sampsyo (changes requested):
This is great! Nice work, and thanks for generating visualizations. Here are some writing suggestions.

sampsyo:
Excellent! Thanks for the revisions.

PR #45, opened 2019-09-26
Project 1: Vril (Vector Bril)

### LINKED ISSUE #24: Project 1 Proposal: Vril (Vector Bril)
**What will you do?**
The idea is to have support for a high level programming language construct to represent vector data structures. This construct will have a one-to-one mapping to Bril IR array types the language also has associated vector operations. The main motivation of this construct and its associated operations is to by-pass the front-end of the compiler that would try to detect data-level parallelism, and automatically generate vector code. Instead, we are interested in focusing on the back-end optimizations of this vector types and vector operations.

**How will you do it?**
We need to define a new type "array of ints" to represent the vector types. We want these to have variable length.  To start, we would like to add to the list of Bril instructions their corresponding vertical and horizontal vector operations: "vadd" and "vredadd", for "add".

After successfully implementing the new type and operations, we will code a few micro-benchmarks that utilize the language construct and their functionally identical scalar version, to generate their corresponding DFGs. To measure sucess, we want to make two observations: a) the two DFGs will contain a similar shape and the same number of basic blocks (BB), and b) the scalar program should be hopping the BBs in the DFG more times than the vector program. In particular, the BBs of the DFG involved in the `for-loop` part of the program should be hopped _vector-length times_ more in the scalar program than the vector one.

With @sa2257 & @qtt2 (audit).

sa2257:
This PR creates a blog post by @hcp922, @qtt2 and @sa2257 for Vril, which adds arrays and vector operation `vadd` to Bril. This closes #24 which proposed this project. Vril is implemented in [Vril](https://github.com/sa2257/vril).


content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+For now there is only one configuration op: `setvl`. 
+```
+vl: int setvl val
+```
+is used to request `val` lanes to the IR and `setvl` returns `vl:=val` if `vl<maxvl` or `vl:=maxvl` if `vl>=maxvl`.
+We have also implemented only one arithmetic vector op `vadd` as a proof of concept:
+```
+arr3: array vadd arr1 arr2
+```
+Which takes two arrays as arguments and performs element-wise vector addition, and stores it into a third array.
```
sampsyo: I'm somewhat surprised that `vadd` adds together two *entire arrays*. Maybe you'd be interested in expanding a little more about why that's the case? Most SIMD ISAs only operate on short, fixed-length vectors—not unboundedly long ones. Especially since the configuration seems to introduce the notation of a "vector length," it seems natural that the vector instruction would add together two N-length chunks, where N is the vector length.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Simulating vector ops in Vrili (Vril interpreter)
+We have extended Brili to interpret vector operations.
+Vrili executes for loops across the different elements of the array operands in order to produce the correct result. This emulates what a vector microarchitecture would do concurrently with a single fetch-decode effort.
+
+Evaluation
+----------------------------
+
+Vril makes two major additions to Bril, adding Arrays and extending Briloperations with vector operations. This section describes how we evaluate each of these.
```
sampsyo:
```suggestion
Vril makes two major additions to Bril, adding Arrays and extending Bril operations with vector operations. This section describes how we evaluate each of these.
```

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+We take three approaches to evaluate arrays,
+- evaluating integration to Bril
+- qualitative evaluation to justify arrays
+- evaluating functionality of arrays
+
+#### Integration tests for arrays
+We ran tests to generate JSON representation of Bril from text format and vice-versa to evaluate whether the array specification is expressible in both available formats. These tests reside in `vtest/parse` and `vtest/print` directories.
+
+#### Qualitative evaluation of arrays
+Arrays provide a convenient method to use loops in Bril. To add two groups of 10 numbers, conventional Brili required 44 lines of code, 20 of which would define the 20 numbers and 10 to add them. Vril required 50 lines of code, 20 of which were still used to define two arrays element by element and 6 lines to add them in a loop body. It is observed that as the size of the arrays grow, conventional Bril would need correspondingly more lines of code where as Vril would still use 6 lines of code in the loop body.
```
sampsyo: FWIW, I don't think it takes much careful argument to justify the existence of arrays—the purpose and usefulness of arrays is pretty well-understood. Even so, I like the idea of showing that the number of instructions is smaller.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+Vril makes two major additions to Bril, adding Arrays and extending Briloperations with vector operations. This section describes how we evaluate each of these.
+
+### Evaluating Arrays
+We take three approaches to evaluate arrays,
+- evaluating integration to Bril
+- qualitative evaluation to justify arrays
+- evaluating functionality of arrays
+
+#### Integration tests for arrays
+We ran tests to generate JSON representation of Bril from text format and vice-versa to evaluate whether the array specification is expressible in both available formats. These tests reside in `vtest/parse` and `vtest/print` directories.
```
sampsyo: To be clear, this _just_ tests the parser and pretty-printer. So I don't think I would call this "evaluating integration to Bril"; it's just testing the text representation machinery alone.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Functional evaluation of arrays
+We first wrote a simple test to initialize an array, store a value to an array and read a value from an array. 
+A second test described a loop which would add two array elements and save it in a third array.
+Both these tests generated the expected result from the interpreter. We also walked through the program to count the dynamic instruction count and number of hops over basic blocks. A program with 50 lines of code created 109 dynamic instructions and 11 hops, which matched the expectation. These tests reside in `vtest/interp` directory.
+
+### Evaluating Vector operation `vadd`
+We took a similar approach to evaluate our addition of `vadd` in Vril.
+- evaluating integration to Bril
```
sampsyo: As above.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+- evaluating functionality of `vadd`
+- evaluating performance of `vadd`
+
+#### Integration tests for Vectors
+We added tests to generate JSON and text representation of Bril from each other to gain confidence about its translatability. These tests are included in `vtest/parse` and `vtest/print`.
+
+#### Functional testing of `vadd`
+We wrote a program with `vadd` and ran this test in `vrili` with different values for `maxvl` which represents the size of vector lenght of the backend. The test generated expected results. This is included in `vtest/interp/array_vector.bril`.
```
sampsyo: The text in this section (and the previous one) is extremely similar to the ones in the other evaluation section, for arrays. Maybe consider deduplicating by describing your general approach up front and then specializing it for the two cases?

sampsyo: What exactly does testing with different `maxvl`s do? Does it change anything in what the program does?

sampsyo:
```suggestion
We wrote a program with `vadd` and ran this test in `vrili` with different values for `maxvl` which represents the size of vector length of the backend. The test generated expected results. This is included in `vtest/interp/array_vector.bril`.
```
Please run spell check.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Performance of `vadd`
+Initial motivation for Vril is to reduce hop counts by having vector operators in Bril. This aspect is tested by comparing dynamic instruction count, hop count and lines of code for a loop executed in a scalar manner and a vector operation.
```
sampsyo:
```suggestion
The initial motivation for Vril is to reduce hop counts by having vector operators in Bril. This aspect is tested by comparing the dynamic instruction count, hop count, and lines of code for a loop executed in a scalar manner and a vector operation. 
```

sampsyo: Can you clarify precisely what you mean by "hop count"? Is it the number of branch/jump instructions executed?

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+#### Functional testing of `vadd`
+We wrote a program with `vadd` and ran this test in `vrili` with different values for `maxvl` which represents the size of vector lenght of the backend. The test generated expected results. This is included in `vtest/interp/array_vector.bril`.
+
+#### Performance of `vadd`
+Initial motivation for Vril is to reduce hop counts by having vector operators in Bril. This aspect is tested by comparing dynamic instruction count, hop count and lines of code for a loop executed in a scalar manner and a vector operation. 
+
+| Program           | Lines of code | Dynamic instructions | Hop count |
+|-------------------|---------------|----------------------|-----------|
+| Conventional Bril | 43            | 41                   | 1         |
+| Scalar Array      | 50            | 109                  | 12        |
```
sampsyo: I'm somewhat surprised that the array with a loop required _more_ instructions than the version with 10 separate variables. Can you expand on why this is the case?

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+| Vector Array      | 48            | 44                   | 3         |
+
+Given that the length of array is N=10 and vector length selected by Vrili at execution is k, and compensating for the start and the end block as shown in the figure, scalar operation iterates N times and vector operation iterates N/k times.
+
+CFG for scalar array operation  | CFG for vector array operation
+:-------------------------:|:-------------------------:
+![](https://github.com/sa2257/vril/blob/82e9da0eed50a98fc5285a32020b0ba59385e398/plots/array_scalar_nv.png) | ![](https://github.com/sa2257/vril/blob/82e9da0eed50a98fc5285a32020b0ba59385e398/plots/array_vector_nv.png)
```
sampsyo: It's possible to include these images right here in the blog post. If you make a directory for your blog post instead of a file, and then put the text of the post in index.md inside that directory, you can put images alongside the Markdown file and refer to them relatively.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+
+Conclusion
+--------------------------------------
+We have extended bril to support array types. We have added two new operations to move data in and out of the arrays so that we can emulate data movements between an array and what it would be a scalar register. We have also extended bril to support vector operations of two types: configuration and arithmetic. Configuration operations allow to modify a vector state and arithmetic operations perform operations on array arguments.
```
sampsyo:
```suggestion
We have extended bril to support array types. We have added two new operations to move data in and out of the arrays so that we can emulate data movements between an array and what it would be a scalar register. We have also extended bril to support vector operations of two types: configuration and arithmetic. Configuration operations allow programs to modify a vector state and arithmetic operations perform operations on array arguments.
```
(Missing subject.)

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+
+The goal of this exercise was to understand how much an IR needs to change in order to express data-parallel operations. For that, we extended bril to be able to express vector operations and to compare their potential against a traditional scalar set of operations. For that we written a benchmark for vector-vector add (vvadd), which adds the elements of two arrays and stores their results into a third array in two versions: a scalar code and a vector code. The end goal is to verify that the CFG generated is very similar for both codes, it should contain the same number of basic blocks (BB). However, the vector code hops on the BB involved in the `for-loop` statement *vector length* times less than the scalar code.
```
sampsyo:
```suggestion
The goal of this exercise was to understand how much an IR needs to change in order to express data-parallel operations. For that, we extended bril to be able to express vector operations and to compare their potential against a traditional scalar set of operations. For that we wrote a benchmark for vector-vector add (vvadd), which adds the elements of two arrays and stores their results into a third array in two versions: a scalar code and a vector code. The end goal is to verify that the CFG generated is very similar for both codes: it should contain the same number of basic blocks (BB). However, the vector code hops on the BB involved in the `for-loop` statement *vector length* times less than the scalar code. 
```

sampsyo: This paragraph suggests that I still don't quite understand the role of the vector length—I thought that `vadd` added together two entire arrays? But this paragraph seems to suggests it adds together only an N-sized chunk of the array, where N is the vector length.

sampsyo (changes requested):
Hello! This is a nice start! I have to admit I'm still somewhat confused, after finishing a complete read-through, about what the "vector length" is and how it influences the operation of the vector operations like `vadd`. It might be useful to be more precise about how these instructions work and to add examples of code that uses them, describing exactly what the instructions do.

sampsyo:
One additional semantics question occurred to me this morning: when you use “id” to assign an array into a new variable, does that create a copy? Or do both names refer to the same underlying array (so changes in one will be reflected in the other)?

content/blog/2019-09-25-vril-vector-bril.md
sa2257: My expectation is to gain confidence that our changes to Bril IR reflect our intentions. I'm trying to do that by writing a program and inspecting the generated IR from the parser.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: Noted.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: Noted.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: Sounds good.

sa2257: We use `maxvl` as backend information for vector length. `setvl` will use this information to decide whether the requested vector length can be provided to the program. Therefore, `maxvl` will change how many iterations the loop will run in the vector add example.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: It is. To be specific, it counts the number of basic blocks the program traverse.
I think I should define what hop count is. I'll add that into the document.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: Yes, this is because array programs have the overhead of initializing the arrays element-wise. I have added this description [here](https://github.com/sampsyo/cs6120/blob/ed80a1e742562fab205319bedf65c46cabbc72d0/content/blog/2019-09-25-vril-vector-bril.md#qualitative-evaluation-of-arrays), but from the [comment](https://github.com/sampsyo/cs6120/pull/45/files#r330823062) this description should be moved here or linked with this table.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: Thanks. Will do. I wasn't sure how to preview this to see if it's working.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: From the programmer's point of view it will add two entire arrays. However, if the architecture does not provide sufficient vector length to do so, it will do this in chunks of N, where N is the architecture's vector length, i.e. `maxvl` in Vrili.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: As to why, it's trying to address https://github.com/sampsyo/cs6120/pull/45/files#r330825016. I think it's better to link this with the results table.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+[Vril](https://github.com/sa2257/vril) is our public repository.
+
+Arrays
+------------------
+
+### Arrays in Vril
+Definition of an array in Vril is a sequence of elements of any type (no type checking carried out) and the length being any Bril literal. 
+
+### Initializing arrays in Vril
+Vril extends the grammar of Bril to include arrays by adding an `init` operator to initialize an array of length `l` as follows,
```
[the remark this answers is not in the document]
hcp922: We initialize new arrays to all zeroes. Technically, one could create and manipulate a variable called arr1[0] and later initialize an array called arr1 which would replace the old value of arr1[0] by a zero. Similarly, one could create a variable called arr2[4] and later initialize an array of 4 elements (which would store zeroes in arr2[i], i=0..3). The user would be able to legally access arr2[i], for i=0..4, while it is not clear that it was their intention.  We should have mentioned that, we will add it to the blog.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Vector ops in Vril
+There are two types of vector ops in Vril: configuration ops and arithmetic/logic ops. Vector arithmetic or logic operations take arrays as arguments and produce an array or a scalar value as a result. Instead, configuration ops help in configuring a vector state.
+
+### Adding vector ops to Vril
+For now there is only one configuration op: `setvl`. 
+```
+vl: int setvl val
+```
+is used to request `val` lanes to the IR and `setvl` returns `vl:=val` if `vl<maxvl` or `vl:=maxvl` if `vl>=maxvl`.
```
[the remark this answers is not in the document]
hcp922: There are two concepts we should have defined with more clarity: vector length (vl) and maximum vector length (max_vl). max_vl represents a limitation from the hardware and vl is the number of lanes currently active which can be only equal or smaller than max_vl. 

Inside a for-loop, the user can request a vector length which is usually the number of iterations of that loop (vl: int = setvl N). If the hardware can handle as many lanes as iterations, setvl instruction will return exactly the number of lanes requested. We use the value returned by setvl to increment the loop index variable (idx += vl). And so, the loop will be done in one single iteration. However, if the vector length requested is greater than max_vl, it will return max_vl and we will need to execute the for-loop in several iterations (in chunks of max_vl).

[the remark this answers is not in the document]
hcp922: This is based on RISC-V vector extension "V", and generally inspired by traditional vector architectures. It is useful since it does not couple the ISA to a specific vector length, like Intel's SSE (128-bit vectors), AVX (256-bit vectors) or AVX-512 (512 bit-vectors). That allows for the binary to be ran on different micro-architectural optimizations without the need to re-compiler, as well as to design machines with longer vectors without having to extend the ISA.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+We have also implemented only one arithmetic vector op `vadd` as a proof of concept:
+```
+arr3: array vadd arr1 arr2
```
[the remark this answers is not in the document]
hcp922: Oops! Thank you.

content/blog/2019-09-25-vril-vector-bril.md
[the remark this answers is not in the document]
hcp922: It does not matter if the arrays have different lengths because the vector operation will operate to vl elements. As long as both arrays have at least vl elements, it will work. For the second situation you propose, we purposely left it for future type checking work.

hcp922: You are absolutely correct. We will expand our explanation on max_vl and vl to clarify that we at most operate to max_vl elements and that we use the value of vl to increment the loop index.

content/blog/2019-09-25-vril-vector-bril.md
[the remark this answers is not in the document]
sampsyo: Got it; thanks for clarifying. (I am familiar with the RISC-V length-independent vector instruction set, but I didn't catch that's what you were going for.)

I think the thing that needs a bit more clarification here is: If the vector length is N, how does the program tell the machine *which* N elements of the arrays to add? That is, if `vadd` just takes two array arguments, how does the machine know when to start at the beginning of the array and when to start later in the array (in subsequent iterations)?

In RISC-V with vectors and ARM SVE, this works by issuing vector loads & stores to a specific offset within the array—i.e., by doing index math, just like in any other scalar or vector loop.

content/blog/2019-09-25-vril-vector-bril.md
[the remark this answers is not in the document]
sampsyo: Got it. It's worth mentioning, perhaps, that `vadd` assumes that the arrays are numeric. And also clarifying whether or not it's an error to `vadd` when the vector length would take you beyond the end of either array.

content/blog/2019-09-25-vril-vector-bril.md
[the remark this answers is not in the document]
hcp922: Great question. We also failed to explain this in the blog but if you look at our vrili examples we pass an index as one of the parameters to indicate the element where to start operating on: arr3: array = vadd arr1 arr2 i. Will add this ASAP!

In that sense, Vrili is similar to RISC-V V or Arm's SVE. But we opted for a RISC-V V or CRAY style, which has a centralized vector length state used by all vector instructions (until a new setvl instruction appears). We did think of Arm's SVE philosophy, where they construct predicate registers with 'whilelt' instructions or similar to then use in vector arithmetic/logic operations as masks. But we thought RISC-V's style would be a bit cleaner to implement in Vrili since we do not need to add the predicate variables, but just two scalar internal variables (vl and max_vl).

content/blog/2019-09-25-vril-vector-bril.md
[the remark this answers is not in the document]
hcp922: Totally agree! We will add this now.

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+
+There are three popular options to program for vector architectres, a.k.a. *vectorization*:
+- manual vectorization: Explicitly using vector operations in assembly or intrinsics.
+- user annotations or [pragmas](https://info.ornl.gov/sites/publications/files/Pub69214.pdf): Help the compiler find the *vectorizable* regions of code and inform the compiler about lack of dependences and other situations that would typically restrict the vectorization.
+- auto-vectorization: Rely directly on the compiler vectorizer which is typically enabled by the flag `-ftree-vectorize` and by default at `-O3` in [gcc](https://www.gnu.org/software/gcc/projects/tree-ssa/vectorization.html).
```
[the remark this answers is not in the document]
hcp922: Sure! We edited the [High-level idea Section](https://github.com/sampsyo/cs6120/blob/bf2c49b882cb61a884aa78f4b9ef7f1914be3acd/content/blog/2019-09-25-vril-vector-bril/index.md#high-level-idea).

content/blog/2019-09-25-vril-vector-bril.md
```diff
@@ -0,0 +1,139 @@
...
+Programmer lead vectorization allows explicit control, but usually becomes architecture specific and assumes vector-like programming from the programmer.
```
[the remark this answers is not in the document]
hcp922: Good point, we added it to [High-level idea Section](https://github.com/sampsyo/cs6120/blob/bf2c49b882cb61a884aa78f4b9ef7f1914be3acd/content/blog/2019-09-25-vril-vector-bril/index.md#high-level-idea).

hcp922:
> One additional semantics question occurred to me this morning: when you use “id” to assign an array into a new variable, does that create a copy? Or do both names refer to the same underlying array (so changes in one will be reflected in the other)?


If you use "id" like: `int: val = id arr1[0]`, the value of arr1[0] would be copied into variable val. But if you used "id" like `int: val = id arr1`, it would throw an error because arr1 is not a variable by itself but its elements (arr[i], for all valid indices `i`) are.

content/blog/2019-09-25-vril-vector-bril.md
sa2257: The fact that it happens in chunks of N is in programmer's view. This N is captured at runtime in `vl` to iterate in the loop. Actual N is microarchitectural and is specified in `maxvl`.

content/blog/2019-09-25-vril-vector-bril/index.md
```diff
@@ -0,0 +1,137 @@
...
+Vril can operate on arrays in two ways: 1) We can perform *scalar* accesses to arrays by moving one element of the array to a variable and performing any arithmetic or logic operation that are already supported in the original Bril, and 2) we can perform *vector* accesses to the arrays by using new vector operations that take arrays as arguments directly.
+
+In order to move data from an element of an array `arr[idx]` to a variable `var`, we can use `a2v` op:
+```
+var: int = a2v array_name idx
+```
+Similarly, we can move a value stored in a variable `var` into one of the array's element `arr[idx]`:
+```
+array_name: int = v2a var index
+```
```
sampsyo: No need to change this now, but it's a little weird that the array name appears in the "destination" position, i.e., on the left-hand side of the =. In every operation, the destination is where the result of the computation goes. But here the operation doesn't define a variable called `array_name`—instead, it puts a value into the array.

content/blog/2019-09-25-vril-vector-bril/index.md
```diff
@@ -0,0 +1,137 @@
...
+vl := (val <= max_vl)? val:max_vl;
+```
+where `max_vl` is the maximum amount of lanes or array elements that the micro-architecture can process in parallel.
+
+We have also implemented only one arithmetic vector op `vadd` as a proof of concept:
+```
+arr3: array = vadd arr1 arr2 idx
+```
+Which takes two arrays as arguments and performs element-wise vector addition to `vl` elements starting at element `idx`, and stores it into a third array. For now, if `arr1`, `arr2`, and `arr3` have different lengths, as long as elements in indices `i:i+vl-1` are valid for the three arrays (two sources and destination), the operation is valid.
```
sampsyo: Got it! This makes much more sense now that there's an `idx` parameter to `vadd`. Thanks for clarifying!

content/blog/2019-09-25-vril-vector-bril/index.md
```diff
@@ -0,0 +1,137 @@
...
+where `max_vl` is the maximum amount of lanes or array elements that the micro-architecture can process in parallel.
+
+We have also implemented only one arithmetic vector op `vadd` as a proof of concept:
+```
+arr3: array = vadd arr1 arr2 idx
+```
+Which takes two arrays as arguments and performs element-wise vector addition to `vl` elements starting at element `idx`, and stores it into a third array. For now, if `arr1`, `arr2`, and `arr3` have different lengths, as long as elements in indices `i:i+vl-1` are valid for the three arrays (two sources and destination), the operation is valid. 
+
+### Other types on arrays
+We have not implemented any array type checking on arrays. While we expect the programmer to only use integer values on array elements, there may not be any error thrown if they stored other types. Vector arithmetic or logic operations have undefined behavior for these cases. We left type checking for future work.
```
sampsyo:
```suggestion
We have not implemented any array type checking on arrays. While we expect the programmer to only use integer values on array elements; there may not be any error thrown if they stored other types. Vector arithmetic or logic operations have undefined behavior for these cases. We left type checking for future work. 
```

(Comma splice.)

content/blog/2019-09-25-vril-vector-bril/index.md
```diff
@@ -0,0 +1,137 @@
...
+```
+arr3: array = vadd arr1 arr2 idx
+```
+Which takes two arrays as arguments and performs element-wise vector addition to `vl` elements starting at element `idx`, and stores it into a third array. For now, if `arr1`, `arr2`, and `arr3` have different lengths, as long as elements in indices `i:i+vl-1` are valid for the three arrays (two sources and destination), the operation is valid. 
+
+### Other types on arrays
+We have not implemented any array type checking on arrays. While we expect the programmer to only use integer values on array elements, there may not be any error thrown if they stored other types. Vector arithmetic or logic operations have undefined behavior for these cases. We left type checking for future work. 
+
+### Simulating vector ops in Vrili (Vril interpreter)
+We have extended Brili to interpret vector operations.
```
sampsyo: I think it might be worth briefly mentioning the (somewhat surprising!) way your interpreter extension works. That is, when you do:

    five: int = const 5;
    foo: array = init five;

The interpreter environment doesn't ever actually contain a variable called `foo`! Instead, it gains _five different_ variables literally named `foo[0]`, `foo[1]`, `foo[2]`, `foo[3]`, and `foo[4]`. So you can access those variables normally, just like any other variable, if you want.

Similarly, the vector length is just a normal variable called `vl`. So you can read and write that variable using any other instruction too.

sampsyo (changes requested):
Looking good; thank you for clarifying all those semantics questions! I understand your extension much better now.

I just have a couple of minor writing suggestions for this round, and one request for a little more detail about your implementation. As you've mentioned above, the way you store arrays in the environment is interesting and worth including.

content/blog/2019-09-25-vril-vector-bril/index.md
hcp922: You could see as an insert operation (there use to be one in earlier versions of RISC-V V): `vinsert v0, a1, a2` which does exactly that: `v0[$a1] = $a2`. Do you mean that it looks too much like assembly and less like an IR language?

sa2257: I see. Yes, it does not have a destination register notion here. The syntax you've suggested indeed makes that evident.

content/blog/2019-09-25-vril-vector-bril/index.md
sa2257: Yes, the intention was to maintain the same assumed execution from Brili. Vrili is not simulating a vector processor, but merely simulating the functionality. I've explained how arrays are created as a bunch of variables in Vrili.

`vl` is indeed a normal variable. `setvl` puts either the requested length (input operant which is also a normal variable), or `maxvl` which is not accessible to the programmer.

PR #48, opened 2019-09-26
Project 1: Floating Points and Fixed-Length Arrays

### LINKED ISSUE #4: Project 1 Proposal: Floating Points and Compiling Bril to LLVM
**What will you do?**
I will implement a compiler from existing Bril code to the LLVM-IR.  This compiler will include selecting integer and floating-point operations based on provided numeric representation. As people implement new features, I'll try to keep this compiler up-to-date.

**How will you do it?**
I'll write the compiler in typescript.  I'll also update Bril with a floating-point type, along with a mechanism for differentiating `int` from `float` at compile time.

For each file, I'll compile to LLVM and test the resulting code.  I'll start with the existing test code first, then add a couple of floating-point tests.

Checkmate50:
Resolves #4


content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+title = "Floating Points and Fixed-Length Arrays in Bril"
+extra.author = "Dietrich Geisler"
+extra.bio = """
+  [Dietrich Geisler](https://www.cs.cornell.edu/~dgeisler/) is a 3rd year PhD student researching Language Design and Compilers.  Enjoys gaming and climbing.
+"""
++++
+
+### Introduction
+
+My goal for this project was to add floating points and fixed-length arrays to [Bril](https://github.com/sampsyo/bril).  The intention behind this decision is to promote Bril as a lower-level intermediate language, with fewer type abstractions than Javascript without the complexity of representation of LLVM.  The hope is that, by writing the Braid IR to provide a _minimal_ but _descriptive_ set of numeric operations, users will be able to explore optimiziations without sacrificing visibility of low-level computational operations.
```
sampsyo:
```suggestion
My goal for this project was to add floating points and fixed-length arrays to [Bril](https://github.com/sampsyo/bril).  The intention behind this decision is to promote Bril as a lower-level intermediate language, with fewer type abstractions than JavaScript without the complexity of representation of LLVM.  The hope is that, by writing the Bril IR to provide a _minimal_ but _descriptive_ set of numeric operations, users will be able to explore optimiziations without sacrificing visibility of low-level computational operations.
```

Braid IR… if only!

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+These goals were mostly successful at the level of Bril semantics, parsing, and interpretation.  What was more difficult than expected (and less successful) were my attempts to model TypeScript values and arrays with [TS2Bril](https://github.com/sampsyo/bril/blob/master/bril-ts/ts2bril.ts).  Before exploring the limitations of the Bril representation, however, we must examine the floating point and array semantics of Bril more formally.
+
+### Floating Points in Bril
+
+Bril now supports the types `double` and `float`.  These represent, respectively,  the stadard IEEE-754 [double-precision](https://en.wikipedia.org/wiki/Double-precision_floating-point_format) and [single-precision](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) floating-point formats.  These types come equipped with the basic arithmetic operations `fadd`, `fsub`, `fmult`, and `fdiv`, along with the comparitives `feq`, `flt`, `fgt`, `fle`, and `fge`.  The introduction of these operations is intended to highlight that floating-point and integer operations should be treated as fundamentally different objects.
```
sampsyo:
```suggestion
Bril now supports the types `double` and `float`.  These represent, respectively,  the stadard IEEE-754 [double-precision](https://en.wikipedia.org/wiki/Double-precision_floating-point_format) and [single-precision](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) floating-point formats.  These types come equipped with the basic arithmetic operations `fadd`, `fsub`, `fmult`, and `fdiv`, along with the comparisons `feq`, `flt`, `fgt`, `fle`, and `fge`.  The introduction of these operations is intended to highlight that floating-point and integer operations should be treated as fundamentally different objects.
```

I don't think people say "comparatives" for these very often?

Also, it sounds like `fadd`, etc. are overloaded to work on both floats and doubles, right? Maybe worth mentioning, perhaps at the beginning of the next paragraph.

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+In Bril, floating-point precisions should not be mixed.  It is expected, for instance, that the following code will produce a static exception:
```
sampsyo: Maybe better to just say "the following code is illegal" or something, without prescribing a way to signal the error.

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+```
+v0: double = const 5;
+v1: float = const 5;
+v2: float = fadd v0 v1
+```
+Similarly, mixing floating-points and integers in Bril is undefined, as well as using floating-point arithmetic operations on integers or vice-versa.  In a future project, I intend to explore the consequences of relaxing these semantic requirements, particularly when mixing floating-point precision.
```
sampsyo: Seems like a cool thing to do in the future might be to add arbitrary-precision floating point…

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Bril Fixed-Length Arrays
+
+Bril now supports types of the form `array=type[size]`, where `type = | int | bool | float | double | array`.  Note the inclusion of type recursion in this definition, which permits multi-dimensional arrays.  Arrays in Bril must contain elements all of the same type, and the size of arrays cannot be changed after initialization.
```
sampsyo: Seems like the type is just `type[size]`, right? What's up with the `array=` prefix?

sampsyo: Looks like there's an extra leading `|` in your list for `type`. Also, I might just say that `type` is any type, including an array, instead of listing all the currently-existing types.

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+There are three operations related to manipulating arrays in Bril: initialization, assignment, and indexing.  Arrays are initialized with the `new` instruction, which takes a type and stores a default initialization of that type to the given destination.  Array elements can be set with the `set` instruction, an effectful instruction which takes the array, an index, and a value in that order.  Finally, array elements can be read using the `index` instruction, which writes a value of the given array element.
```
sampsyo: For `new`, I assume that the type must be an array type? Or can it be any type?

sampsyo: I know this is bikeshedding, but because one of the operations is `set`, I probably would have called the other one `get` instead of `index`. :smiley:

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+Array operations are concisely summarized in the following code snippet:
+```
+val: int = const 5;
+ind: int = const 1;
+
+arr: int[3] = new int[3];   // arr = [0, 0, 0]
+set arr ind val;            // arr = [0, 5, 0]
+res: int = index arr ind;   // res = 5
+```
+Arrays can be written to freely in any Bril block.  As a result, array values are inherently stateful and should be treated by users carefully.  Writing values of the wrong type, however, is expected to not be permitted.
```
sampsyo: I'd just take a strong stance and say "not permitted," rather than hedging with "expected to not be permitted."

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Implementation
+
+Floating points and fixed-length arrays have been implemented as described, and can be reasoned about by the `bril2json` parser, the `Brili` interpreter, and the `bril2txt` translator.  A subset of TypeScript can be written to Bril-recognized JSON by the `ts2bril` compiler.  While these conversions work generally as described above, there are some features which bear special notice.
```
sampsyo:
```suggestion
Floating points and fixed-length arrays have been implemented as described, and can be reasoned about by the `bril2json` parser, the `brili` interpreter, and the `bril2txt` translator.  A subset of TypeScript can be written to Bril-recognized JSON by the `ts2bril` compiler.  While these conversions work generally as described above, there are some features which bear special notice.
```

Would you mind terribly much replacing Brili with brili when you refer to the command, in Markdown code backticks? The command is literally lower-case when you invoke it, after all…

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+The `Brili` interpreter supports integers, doubles, and floating-points all as different objects.  Integers are represented as Javascript [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) objects.  This choice of representation means that `Brili` does not correctly reason about integer overflow -- while this should be implemented in the future, I have not observed a need for it in the current Bril projects.
```
sampsyo:
```suggestion
The `Brili` interpreter supports integers, doubles, and floating-points all as different objects.  Integers are represented as JavaScript [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) objects.  This choice of representation means that `Brili` does not correctly reason about integer overflow -- while this should be implemented in the future, I have not observed a need for it in the current Bril projects.
```

sampsyo: Indeed; I haven't even specified yet what's supposed to happen on integer overflow!

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+Doubles can be represented simply as `number`, while floats are derived with the [Math.fround](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/fround) function.  Note that `fround` is only applied _after_ each floating-point operation; this is permitted due to the [IEEE-754 restrictions](https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html) on floating point operation precision loss (or lack of).  Extending Bril to support other mathematical operations (such as `pow` or `sqrt`), may require that this implementation choice be updated.
```
sampsyo:
```suggestion
Doubles can be represented simply as `number`, while floats are derived with the [Math.fround](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/fround) function.  Note that `fround` is only applied _after_ each floating-point operation; this is permitted due to the [IEEE-754 restrictions](https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html) on floating point operation precision loss (or lack thereof).  Extending Bril to support other mathematical operations (such as `pow` or `sqrt`) may require that this implementation choice be updated.
```

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+TypeScript represents every number as a `double` by default, thus operations in TypeScript are translated to Bril as floating-point operations.  The TypeScript `bigint` type can be used to specially generate integers in Bril.  Operations between TypeScript `bigint` values are not supported at the time of this writing; however, this should be changed shortly.  Floating points cannot be easily represented by TypeScript code; however, I will look into using the `fround` function as a mechanism for generating Bril `float` values.
```
sampsyo:
```suggestion
TypeScript represents every number as a `double` by default, and thus operations in TypeScript are translated to Bril as floating-point operations.  The TypeScript `bigint` type can be used to specially generate integers in Bril.  Operations between TypeScript `bigint` values are not supported at the time of this writing; however, this should be changed shortly.  Floating points cannot be easily represented by TypeScript code; however, I will look into using the `fround` function as a mechanism for generating Bril `float` values.
```

In the first sentence, those are independent clauses, so they need a conjunction (and a comma)—not just "thus." https://capra.cs.cornell.edu/styleguide/#however

sampsyo: > Operations between TypeScript `bigint` values are not supported at the time of this writing; however, this should be changed shortly.

What do you mean by this? I think `x + y` works if `x` and `y` are both `bigint`s—that's how the interpreter's integer values work. Maybe you mean when `x` is a `bigint` and `y` is a `number`?

If so, I'm not entirely sure this is relevant because, as you noted above, the types can't be mixed in operations.

sampsyo: > Floating points cannot be easily represented by TypeScript code; however, I will look into using the `fround` function as a mechanism for generating Bril `float` values.

I think the previous paragraph says that you've already done this?

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+Fixed-length arrays are reasoned about naively by the `Brili` interpreter; in particular, it makes no attempt to check array length before indexing.  `Brili` arrays are filled with 0s by default when initialized.
```
sampsyo: What if it's an array of a non-numeric type (e.g., an array of arrays)?

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+`ts2bril` can compile TypeScript arrays of consistent types to a `Bril`-like language.  This language differs from Bril only in that fixed-length arrays may have a size of `-1`, indicating that the size of the array is not known.  I am planning to explore ways to fix this issue; as TypeScript does not track array length and the Bril semantics for arrays is extremely limited, it should be very straightforward to take a second pass after compiling to replace each instance of unknown array length with the correct size.
```
sampsyo: To clarify, does the compiler _always_ emit -1 for the array size? Or are there circumstances when it can generate a real size?

I would actually posit that a second pass to deduce array sizes might actually be very hard to do. Especially if the array actually grows and shrinks, which TypeScript arrays can do!

So maybe the sensible thing would be to compile TypeScript tuples, which have fixed sizes, to Bril arrays. Or to add variable-length arrays to Bril.

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+I found working with `ts2bril` to be _extremely_ painful when adding arrays, and spent far too long trying to grapple with the TypeScript type system.  For me, this project highlighted the lack of internal documentation for TypeScript type structure.
```
sampsyo: > internal documentation for TypeScript type structure.

Perhaps you mean documentation for the TypeScript compiler? I think the type system itself has fine documentation—but the compiler implementation itself is quite inscrutable.

content/blog/2019-09-25-floats-static-arrays.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Evaluation
+
+Testing was focused on correctly interpreting several new files added to the Bril [test suite](https://github.com/Checkmate50/bril/tree/arrays/test).  These tests all work as intended; however, it is worth noting that the TypeScript array interpretation file was simplified somewhat as the limitations of the compiler became clear.  Speed of implementation was not a concern in this project.
+
+Some evaluation details of note include floating-point exactness testing and TypeScript compilation limitations.  Some simple floating-point operations [provided](https://github.com/Checkmate50/bril/blob/master/test/interp/float.bril) to `Brili` were compared to a similar C implementation ([compiled online](https://www.onlinegdb.com/online_c_compiler)), and found to be identical in precision.
```
sampsyo:
```suggestion
Some evaluation details of note include floating-point exactness testing and TypeScript compilation limitations.  Some simple floating-point operations [provided](https://github.com/Checkmate50/bril/blob/master/test/interp/float.bril) to `Brili` were compared to a similar C implementation ([compiled online](https://www.onlinegdb.com/online_c_compiler)) and found to be identical in precision.  
```

Not an independent clause, so no comma is necessary.

sampsyo (changes requested):
Sounds fantastic! Thanks for the detailed discussion of the language design. Here are some writing suggestions.

sampsyo:
Excellent! Nice work on this.

PR #55, opened 2019-10-02
Codestitcher blog post

splashofcrimson:
Codestitcher blog post (with @Blue9)


content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
++++
+title = "Codestitcher: inter-procedural basic block layout optimization"
```
sampsyo: We've been using Title Case for blog post titles so far.

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+extra.bio = """
+Eashan Garg is an undergraduate studying CS and English.
+Gautam Mekkat is an undergraduate studying CS and ECE.
+"""
+extra.author = "Eashan Garg and Gautam Mekkat"
++++
+
+## Introduction
+
+So far, we've covered traditional compiler optimizations such as dead code elimination, copy propagation, and constant propagation. These techniques are centered around *elimination*--- reducing the number of instructions in a program. This is useful, but programs have dramatically increased in size over the years. Eliminating unnecessary code can only do so much. Meanwhile, the capacities of memory units like L1 caches and the TLB haven't grown to account for larger program sizes. One reason for this is that as cache size increases, latency increases. Because of this, we want to optimally place instructions in caches, making sure the most commonly run code lives in the cache. Code layout is critical here. In this post we will discuss this paper's approach to code layout.
```
sampsyo:
```suggestion
So far, we've covered traditional compiler optimizations such as dead code elimination, copy propagation, and constant propagation. These techniques are centered around *elimination*---reducing the number of instructions in a program. This is useful, but programs have dramatically increased in size over the years. Eliminating unnecessary code can only do so much. Meanwhile, the capacities of memory units like L1 caches and the TLB haven't grown to account for larger program sizes. One reason for this is that as cache size increases, latency increases. Because of this, we want to optimally place instructions in caches, making sure the most commonly run code lives in the cache. Code layout is critical here. In this post we will discuss this paper's approach to code layout.
```

Do not put spaces around em dashes. https://capra.cs.cornell.edu/styleguide/#emdash

sampsyo: Because your audience for this post includes "the rest of the world," it's a little odd to say "we" have covered traditional compiler optimizations. Maybe clarify that you're talking about the lectures in the first few weeks of CS 6120? Or just leave out the whole "we" notion altogether?

sampsyo: > making sure the most commonly run code lives in the cache.

Not just that's in the cache, but that it's in the cache when it's time to execute it. Namely, that most accesses to load instructions hit in the L1 I-cache.

sampsyo: > optimally place instructions in caches

We unfortunately don't get to place anything in caches explicitly. Caches are implicit. All we can do is place things in "memory" (abstractly) nearby to each other and hope that this means the cache treats them well.

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+1. Generate an undirected call graph, where nodes are functions and edges represent calls from functions to other functions. More formally, a function A and a function B would have an edge between each other if A calls B or B calls A.
+2. Run a benchmark workload on this program, keeping track of what calls are made.
+3. Generate a weighted call graph, where edge weights are proportional to the frequency of calls between the two corresponding functions.
+4. Go through the edges in decreasing weight, joining the nodes connected by the edge. When nodes join, their edges combine, summing weights when necessary. Node names are also concatenated, denoting the corresponding function order.
+    - The node names concatenate to maximize the sum of the weights between consecutive nodes. This can involve reversing node names.
+6. After this process is complete, no edges will remain in the graph, and the order in which nodes are joined denote the function order.
+
+Let's go through a brief example. Suppose the weighted call graph is as follows.
+
+<img src="https://i.imgur.com/tvee3J0.png" width="400px">
```
sampsyo: Hotlinking Imgur-hosted images doesn't seem to work for me. They get 403 errors. (Did it work for you?)
Screen Shot 2019-10-03 at 3 15 57 PM

You can put images right here in the 6120 website. Just make you post a directory instead of a file, and call the body `index.md`. Then you can put images in the same directory.

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+Here, we obviously combine *A-B* and *C-D*, but recall the rule about concatenating node names. There are four ways to coalesce the nodes:
+- *A-B-C-D*: Here, *BC* has a weight of 30.
+- *A-B-D-C*: *BD* has a weight of 0 (no edge), so this is worse than the above ordering.
+- *B-A-C-D*: *AC* has a weight of 90, which is now the best ordering.
+- *B-A-D-C*: *AD* has a weight of 0.
+
+Based on the above, *B-A-C-D* is the optimal ordering, since it maximizes the weight between the coalesced nodes.
+
+### Additional optimization by PH
+PH also conducts some basic block reordering to maximize spatial locality of the program. This is explored further when discusses Codestitcher's basic block chaining stage.
```
sampsyo: In particular, it also does the thing where it splits a function _once_ into a "hot" part and a "cold" part, right? That seems like a big factor in PH that Codestitcher relaxes.

Specifically, this section seems to give the reader the impression that PH does not reorder individual basic blocks; it only works at the function granularity. That's not quite true.

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+## Codestitcher
+
+This brings us to Codestitcher, an optimizer that integrates basic block reordering and function reordering to improve upon the PH model. Below is a high-level overview of the steps taken in Codestitcher to generate the optimized code layout, given an input program.
+
+1. Compile the program to LLVM bitcode. This serves as the baseline program. As the program is built, symbols are emitted that allow us to generate a CFG for the program.
+2. Profile the program, generating weights for CFG edges.
+3. Using the weighted CFG, perform basic block chaining, ordering basic blocks in a way to minimize jumps.
+4. Perform hierarchical code collocation, ordering these basic block chains to form higher-level layouts. Here, the *d*-close partial layout algorithm works to build layouts to maximize spatial locality.
+
+There is not much to say about step 1, so let us explore steps 2, 3, and 4 in more detail.
```
sampsyo: TBH, I think this way they summarized the implementation was a little bogus. Indeed, who cares about compiling to LLVM? And I would even add, who cares about the step where you profile the CFG edges? _Obviously_ you have to do this to make any system like this work.

I think I would just say that this approach is going to work starting with an edge-profiled program, i.e., a CFG annotated with frequency counts on every edge. And that you can get that with profiling.

Then the only two steps that are actually part of Codestitcher are 3 and 4, the interesting ones.

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Profiling
+In order to build our weighted CFG for layout analysis, we need to collect data about "control flow transfers" (when control transfers between CFG nodes). Luckily, most Intel CPUs come with a "last branch record", or LBR--- storing recently executed instructions such as jumps, branches, and calls. With the Linux `perf` tool, this data can be gathered and used to generate a weighted CFG for optimizing program layout.
```
sampsyo:
```suggestion
In order to build our weighted CFG for layout analysis, we need to collect data about "control flow transfers" (when control transfers between CFG nodes). Luckily, most Intel CPUs come with a "last branch record", or LBR---storing recently executed instructions such as jumps, branches, and calls. With the Linux `perf` tool, this data can be gathered and used to generate a weighted CFG for optimizing program layout.
```

sampsyo: While nifty, I agree that this only deserves as much space as you give it. This discussion can probably be folded into the overview just to describe where the weighted CFG comes from.

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Basic Block Chaining
+As we reorder blocks, we will need to insert additional jumps to preserve the control flow. Therefore, we wish to order basic blocks in a way that minimizes these unconditional jumps. This is to prevent unnecessary instruction pointer moves during the execution of the program. We define a basic block chain as a directed sequence of basic blocks where no unconditional jumps occur except possibly in the last block. Generating a maximum-cardinality set of these chains would constitute the fewest number of unconditional jumps required. The authors denote this problem the *fall-through maximization problem*. This problem is the same as the maximum path cover problem, which is **NP-hard**. PH uses a greedy heuristic to approach this but does not provide a theoretical guarantee of how close to optimal the generated solution is.
```
sampsyo: > As we reorder blocks, we will need to insert additional jumps to preserve the control flow. Therefore, we wish to order basic blocks in a way that minimizes these unconditional jumps. This is to prevent unnecessary instruction pointer moves during the execution of the program.

This is part of the problem setup for _any_ code layout technique—so if you need to make this point, make it earlier.

sampsyo: > where no unconditional jumps occur except possibly in the last block.

In fact, no jumps at all (conditional or unconditional) can appear anywhere but at the end.

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+There *is* is a 1/2-approximation algorithm for this problem guarantees that the weight of the path cover is within *1/2* of the optimal solution. While its performance does not absolutely exceed that of PH's heuristic, Codestitcher uses a hybrid of PH's method and the approximation algorithm, which provides better results than either algorithm run individually.
```
sampsyo: This was a weird thing about the paper: the improvement they claim from this 1/2-approximation algorithm is 0.03%, which is basically nothing (they don't even quantify the variance, but I'd expect this to be in the noise). For that reason, I think their whole 1/2-approximation algorithm is just a distraction. It doesn't deserve too much attention. The more important part is the whole layout distance thing (and the free-form reordering of basic blocks from different functions).

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Hierarchical Code Collocation
+
+Coming up with a framework to integrate intraprocedural and interprocedural ordering optimizations while taking *layout distance* (the distance between two instructions) into account is difficult. We will first try to formalize this problem.
+
+#### *d*-Close Partial Layout
+
+It would help if we could formally define what layout distance really means. The paper's definition is quite intuitive: the number of bytes between two basic blocks. To take advantage of spatial locality, we want to ensure that related instructions that are executed frequently have as few bytes between them as possible. (This is because the instruction cache can only hold so many instructions.)
```
sampsyo: FWIW, I think this is the central insight of the paper: that basic block layout should be based on minimizing the byte distance between pairs of blocks that are frequently executed together.

sampsyo: > (This is because the instruction cache can only hold so many instructions.)

It is also because caches benefit from spatial locality because they're organized into cache lines. (That is, when you bring in one byte of a cache line, you get the rest of the line "for free," so it behooves you, as the compiler, to put frequently-used-together instruction within a given cache line. But you can't know for sure what the line size/alignment is, so it's almost as good to just generally put things as close together as you can.)

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+Given a specific distance value *d*, we can try to lay out our program to maximize the number of times control is transferred between two basic blocks within *d* bytes of each other. However, this is not enough. We need to optimize over many different *d* values to account for hierarchical memory systems with multiple cache levels. Our goal is to find the *finest grain* layout that maximizes control transfers across all combinations of basic blocks, across multiple distances.
+
+How do we solve this? The paper suggests a greedy approach that encodes the problem as a graph of basic block sequences, with edges weighted by the number of control flow transfers within *d* bytes. Nodes of the graph are iteratively coalesced in order of descending weight until no edges remain, and the remaining node's ordering is output.
```
sampsyo: > and the remaining node's ordering is output.

I don't quite understand this part of the sentence—what is the "remaining node"? Maybe you mean plural "nodes"? If so, what is their ordering? If no edges are left, there isn't an ordering anymore, right?

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Finishing the Codestitcher layout optimization
+
+Using the *d*-close partial layout results, Codestitcher orders the final basic block chains based on a ratio of their total execution counts and size in bytes. This layout is then used to recompile an optimized version of the original program.
```
sampsyo: > a ratio of their total execution counts and size in bytes.

This is pretty surprising—can you explain a little more about why that ratio is expected to be meaningful?

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+| Application | Instruction cache MPKI | Instruction TLB MPKI |
+|-------------|------------------------|----------------------|
+| MySQL       | 62.44                  | 9.35                 |
+| Clang       | 8.14                   | 1.01                 |
+| Firefox     | 9.16                   | 1.54                 |
+| PHP Server  | 7.63                   | 0.96                 |
+| Python      | 3.40                   | 0.19                 |
+
+As expected, when the load on both the cache and TLB increase, code layout optimization is more effective--- this is evidence that these optimization methods in general are taking advantage of the processor's memory units to improve performance. However, we are not convinced that this is enough evidence to conclude that Codestitcher is superior to PH's method, especially because this is only evaluated on one specific processor.
```
sampsyo:
```suggestion
As expected, when the load on both the cache and TLB increase, code layout optimization is more effective---this is evidence that these optimization methods in general are taking advantage of the processor's memory units to improve performance. However, we are not convinced that this is enough evidence to conclude that Codestitcher is superior to PH's method, especially because this is only evaluated on one specific processor.
```

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Other Metrics
+
+While a high-level analysis of the performance of Codestitcher is helpful, the diversity of methods evaluated allows a deeper dive into the specific claims that govern Codestitcher.
+
+1. Function reordering as applied to PGO had little to no impact on performance, with a maximum improvement of 4.1% for instruction cache miss rates. The paper claims, that block-based reordering show sisgnificant improvement here, but use the performance of CS as a point of comparison. We were somewhat wary of this - perhaps comparing PGO-PH and PGO with some sort of block-based reordering technique would be a more fair analysis of their differences in performance.
```
sampsyo:
```suggestion
1. Function reordering as applied to PGO had little to no impact on performance, with a maximum improvement of 4.1% for instruction cache miss rates. The paper claims that block-based reordering shows sisgnificant improvement here but uses the performance of CS as a point of comparison. We were somewhat wary of this---perhaps comparing PGO-PH and PGO with some sort of block-based reordering technique would be a more fair analysis of their differences in performance.
```

content/blog/2019-10-02-codestitcher.md
```diff
@@ -0,0 +1,166 @@
...
+
+Ultimately, this paper explores a very interesting technique for optimizing code layout, but its evaluation does not do a great job of convincing us that Codestitcher provides substantial benefits over other code layout optimizers.
+proach that encodes the problem as a graph of basic block sequences, with edges weighted by the number of control flow transfers within *d* bytes. Nodes of the graph are iteratively coalesced in order of descending weight until no edges remain, and the remaining node's ordering is output.
```
sampsyo: The last line looks duplicated from elsewhere in the post.

sampsyo (changes requested):
Nice work, both of you! Here are some suggestions and questions.

content/blog/2019-10-02-codestitcher.md
Blue9: When two nodes are coalesced, the new node stores an internal ordering of the two original nodes. For example, if node A and node B were coalesced, the resulting node would store either A-B or B-A. Since we coalesce until no edges remain, we'll end up with one node, which will contain the final function ordering. I've updated this portion to clarify.

sampsyo:
Looks great! I merged this and fixed some of your CSS syntax (it's `width: 100%`, not `width=100%`).

PR #76, opened 2019-10-07
Blog Post: Compiler Optimizations for Improving Loop Locality

Chillee:


content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+actually 1-dimensional arrays in memory, the cache lines must align along a
+particular axis.
+
+Now, compare the 2 access patterns (the blue line and the red line). The blue
+access pattern loads a single cache line and then accesses all of its
+elements before moving onto the next cache line. The red access pattern,
+however, loads a single cache line, accesses a single element, and then moves
+onto the next cache line.
+
+<img src="cache_lines.jpeg" style="width: 100%">
```
sampsyo: This diagram is awesome!!

content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+of canonicalization, and can also allow for other loop optimizations to be
+applied. In their paper, loop reversal didn’t improve data locality in any of
+their benchmarks.
+
+#### Loop Fusion
+Loading in each cache line takes a significant amount of time. If we have
+multiple loops, we run into the possibility that we will load a single cache
+line multiple times, wasting time.
+
+For example, take (online example can be found [here](http://ideone.com/OnbRXU)).
```
sampsyo:
```suggestion
For example, take this code (an online example can be found [here](http://ideone.com/OnbRXU)):
```
FWIW, I think it's nicest to use a colon before inline examples, such as code listings.

content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+```
+for (int i = 0; i < MAXN; i++)
+        A[i] += j;
```
sampsyo:
```suggestion
    A[i] += j;
```

content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+for (int i = 0; i < MAXN; i++)
+    A[i] *= j;
+```
+It’s easy to see that this code is performing redundant cache line loads. We
+load the cache line that `A[0]` belongs to twice - once in the first loop and
```
sampsyo:
```suggestion
load the cache line that `A[0]` belongs to twice—once in the first loop and
```
https://capra.cs.cornell.edu/styleguide/#emdash

content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+```
+Thus iterating in the optimal order for both arrays.
+
+## Cost Model
+
+For class discussion, think about the following algorithm:
+
+<img src="cost_model.png" style="width: 100%">
+
+They use this cost model for determining the optimal sequence of loop fusion/fission/permutation to apply.
```
sampsyo: Maybe it's worth briefly mentioning the motivation—why is it important to find the right sequence? And why is a cost model the right approach to doing it?

content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+
+# Future Work
+The loop optimizations presented here are still used everywhere. However,
+these are perhaps the more straightforward optimizations to make. In
+particular, they don't deal with any optimizations (besides loop reversal)
+that change the structure of a loop. For example, unrolling (and the
+vectorization opportunities that poses), as well as tiling.
```
sampsyo: The last sentence is a fragment—consider joining it with the previous sentence or adding a verb to say more about why you're mentioning those examples.

content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+
+Incorporating these optimizations in an automated cost model is an ongoing
+problem. Several frameworks for reasoning about these optimizatons have been
+proposed (polyhedral optimization, Halide).
```
sampsyo: This brief mention of related work could also use a tiny bit more context. How exactly is the polyhedral model relevant to these optimizations? And Halide doesn't by itself give you a cost model—but it does of course make it convenient to apply similar optimizations.

content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+
+In addition, more cost models have been proposed for handling these
+optimizations as well.
```
sampsyo: This is a bit vague—if they've been proposed, can you refer to them specifically?

sampsyo (changes requested):
Looking good so far! I really like your example-driven approach to explaining the optimizations that the paper proposes.

I've made a few writing suggestions, and I think you could spend a bit more time on the tail end of the post—both on the cost model (the section currently just says "think about it," but some motivation to do so would not go awry) and on the future work (which could more solidly connect to the papers & lines of work that it mentions).

content/blog/2019-10-09-loops/index.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+B. Row-Major
+```
+int A[DIM1][DIM2];
+for (int iter = 0; iter < iters; iter++)
+    for (int j = 0; j < DIM2; j ++)
+        for (int i = 0; i < DIM1; i++)
+            A[i][j]++;
+```
```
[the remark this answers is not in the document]
Chillee: I swear, they were different when I wrote this blog post :sweat_smile:

content/blog/2019-10-09-loops/index.md
Chillee: Completely thanks to @g-c-c. Their notes are also a wonder to look at.

Chillee:
Responded to comments, and updated the future work section by a significant amount.

I'm having a bit of a hard time coming up with a good tiling code example. My example makes sense to me in theory :thinking: , but it's not actually faster in practice.

sampsyo:
Excellent work!

Indeed, it seems pretty hard to get small examples to work in general here (although I don't have an explanation for why tiling would be particularly difficult). It's awesome you got any to work at all! In the presence of prefetchers in modern CPUs, you get get surprisingly good performance out of surprisingly bad-looking code.

PR #77, opened 2019-10-09
Update Loops blog post with new figures.

Chillee:
@g-c-c drew some more figures! It would be a shame not to include their figures in the blog post, so this PR.

There's also some minor formatting changes.


sampsyo:
Awesome; thank you for the update, both of you!

PR #78, opened 2019-10-16
Blog Post: double-checked locking

0x486F626F:


content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+
+
+### Volatile in JDK5
+
+Java introduced "volatile" since JDK5. 
+A "volatile" makes the variable be stored and accessed in main memory.
+Every read of a volatile is from main memory.
+Every write of a volatile is to main memory.
+
+The "volatile" also provides a "happen-before" guarantee:
```
sampsyo: Hmm, I don't think I'd use the term "happens-before"… it's too intimately tied to the Lamport ordering notion. https://lamport.azurewebsites.net/pubs/time-clocks.pdf

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+                    helper = new Helper();
+            }
+        }
+        return helper;
+    }
+}
+```
+
+However, since all read and write operations of a volatile variable need to
+access main memory, which can be very slow.
```
sampsyo: Again, it's not just that they "access main memory"—they also use special hardware support for atomic/synchronized accesses (i.e., the compiler inserts barriers anyway), so nothing is really gained.

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+                if (h == null) {
+                    h = new Helper();
+                    helper = h;
+                }
+            }
+        }
+        return h;
+    }
+}
+
```
sampsyo: Extra blank line.

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+```
+
+In cases that the `helper` in already initialized, 
+   this optimization can reduce one volatile read by returning the local variable.
+
+### Explicit Memory Barrier
+
+The last fix with two synchronized sections does not work because releasing lock
+is an implicit "one-way" memory barrier.
```
sampsyo: Do you mean the _first_ fix? I think the _last_ fix, just above, actually does work…

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+* `std::atomic_thread_fence(std::memory_order_release)` guarantees that 
+read/write before a memory barrier cannot be reordered after the memory barrier.
+
+Similar to the Java volatile case, 
+the initialization must be done before it stores `h` to `helper` 
+with the memory barrier.
+
+### Static Singleton
+
+If the `helper` is static, i.e. all the instances of class `Foo` shares the
```
sampsyo:
```suggestion
If the `helper` is static, i.e., all the instances of class `Foo` share the 
```

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+which is a safe and efficient concurrent lazy initialization 
+for all Java versions.
+
+* **Lazy Initialization**: as specified by the Java language,
+    the static class `HelperSingleton` is not initialized until the first it is
+    referenced.
+* **Safe**: the class initialization is guaranteed to be synchronized by 
+    Java language.
+* **Efficient**: all subsequent calls of `getHelper` will return the initialized
+    `helper` without any synchronization overhead.
```
sampsyo: Wow—I'm not sure I understand this deeply Java-specific trick. You say that class initialization is guaranteed to be "synchronized," but synchronized _with what_?

That is, regardless of how it goes about it, a safe version of this will need to somehow use hardware synchronization to order the initialization with the use. So it does not suffice just to have all initializations be synchronized _with each other_. How does the JVM avoid acquiring a lock on subsequent accesses to `HelperSingleton.helper`?

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+### Thread Local
+
+`ThreadLocal` is a variable that each thread will have its own copy of its value.
+Each thread can only access and modify its own copy of a thread local variable
+independently of other threads.
+
+A thread local can be used to maintain the state of "whether the state has gone
+through the synchronized initialization". 
+Since if a thread has gone through the synchronized initialized once,
+it can be confident about that object is already initialized.
```
sampsyo:
```suggestion
it can be confident that that object is already initialized.
```

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+
+### Thread Local
+
+`ThreadLocal` is a variable that each thread will have its own copy of its value.
+Each thread can only access and modify its own copy of a thread local variable
+independently of other threads.
+
+A thread local can be used to maintain the state of "whether the state has gone
+through the synchronized initialization". 
+Since if a thread has gone through the synchronized initialized once,
```
sampsyo:
```suggestion
If a thread has gone through the synchronized initialization once,
```

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+it can be confident about that object is already initialized.
+
+Inside the synchronized initialization section, 
+only the first thread will find the object is `null` and initialized object.
+All threads will change their per-thread state, 
+    so that they will not enter the synchronized section again.
```
sampsyo: Maybe it's worth expanding on the fact that _every_ thread will then execute the critical section _exactly once_? That's clearly more costly than the "ideal" double-checked locking pattern, if it were to work. But it's not much worse, so that's probably fine.

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+It analyzed why some intuitive solutions do not work,
+   and also analyzed some working solutions.
+
+Writing multi-threaded program is hard. 
+Writing correct and safe multi-threaded program is even harder.
+When analyzing the correctness of multi-threaded programs,
+it requires the considerations of multiple components, 
+including compilers, systems, and processors.
+On the other hand, when designing compilers, systems, or processors,
+it also needs to take consideration of common used design pattern.
```
sampsyo:
```suggestion
one also needs to take into consideration commonly used design patterns.
```

sampsyo (changes requested):
Hi, Hongbo! This is a good start. Please see some comments inline—in particular, I think it would be really useful to focus a bit more on explaining clearly why the original version does _not_ work. The discussion of the solutions is interesting but less important than helping people completely "get" why the intuitive version is broken because of compilers and processors.

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,353 @@
...
+        }
+        return helper;
+    }
+}
+```
+
+### 32-bit Primitives
+
+If the initialized value is a 32-bit primitive value, such as `int` or `float`,
+   the double-checked locking will work.
```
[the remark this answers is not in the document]
0x486F626F: Maybe I just don't include this in my post, I don't quite like/understand it either...

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,407 @@
...
+Intuitively, we may come up with the following sequence to do this job:
+
+1. Check if the object is initialized without locking.
+If it is, then return the object immediately.
+2. Acquire the lock and check again if the object is initialized.
+If another thread has grabbed the lock before, 
+   the current thread can see the object is created, and return the object.
+3. Otherwise, the current thread will create the object and return.
+
+With the guideline above, we will get a code like following:
```
sampsyo:
```suggestion
With the guidelines above, we will get code like the following:
```

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,407 @@
...
+                    ptr.field1 = initField1();
+                    ptr.field2 = initField2();
+                    helper = ptr;
+                }
+            }
+        }
+        return helper;
+    }
+}
+```
```
sampsyo: Inlining the initialization method is a really good idea! This makes it clearer.

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,407 @@
...
+
+In order to improve overall performance,
+some compilers, memory systems, or processors may reorder the instructions,
+like moving `helper = ptr` before initializing fields of the object.
+
+```Java
+if (helper == null) {
+    synchronized(this) {
```
sampsyo: Maybe include the "outer" `if` check here for consistency?

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,407 @@
...
+            ptr.field2 = initField2();
+        }
+    }
+}
+return helper;
+```
+
+This reordering is legal, since there is no data dependency between `helper = ptr`
+and the instructions for initializing fields. 
+However, this reordering could let other thread access uninitialized fields of the object.
```
sampsyo:
```suggestion
However, this reordering could let other threads access uninitialized fields of the object.
```

sampsyo (commented):
Nice! This works really well; I think this will really help people understand. Good work.

sampsyo (changes requested):
This is looking really great! I have a few proofreading comments, and also two points where I still don't quite understand the technical argument behind a couple of the "fixes."

content/blog/2019-10-18-double-checked-locking/index.md
```diff
@@ -0,0 +1,491 @@
...
+
+#### 32-bit Primitive Variables
+
+Read and write operations of most primitive variables 
+(except `long` and `double` since they are 64-bit)are atomic.
+If the initialized value is a 32-bit primitive variable,
+   assignment to the variable will only happen once the data is available.
+Since the write operation is atomic, 
+other threads will either see a ready-to-use value or 0.
+There is no middle state of "initializing".
```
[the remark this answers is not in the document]
0x486F626F: Ah I see the point. Since it is not `private volatile int magicNumber`, but it somehow behaves like a volatile variable, so it needs some specific memory model support from Java.

sampsyo (approved):

sampsyo:
Very nice work, Hongbo! Published.

PR #84, opened 2019-10-24
Loop Unswitching Optimization

### LINKED ISSUE #61: Project 2 Proposal: Loop Unswitching
**What will you do?**
Implement the loop unswitching optimization as described  
[on this wiki](https://en.wikipedia.org/wiki/Loop_unswitching).    I will implement this using adaptive software development principles so it can be easily extended to include other loop optimizations.

**How will you do it?**
I will first create a for loop abstraction that can be used in the SSA intermediate form.  I will then follow the algorithm described in [this paper](https://www.cs.rit.edu/~mtf/student-resources/20155_surawski_mscourse.pdf).  The majority of my effort will be in finding conditional branch blocks on invariant values within a loop's body.  I may make slight modifications to allow for multiple nested loops and conditions (nested conditionals that are independent of their parent loop may need not to be unswitched). 

Relevant link:  http://www.ucw.cz/~hubicka/papers/proj/node10.html

To measure success, I will email the authors in the aforementioned paper to acquire their testing benchmark (no response).  Most empirical success is based off of the success of this project.  Since unswitching allows for the possibility of separate loop optimizations, implementing this allows other optimizers to optimize components that may never be hit.  Another way to test this, as it has been pointed out, is to count the number of branches.

Edited with suggestions from below

**Team members:**
Solo

sameerlal:
closes #61


content/blog/2019-10-23-loop-unswitching/index.md
```diff
@@ -0,0 +1,252 @@
...
+    contents[if_for_loop_logic] = contents[loop_logic]
+
+    else_for_loop_logic = loop_logic  + hash(else)
+    contents[else_for_loop_logic] = contents[loop_logic]
+```
+   
+    
+Now we are ready to create the loop body blocks.  Here, the contents should include instructions that dominates the ```conditional t``` block and is also dominated by the big loop block.  Here, we use set intersections in the domination dictionary.
+    
+Now, as an additional optimization, we reorder these blocks according to the original CFG.  I will delegate the details of this for those interested to the linked code above.
```
sampsyo: Aha! This answers my question from above. But I'm actually still confused about one thing: you emphasize below that the correct order needs to be preserved for correctness, but here you say that the block order is "an additional optimization." These seem to be in conflict—would the program still work correctly if you didn't carefully do the ordering? If not, then I don' think I'd call it an optimization…

content/blog/2019-10-23-loop-unswitching/index.md
```diff
@@ -0,0 +1,252 @@
...
+```
+b = True
+for i in range(10):
+    for j in range(10):
+        if b then:
+            do something()
+        else:
+            do something else()
+```
+will not be completely unswitched since the conditional statement lies inside of a nested for loop.  Instead, the conditional will be unswitched outside of the first for loop, which only midly decreases the numbeer of branches.  Implementing this is not too difficult to do, though it requires  additional abstractions for nested for loops.
```
sampsyo:
```suggestion
will not be completely unswitched since the conditional statement lies inside of a nested for loop.  Instead, the conditional will be unswitched outside of the first for loop, which only midly decreases the number of branches.  Implementing this is not too difficult to do, though it requires  additional abstractions for nested for loops.
```
Please run spell check.

sampsyo: Would it get moved out of the outer loop if you ran your optimization twice?

content/blog/2019-10-23-loop-unswitching/index.md
```diff
@@ -0,0 +1,252 @@
...
+The results of this optimization were quite interesting.  I primarily evaluated the code on percentage difference in the number of branches.  To do this, I modified the Bril Typescript interpreter to keep track of the number of branches for each run.  For a program:
+```
+    for _ in range(maxiter):
+        prebody
+        if b:
+            ifbody
+        else:
+            elsebody
+        postbody
+```
```
sampsyo: Again, no need for both indentation and fences.

content/blog/2019-10-23-loop-unswitching/index.md
```diff
@@ -0,0 +1,252 @@
...
+            prebody
+            ifbody
+            postbody
+    else:
+        for _ in range(maxiter):
+            prebody
+            ifbody
+            postbody
+```
+the improvement comes from eliminating branches in the "if...else" decision.  Through testing, one loop unswitching produces a decrease in the number of branches of approximately 14%, on average.
```
sampsyo: On average across what? It sort of looks like this is just one program, so I would have expected there to be only one number for "reduction in branches." Or maybe you use multiple benchmarks? That would be cool to know!

content/blog/2019-10-23-loop-unswitching/index.md
```diff
@@ -0,0 +1,252 @@
...
+            postbody
+    else:
+        for _ in range(maxiter):
+            prebody
+            ifbody
+            postbody
+```
+the improvement comes from eliminating branches in the "if...else" decision.  Through testing, one loop unswitching produces a decrease in the number of branches of approximately 14%, on average.  
+
+This result is highly dependent on the structure of the program.  For instance, after unswitching, ```ifbody``` and ```elsebody``` can separately be optimized which would compound the results even further.  The results from the benchmark tests assume that the loops CANNOT be independently optimized, so our claimed average can be thought to be a lower bound on the decrease in the number of branches.
```
sampsyo:
```suggestion
This result is highly dependent on the structure of the program.  For instance, after unswitching, ```ifbody``` and ```elsebody``` can separately be optimized which would compound the results even further.  The results from the benchmark tests assume that the loops *cannot* be independently optimized, so our claimed average can be thought to be a lower bound on the decrease in the number of branches.
```

We have italics, so no need to yell. :smiley:

content/blog/2019-10-23-loop-unswitching/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+In literature, benchmarking code unswitching involves measuring the increase in code size.  In my implementation, code size roughly doubled for each unswitching which is consistent with the literature.
+
+Overall, this project was a success, and the natural extension is to extend this optimization to work for nested loops.
+
+##### Evaluation 
+
+This optimization was tested on many programs both on defined behavior and undefined behavior.  The testing suite I generated tests on the following (though not limited to) attributes:
```
sampsyo: What do you mean by "undefined behavior"? I wasn't aware Bril had any.

content/blog/2019-10-23-loop-unswitching/index.md
```diff
@@ -0,0 +1,252 @@
...
+```
+the improvement comes from eliminating branches in the "if...else" decision.  Through testing, one loop unswitching produces a decrease in the number of branches of approximately 14%, on average.  
+
+This result is highly dependent on the structure of the program.  For instance, after unswitching, ```ifbody``` and ```elsebody``` can separately be optimized which would compound the results even further.  The results from the benchmark tests assume that the loops CANNOT be independently optimized, so our claimed average can be thought to be a lower bound on the decrease in the number of branches.
+
+In literature, benchmarking code unswitching involves measuring the increase in code size.  In my implementation, code size roughly doubled for each unswitching which is consistent with the literature.
+
+Overall, this project was a success, and the natural extension is to extend this optimization to work for nested loops.
+
+##### Evaluation
```
sampsyo: Please move the experimental setup section before the results section.

content/blog/2019-10-23-loop-unswitching/index.md
```diff
@@ -0,0 +1,252 @@
...
+##### Evaluation 
+
+This optimization was tested on many programs both on defined behavior and undefined behavior.  The testing suite I generated tests on the following (though not limited to) attributes:
+* Should a conditional be unswitched?
+* Simple/Complex blocks before/after conditional
+* Non-unswitched branches before conditional to be unswitched
+* Blocks with no terminating jump instructions
+* Nested if/else conditionals should never be unswitched
+
+I would be very happy with suggestions for additional test cases or pull requests that test this optimization on more test cases.  Furthermore, for questions or comments on design, please reach out by e-mail.
```
sampsyo: It's hard to to email you without your email address or some link thereto. :smiley:

sampsyo (changes requested):
This seems great so far! One thing that's important to add is a little more detail on the programs you used for your evaluation. I think you used several hand-written (i.e., contrived) benchmarks—giving some statistics on how big and how complicated these are would be useful. It might also be worth explicitly acknowledging that you didn't try to use real-world code, probably because there are not that many real-world Bril programs out there.

content/blog/2019-10-23-loop-unswitching/index.md
sameerlal: Yeah, when implementing I decided to only unswitch one loop at a time.  I was trying to plot a curve something along the lines of "% decrease in branches" vs "# of unswitched loops". I'll clarify to address this.

PR #86, opened 2019-10-24
STOKE Blog Post

### LINKED ISSUE #62: Project 2 Proposal: Stochastic Superoptimizer
**What will you do?**
I will implement a minimal stochastic superoptimizer modeled after STOKE (http://stoke.stanford.edu/). The goal is to do a random search over alternate equivalent programs, and evaluate them per some cost metric to generate an optimized representation of the source code.

Inspiration: https://blog.acolyer.org/2017/03/30/stochastic-program-optimization/

**How will you do it?**
For simplicity, the source (STOKE) targets loop-free code, so I will do the same. My intended approach is as follows:
1. Build infrastructure to apply perturbations to the original program and generate a sample space of similar programs to explore/evaluate (remove, add, and adjust instructions).
2. Evaluate correctness of the new programs. The original STOKE paper uses a two-step process to validate equality of two x86 assembly sequences (The original program X and the perturbed program Y).
    - Assert that X and Y produce identical outputs given a series of live inputs. Since I might want to generate programs on the order of thousands in step one, this allows the superoptimizer to quickly disregard programs that are incorrect.
    - If X and Y produce identical outputs, pass X and Y to an SMT solver (Something like https://github.com/Z3Prover/z3) and verify that they are indeed the same.
3. Write a cost model to quantify the performance of each program (Wall clock time/Some other cost metrics)
4. Use an MCMC technique like the Metropolis-Hasting algorithm to select sample programs based on some approximate probability density function.
5. Output a list of programs and scores sorted in accordance with the cost model.

Given a list of valid representations of the original program, I will take the best 5 - 10 and use them as a basis for my evaluation. There should already be a theoretical guarantee of correctness, so I will focus primarily on performance.

I will put together a test suite of loop-free programs to run through the superoptimizer. The output code sequences will then be evaluated against two metrics: static instruction count and actual execution time of the Bril IR on the interpreter. I can do a similar evaluation on both the original Bril program and a Bril program optimized using more traditional techniques (DCE, Copy prop, Const fold, etc.). The cost model itself will have a large effect on how optimal programs are selected, so a cool stretch goal would be to test a few different cost metrics and evaluate the results.

As an extra sort of evaluation, I'm really curious to see if the superoptimizer is capable of uncovering these more 'traditional' optimizations through random search. I'll do a more qualitative sweep of the programs that are output and see what techniques it manages to discover!

**Team members:**
Myself!

splashofcrimson:
Stoked! Closes #62


sampsyo (changes requested):
Nice work! I have a few conceptual questions within and some syntax suggestions.

PR #87, opened 2019-10-24
Project 2: Profile-Guided Code Layout

### LINKED ISSUE #63: Project 2 Proposal: Profile-Guided Function Reordering
**What will you do?**
I will implement profile-guided procedure reordering. I will use my project 1 extension, [Bril()](https://github.com/sampsyo/cs6120/issues/8), to allow for function calls and variable input data (via command-line arguments). If time permits, I will extend this optimization to operate on basic blocks. However, the main focus will be on function ordering.

**How will you do it?**

1. Run each sample input in the profiling data (provided by the programmer) through the program and track the function calls made during program execution.
2. Generate a weighted directed function call graph, where nodes denote functions and edges denote function calls (from tail to head) each weight corresponds to the frequency of the call (calculated in previous step).
3. Iterate over the edges in order of decreasing weight and coalesce the edge's nodes until no edges remain.

This closely follows the approach outlined in [Optimizing Function Placement for Large-Scale Data-Center Applications](https://research.fb.com/wp-content/uploads/2017/01/cgo2017-hfsort-final1.pdf). I chose this approach because it improves upon the [traditional function placement technique introduced by Pettis and Hansen](https://dl.acm.org/citation.cfm?id=93550), in that it uses directed call graphs instead of undirected call graphs.

I will run various Bril programs with large numbers of functions that call each other. Also, the linked papers above contain some interesting function call graphs that I can model in Bril and validate using the papers' results.

The primary goal of function reordering is to improve spatial locality of code. Specifically, we want (1) functions to live close together in the program if they often execute consecutively, and (2) "hot" code sections to be close to each other to improve instruction cache utilization.

Measuring goal 1 is straightforward: we can use the total distance (in lines) traversed by the instruction pointer (which in our case is the index of the currently executing instruction in the interpreter). The more the instruction pointer moves, the poorer the spatial locality of a program. If we add line numbers to the instruction's JSON representation and keep track of the line number changes between consecutively run instructions, we can determine the total number of lines traversed by the instruction pointer when running the program.

Evaluating the second goal is more challenging. Since Bril is run on TypeScript, its instructions do not directly live in the instruction cache, making it difficult to measure the instruction cache utilization. I will try to use [Cachegrind](http://valgrind.org/docs/manual/cg-manual.html) to profile instruction cache utilization, but I am not sure if it is compatible with Bril. In the case that it does not work, I do not think it is a major problem, as the first goal will correlate with the second. In other words, by decreasing the total distance traversed by the instruction pointer, we improve spatial locality, and instruction cache utilization will increase.

**Team members:**
None.

Blue9:
Closes: #63

Implementation: https://github.com/sampsyo/bril/pull/35


sampsyo (changes requested):
Extremely nice work! Your evaluation is very complete, and it was easy to understand how your entire workflow works. I just have a few small writing suggestions here.

sampsyo (approved):
Looks great!!

PR #88, opened 2019-10-24
Project 2: Loop-Level Automatic Vectorization

voorka:


content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+  va: int = vload ai;
+  vb: int = vload bi;
+  vc: int = vadd va vb;
+  vstore vc ci;
+
+  i: int = add i four;
+  done: bool = ge i size;
+  br done vvadd_done vvadd_loop;
+```
+The second loop executes for a forth the number of iterations of the first loop while behaving identically by using `vload`, `vstore`, and `vadd` that operate on four array elements at a time. This allows for *i* to be incremented by four each iteration instead of by one.
```
sampsyo: forth -> fourth? Overall, however "a fourth the number of iterations" is somewhat awkward phrasing, in case you want to try another tactic.

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+  vc: int = vadd va vb;
+  vstore vc ci;
+
+  i: int = add i four;
+  done: bool = ge i size;
+  br done vvadd_done vvadd_loop;
+```
+The second loop executes for a forth the number of iterations of the first loop while behaving identically by using `vload`, `vstore`, and `vadd` that operate on four array elements at a time. This allows for *i* to be incremented by four each iteration instead of by one. 
+
+For this project, we designed and implemented automatic loop vectorization by converting serial operations on array elements to their vector counterparts. We build on  Philip Bedoukian's work that brings [vector instruction support into the Bril interpreter](https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/interpreter-vector-support/). We use his array implementation and the vector operations that he provides which operate on vectors of length four. We did not use his C++ dynamic library, and instead wrote a new dynamic library in Rust with a wider range of function calls as we were more familiar with the Rust specification. In addition, this choice proved to be a valuable learning opportunity.
```
sampsyo: Awesome work reusing existing language extensions!!

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+Now that we are able to find a loop, it must have the following elements to be vectorizable:
+- A branch statement (to exit the loop)
+    - e.g., `br done exit loop_header;` 
+- A condition variable (boolean variable for the branch argument)
+    - e.g., `done: bool = eq i size;`
+- An *induction variable* that increases or decreases each iteration
+    - e.g., `i: int = add i one;`
+
+These are the minimum requirements for creating a loop that iterates a set number of times. To actually get these values, we need to enforce that there is only one branch statement in the loop. We do not handle loops with multiple branches in this project as it is much more complex to hadle vectorization for nested or loops where some statements run more iterations than other statements. We also restrict the induction variable, condition variable, and branch statement to be the last three instructions in the loop, which
```
sampsyo: Incomplete sentence at the end there?

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+From these, we can find more information such as the variable that specifies the bound, which can be deduced from the arguments for the condition variable as one argument must be the induction variable, leaving the other to be the bound. 
+
+We now want to verify properties to ensure that this loop is indeed vertorizable. For example, we need to know ensure that the bound variable is a constant and not something loaded from memory, or we won't be able to determine how many times the loop is executed. This requires us to know the latest definition of variables, which we get using a reaching definitions analysis. It is possible to find that statement walking backwards through the cfg, but we found using a dataflow analysis to be a much cleaner approach. Furthermore, we'd also like to know values of variables at different program points, such as the base pointers for array loads or stores, which we can get through a copy propagation analysis.
```
sampsyo: "know ensure" is redundant—pick one?

sampsyo: Capitalize "cfg."

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Validity Checks
+Now that we have reaching definitions and copy propagation per statement, we can check whether a loop is vectorizable.
+
+The primary reason a loop is not vectorizable is due to *flow-dependencies*. A flow-dependency is when a variable uses information from the pervious iteration of a loop, such as when a value is loaded from an array position that was written to in the previous iteration of the loop. To detect this, we use the fact that indexing into an array is typically done by adding an offset to the base pointer of an array. We are able to find the variables representing array pointers by examing load and store instructions' arguments, and then we use information from our reaching definitions analysis to check whether those variables are computed each iteration as an addition of a constant and another variable. 
+
+Since our vector load/store instructions accesses four consecutive values, we also need to enforce that the non-constant variable is the induction variable, and also that the induction variable must increment or decrement by exactly 1 every iteration. This ensures that arrays are always accessed sequentially.
```
sampsyo:
```suggestion
Since our vector load/store instructions access four consecutive values, we also need to enforce that the non-constant variable is the induction variable, and also that the induction variable must increment or decrement by exactly 1 every iteration. This ensures that arrays are always accessed sequentially. 
```

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+To be able to convert singular addition operations to vector addition, we also check that operations on loaded values must only involve loop-invariant variables (i.e., variables that do not differ per iteration) so that those operations are not flow-dependent. We then check that stores only store variables that are either constant or are results from operations on loaded values (which we previously enforced to be flow-independent).
+
+After we are confident that the array operations in a loop are vectorizable, we now need to convert the loop structure to its equivalent vectorized form.
+
+### Strip Mining
+Since we are working with vector operations on four consecutive array elements, we "chunk" the sequence of loop iterations in blocks of four, which is known as *strip mining*. 
+
+Trivially, strip mining can be done by finding the statement that increments the induction variable and changing it from incrementing/decrementing 1 to incrementing/decrementing 4. All operation on array elements are changed to their vector counterparts, e.g., `add` would become `vadd`. This would work as long as the loop only included operations on arrays, but we found that to often not be the case.
```
sampsyo:
```suggestion
Trivially, strip mining can be done by finding the statement that increments the induction variable and changing it from incrementing/decrementing 1 to incrementing/decrementing 4. All operations on array elements are changed to their vector counterparts, e.g., `add` would become `vadd`. This would work as long as the loop only included operations on arrays, but we found that to often not be the case. 
```

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+  i: int = add i inc;
+  print i;
+  i: int = add i inc;
+  done: bool = eq i size;
+  br done exit loop;
+```
+
+With this method, array operations are allowed to be vectorized while preserving serial instructions because serial instructions are unrolled into four copies, where each copy operates with a different induction variable value. This loop also increments the induction variable by four every iteration which preserves the performance increase from reduced branch overhead. 
+
+In terms of implentaion, we found it much easier to first coalesce the loop into one big block before running the strip mine algorithm because we could treat this sequence of instructions as one array instead of having to worry about jumps and label renaming from inserting new blocks. Aggregating all the blocks of this loop was possible because we previously enforced that there can be exactly one branch instruction located at the end of the loop.
```
sampsyo:
```suggestion
In terms of implementation, we found it much easier to first coalesce the loop into one big block before running the strip mine algorithm because we could treat this sequence of instructions as one array instead of having to worry about jumps and label renaming from inserting new blocks. Aggregating all the blocks of this loop was possible because we previously enforced that there can be exactly one branch instruction located at the end of the loop.
```

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+Up to here, we have been operating on the assumption that array sizes are divisible by the vector size---four. To account for arrays not divisible by four, we append a copy of this loop (without any optimizations) to a block that follows the main optimized loop. This serially executes the remaining loop iteration which allows us to maintain correctness.
```
sampsyo: Just to check, this seems to imply that the `eq i size` check in the above code listing is subtler than it seems, right? Here, `size` must be the "rounded-down" (to the nearest 4) version of the original size to make that work?

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+### Rust Dynamic Library with Foreign Function Interface
+In order to utilize SIMD intrinsics, we use the Rust crate for SIMD that targets the x86 platform. We design a dynamic library with functions for each of the SIMD operations we support, namely vector-add, vector-multiply, and vector-subtract. These functions are called in the dispatch loop of the interpreter for the corresponding vectorized instructions.
```
sampsyo: Wow! That's quite a thing. I think you could have gotten away with "fake" vectorization in the interpreter, but this is even better.

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+case "vadd": {
+    let vecA = getVec(instr, env, 0);
+    let vecB = getVec(instr, env, 1);
+    let vecC = new Int32Array(fixedVecSize);
+    vadd(vecA, vecB, vecC);
+    env.set(instr.dest, vecC);    
+    return NEXT;
+  }
+```
+These external calls add potentially significant overhead to the execution. We quantify this overhead to more accurately evaluate execution time in the interpreter. We run a vvadd benchmark, documented below, varying array sizes of (128, 1024, 2048, 4196, 8192). We run 10 iterations of each configuration and average the execution time between them. We then compare execution time with and without calls to the Rust library. We find there is a 16% overhead for making Rust calls. While we predict this will be offset for especially large arrays, we decided to add rust calls for serialized instructions as well to isolate the vectorized instructions as a variable in our experiment from the Rust call overhead variable.
```
sampsyo:
```suggestion
These external calls add potentially significant overhead to the execution. We quantify this overhead to more accurately evaluate execution time in the interpreter. We run a vvadd benchmark, documented below, varying array sizes of (128, 1024, 2048, 4196, 8192). We run 10 iterations of each configuration and average the execution time between them. We then compare execution time with and without calls to the Rust library. We find there is a 16% overhead for making Rust calls. While we predict this will be offset for especially large arrays, we decided to add Rust calls for serialized instructions as well to isolate the vectorized instructions as a variable in our experiment from the Rust call overhead variable.
```

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+## Evaluation
+### Correctness
+In order to verify correctness, we chose self-verifiying programs such that execution of the program can prove the program computed the same value, and thus is correct  in terms of inputs/outputs. These programs include vector-vector add, vector-vector multiply, and vector-vector subtract.
```
sampsyo: That's fine, I suppose, but it would also have been interesting to try a program or two with a slightly different loop structure than just mapping a single scalar operand over all elements of two arrays.

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Performance
+Determining how to measure the impact of this optimization was challenging as the primary benefit comes from special instructions in the SIMD instruction set that can operate on multiple data points in parallel.
+
+One option is to implement 4 add operations in the interpreter's dispatch loop for vadd, 4 sub operations for vsub, etc. While this would not affect program correctness, it does not attempt to mimic SIMD registers that perform a SIMD operation in the processor. Therefore, it is likely to mispredict realistic speedup by naively ignoring external factors. In addition, the performance benefit seen under this approach would likely only reflect the decrease in number of instructions, rather the specifically the instrinsics of the SIMD instructions.
+
+A second option is to use SIMD instructions in x86 with SIMD intrinsics. This approach involves execution of SIMD instructions that ideally much more closely mirror behvaior of SIMD operations. We modeled these SIMD operations this way for this reason using the aforementioned Rust Dynamic Library.
```
sampsyo:
```suggestion
A second option is to use SIMD instructions in x86 with SIMD intrinsics. This approach involves execution of SIMD instructions that ideally much more closely mirror behvaior of SIMD operations. We modeled these SIMD operations this way for this reason using the aforementioned Rust dynamic library.
```

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+In addition, the calls from TypeScript to the Rust FFI are opaque and it is unclear how exactly they are made. It is also unclear how the arguments are passed in to the function, and what the associated costs are. Therefore, it is possible that some of the variance in the results can be attributed to the variance of these calls.
+
+A next step to better isolate these vectorized instructions frmo the rust call overhead would be to write the entire interpreter in Rust, or some other language, and executing the SIMD instructions inline without function calls.
+
+#### Coupling of Loads and Stores
+As mentioned in the Rust FFI section above, the Rust functions that are invoked for a vector add are coupled with vload and vstore as TypeScript does not have a corresponding type for ```--m128i```. Therefore, it is possible that there are extra vloads and vstores performed for the vectorized programs. This would also dimish the potential performance impact.
```
sampsyo: Are those dashes in the type supposed to be underscores?

sampsyo: Also, "dimish."

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+In addition, the calls from TypeScript to the Rust FFI are opaque and it is unclear how exactly they are made. It is also unclear how the arguments are passed in to the function, and what the associated costs are. Therefore, it is possible that some of the variance in the results can be attributed to the variance of these calls.
+
+A next step to better isolate these vectorized instructions frmo the rust call overhead would be to write the entire interpreter in Rust, or some other language, and executing the SIMD instructions inline without function calls.
+
+#### Coupling of Loads and Stores
+As mentioned in the Rust FFI section above, the Rust functions that are invoked for a vector add are coupled with vload and vstore as TypeScript does not have a corresponding type for ```--m128i```. Therefore, it is possible that there are extra vloads and vstores performed for the vectorized programs. This would also dimish the potential performance impact. 
+
+To isolate each operation in Bril, we would need to write the interpreter in a language that had types compatible with a 128-bit packed inte
```
sampsyo: Incomplete word (and sentence :smiley:).

sampsyo (changes requested):
Looks great! This was a very clear writeup, and it sounds like a challenging but successful implementation. Please see some low-level comments within.

content/blog/2019-10-23-AutoVectoring/index.md
kavoor:
```suggestion
As mentioned in the Rust FFI section above, the Rust functions that are invoked for a vector add are coupled with vload and vstore as TypeScript does not have a corresponding type for ```__m128i```. Therefore, it is possible that there are extra vloads and vstores performed for the vectorized programs. This would also diminish the potential performance impact. 
```

content/blog/2019-10-23-AutoVectoring/index.md
kavoor:
```suggestion
We now want to verify properties to ensure that this loop is indeed vertorizable. For example, we need to know ensure that the bound variable is a constant and not something loaded from memory, or we won't be able to determine how many times the loop is executed. This requires us to know the latest definition of variables, which we get using a reaching definitions analysis. It is possible to find that statement walking backwards through the CFG, but we found using a dataflow analysis to be a much cleaner approach. Furthermore, we'd also like to know values of variables at different program points, such as the base pointers for array loads or stores, which we can get through a copy propagation analysis. 
```

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+In addition, the calls from TypeScript to the Rust FFI are opaque and it is unclear how exactly they are made. It is also unclear how the arguments are passed in to the function, and what the associated costs are. Therefore, it is possible that some of the variance in the results can be attributed to the variance of these calls.
+
+A next step to better isolate these vectorized instructions frmo the rust call overhead would be to write the entire interpreter in Rust, or some other language, and executing the SIMD instructions inline without function calls.
+
+#### Coupling of Loads and Stores
+As mentioned in the Rust FFI section above, the Rust functions that are invoked for a vector add are coupled with vload and vstore as TypeScript does not have a corresponding type for ```__m128i```. Therefore, it is possible that there are extra vloads and vstores performed for the vectorized programs. This would also diminish the potential performance impact. 
+
+To isolate each operation in Bril, we would need to write the interpreter in a language that had types compatible with a 128-bit packed inte
```
kavoor:
```suggestion
To isolate each operation in Bril, we would need to write the interpreter in a language that has types compatible with 128-bit packed integers in order to return the value from a vector load in the Dynamic Library to the interpreter. Similarly, we could then pass these 128-bit packed integers back to the Dynamic Library for the vectorized arithmetic operations.
```

content/blog/2019-10-23-AutoVectoring/index.md
liuhenry4428:
```suggestion
The second loop executes for a fourth the number of iterations of the first loop while behaving identically by using `vload`, `vstore`, and `vadd` that operate on four array elements at a time. This allows for *i* to be incremented by four each iteration instead of by one. 
```

content/blog/2019-10-23-AutoVectoring/index.md
liuhenry4428:
```suggestion
We now want to verify properties to ensure that this loop is indeed vertorizable. For example, we need to know that the bound variable is a constant and not something loaded from memory, or we won't be able to determine how many times the loop is executed. This requires us to know the latest definition of variables, which we get using a reaching definitions analysis. It is possible to find that statement walking backwards through the cfg, but we found using a dataflow analysis to be a much cleaner approach. Furthermore, we'd also like to know values of variables at different program points, such as the base pointers for array loads or stores, which we can get through a copy propagation analysis. 
```

liuhenry4428:
```suggestion
We now want to verify properties to ensure that this loop is indeed vertorizable. For example, we need to know that the bound variable is a constant and not something loaded from memory, or we won't be able to determine how many times the loop is executed. This requires us to know the latest definition of variables, which we get using a reaching definitions analysis. It is possible to find that statement walking backwards through the cfg, but we found using a dataflow analysis to be a much cleaner approach. Furthermore, we'd also like to know values of variables at different program points, such as the base pointers for array loads or stores, which we can get through a copy propagation analysis. 
```

content/blog/2019-10-23-AutoVectoring/index.md
liuhenry4428:
```suggestion
Up to here, we have been operating on the assumption that array sizes are divisible by the vector size---four. To account for arrays not divisible by four, we append a copy of this loop (without any optimizations) to a block that follows the main optimized loop. This serially executes the remaining loop iteration which allows us to maintain correctness. This also means that we need to floor the loop bound of the vectorized loop to the nearest multiple of four but keep the original bound for the serial loop that follows.
```

content/blog/2019-10-23-AutoVectoring/index.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+## Evaluation
+### Correctness
+In order to verify correctness, we chose self-verifiying programs such that execution of the program can prove the program computed the same value, and thus is correct  in terms of inputs/outputs. These programs include vector-vector add, vector-vector multiply, and vector-vector subtract.
```
kavoor:
```suggestion
In order to verify correctness, we chose self-verifying programs such that execution of the program proves the program computed the same value, and thus is correct  in terms of inputs/outputs. These programs include vector-vector add, vector-vector multiply, and vector-vector subtract.
In addition, we tested programs with combinations of these operations, as well as programs with dependencies to ensure we only conservatively unroll. 
For example, consider the following loop example.
```
 a[i] = mul sa[i] a[i]
b[i] = mul sb[i] b[i]
c[i] = add a[i] b[i]
```
In this case, while there are dependencies within the loop body, there are no dependencies between iterations of the loop. Therefore, we expect this operation will be vectorized.
```

sampsyo (approved):
Looks great; thanks!!

Issue #50, opened 2019-09-29
Project 2 Proposal: Induction Variable Elimination


dz333:
**What will you do?**
I will implement an induction variable elimination pass. The goal is to eliminate unnecessary loop variables by replacing references to those variables with appropriate computation.
I will also add the strength reduction of induction variable multiplications into additions.

**How will you do it?**
This involves:
 1) Detecting natural loops
 2) For each loop:
     - Identifying basic induction variables
     - Identifying induction variable families
     - Removing all basic induction variables only used for loop tests & induction variable calculation

This process requires a number of other steps to actually calculate, such as finding dominators, back-edges, calculating number of definitions and liveness of variable definitions within the loop and inserting new pre-header and post-loop basic blocks.

**How will you empirically measure success?**

First, I will sanity check that:
 - unnecessary loop variables are being eliminated
 - code correctness is preserved

For this I will use the existing test suite and benchmarks from here: https://github.com/xu3kev/bril-benchmark/tree/master *and* I will write some pointed tests to stress loop-based optimization correctness (e.g. nested loops).

To measure the effectiveness of the optimization I will measure the following statistics:
 - static & dynamic instruction count
 - total wall-clock runtime

I expect that I will need programs with a large number of loop iterations to have a noticeable effect on dynamic instruction count and it may not be possible to find a difference in wall-clock due to the general interpreter overhead.

For different programs, I expect that there will be a tradeoff between static instruction count and runtime; in particular, when considering the strength reduction optimization this could introduce new variables but replace "expensive" multiplications with "cheap" additions.

**Team members:**
Edit: @Dan12

sampsyo:
Sounds great!! It perhaps goes without saying, but your optimized loops will probably want to manipulate arrays. That is you'll want to have programs that look like this:

```js
for (let i = ...) {
  a[i] = b[i] + c[i];
}
```

…which means you'll probably want to use your extension from Project 1. Your optimization could turn more complex `ptradd` that this will naively use into a simpler version that just bumps the pointers by 1 each time. This kind of thing is the classic "big win" from induction variable elimination.

dz333:
Yup! I definitely plan on using this with the `ptradd` instruction from our extension; I should have mentioned that!

Dan12:
Adding myself as team member

sampsyo:
Closed by #90.

PR #90, opened 2019-10-24
Project 2: Induction Variable Optimizations

dzagieboylo:
This is the blog post for #50.


sampsyo (changes requested):
Really excellent work! This is a solid implementation effort with a good design that's clearly explained. Kudos.

Here are a few minor writing suggestions. Also, you may want to stay consistent about whether the period goes inside the quotes ("this.") or outside ("this".); you've used both in your post.

sampsyo (approved):

PR #92, opened 2019-10-24
Project 2: Tail Call Elimination

### LINKED ISSUE #69: Project 2 Proposal: Tail Call Elimination
**What will you do?**
I will implement [Tail Call](https://en.wikipedia.org/wiki/Tail_call) elimination in the interpreter. This way we can save on stack space and potentially prevent stack overflows. For example, this program should run without any stack overflows:
```
int loop(n: int) {
  zero: int = const 0;
  is_zero: bool = eq n zero;
  br is_zero if_true if_false;
  if_true:
    ret zero;
  if_false:
    one: int = const 1;
    n': int = sub n one;
    new_res: int = call loop(n');
    ret new_res;
}
main: {
  max_int: int = 2147483647;
  result: int = call loop(max_int);
  print result;
}
```

**How will you do it?**
I will extend one of the extensions to Bril that supports function calls (#8 or #14). Because of the "regularity" of Bril operations, detecting tail calls are slightly more tedious. Tail calls usually look like `return f()`, since the call to `f` is the last thing the function does and returns the result of `f`. However, in Bril, tail calls look like `res: _ = foo(); ret res`. I will detect and mark all tail calls in the program, then have the interpreter replace the current stack frame with the next stack frame.

First I will write multiple Bril programs, likely using the TypeScript frontend, that have a mix of tail calls and non-tail calls to make sure the result of the interpreter is what we expect. Then I will write a few tail recursive functions and test that they don't stack overflow. Finally, I can evaluate the performance of the tail recursive functions against the naive interpreter, making sure to pass inputs that won't cause a stack overflow. Additionally, I can compare the performance against imperative loop versions of the functions (e.g., a functional factorial vs. an imperative factorial) to determine if there is any significant overhead in this implementation.

**Team members:**
Just me.

chrisroman:
Closes #69

### DIFF
--- a/content/blog/2019-10-23-tail-call-elimination.md
+++ b/content/blog/2019-10-23-tail-call-elimination.md
@@ -119,7 +119,7 @@ v: int = retval;
 ret v
 ```
 Thus we just need to look for `call`s that are immediately and optionally
-followed by `reval`, and immediately followed by a `ret`.
+followed by `retval`, and immediately followed by a `ret`.
 
 This doesn't take into account more complex cases where there isn't an explicit
 return of a function call, but the value returned only comes from the same
--- a/content/blog/2019-10-23-tail-call-elimination.md
+++ b/content/blog/2019-10-23-tail-call-elimination.md
@@ -205,18 +205,18 @@ big to handle. `n` is the argument passed to the recursive function.
 `loop` is a Bril program that simply loops `n` times using recursion. `factorial`
 is a tail recursive implementation of factorials. `mutual_rec` is a program that
 checks whether a program is even or odd in a mutually recursive way. The code for
-these can be found at (TODO: INSERT LINK HERE).
+these can be found [here](https://github.com/sampsyo/bril/pull/37).
 
 **Percentage Change in Memory Usage Using TCE**
 |            |   n = 1  |  n = 100 | n = 10000 | n = 100000 |
-|:----------:|:--------:|:--------:|:---------:|:----------:|
+|------------|----------|----------|-----------|------------|
 |    loop    |   +0.1%  |   -2.7%  |   -31.1%  |   -79.5%   |
 |  factorial |   +0.2%  |   -2.5%  |   -79.1%  |     X      |
 | mutual_rec |   +0.2%  |   +13.8% |   -31.2%  |   -78.4%   |
 
 **Percentage Change in Execution Time Using TCE**
 |            |   n = 1  |  n = 100 | n = 10000 | n = 100000 |
-|:----------:|:--------:|:--------:|:---------:|:----------:|
+|------------|----------|----------|-----------|------------|
 |    loop    |  +2.1%   |   +2.2%  |   -10%    |   -29.8%   |
 |  factorial |  +1.1%   |   +5.5%  |   -1.6%   |      X     |
 | mutual_rec |  +1.1%   |   -1%    |   +5.7%   |   +5.07%   |
--- a/content/blog/2019-10-23-tail-call-elimination.md
+++ b/content/blog/2019-10-23-tail-call-elimination.md
@@ -174,7 +174,7 @@ endif.6:
 ```
 
 Then we can analyze the CFG backwards to see that indeed we can replace
-the calls to a `jmp`.
+the calls with `jmp` instructions.
 ```
 then.6:
   ...
--- a/content/blog/2019-10-23-tail-call-elimination.md
+++ b/content/blog/2019-10-23-tail-call-elimination.md
@@ -14,7 +14,7 @@ call is a call to a function whose value is immediately returned. For example,
 we do not need to create a new stack frame for it. We can simply "overwrite"
 the current stack frame since won't need any of the values in the frame anymore.
 This is crucial for programming languages that use *tail recursion* as a
-programming idiom (e.g. OCaml, Haskell, etc.). Consider the following (somewhat
+programming idiom (e.g., OCaml, Haskell, etc.). Consider the following (somewhat
 contrived) TypeScript program:
 ```
 function loop(n: number) {
--- a/content/blog/2019-10-23-tail-call-elimination.md
+++ b/content/blog/2019-10-23-tail-call-elimination.md
@@ -83,7 +83,7 @@ Here, if `foo` returned 0, then `r` would have the value 0.
 
 ### Extending the Interpreter
 The interpreter needed to be extended to implement the above semantics. To model
-a stack frame, I explicitly keep track of the program counter (i.e. which
+a stack frame, I explicitly keep track of the program counter (i.e., which
 instruction is being interpreted), the name of the current function, and an
 environment. I made this explicit because it made jumping to other functions
 easier to implement. The interpreter simply tries to evaluate the frame at the
--- a/content/blog/2019-10-23-tail-call-elimination.md
+++ b/content/blog/2019-10-23-tail-call-elimination.md
@@ -208,15 +208,17 @@ checks whether a program is even or odd in a mutually recursive way. The code fo
 these can be found [here](https://github.com/sampsyo/bril/pull/37).
 
 **Percentage Change in Memory Usage Using TCE**
+
 |            |   n = 1  |  n = 100 | n = 10000 | n = 100000 |
-|------------|----------|----------|-----------|------------|
+|:----------:|:--------:|:--------:|:---------:|:----------:|
 |    loop    |   +0.1%  |   -2.7%  |   -31.1%  |   -79.5%   |
 |  factorial |   +0.2%  |   -2.5%  |   -79.1%  |     X      |
 | mutual_rec |   +0.2%  |   +13.8% |   -31.2%  |   -78.4%   |
 
 **Percentage Change in Execution Time Using TCE**
+
 |            |   n = 1  |  n = 100 | n = 10000 | n = 100000 |
-|------------|----------|----------|-----------|------------|
+|:----------:|:--------:|:--------:|:---------:|:----------:|
 |    loop    |  +2.1%   |   +2.2%  |   -10%    |   -29.8%   |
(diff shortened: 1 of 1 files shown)


content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+even function calls in general. To enrich the language, we can make Bril more
+closely resemble something like x86 assembly. Intuitively, all Bril variables
+implictly live on the stack. To pass arguments in a function call, we must
+explicitly push them onto the stack. This is akin to pushing values on the stack
+in assembly. When a function returns, those arguments are implictly popped off the stack.
+Additionally, the return value of the function is obtained using a special
+keyword that is akin to getting the return value from `rax` according to the
+[System V Calling Conventions](https://en.wikipedia.org/wiki/X86_calling_conventions#x86-64_calling_conventions).
+
+Importantly, we need the capability to jump program locations other than the
```
sampsyo: "jump program locations": missing "to"?

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+labels in the program. This way, if we have a tail call, then we can jump to
+the beginning of the callee instead of creating a new stack frame. This is [how
+tail call elimination is implemented](https://www.cs.cornell.edu/courses/cs4120/2019sp/lectures/34functional/lec34-sp19.pdf?1556325712)
+in x86 assembly.
```
sampsyo: It seems like the more salient thing about this is the ability to jump to *other functions*, not to jump to *unlabeled* locations as the original setup seems to imply.

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+  ...
+}
+```
+Functions must specify a return type, a name, and a potentially empty list of
+typed arguments.
+
+Here are the new effect operations and their semantics:
+- `push arg1 ... argn`: push arguments to the stack, which can be used by the
+first function to be `call`'d after this instruction.
+- `call foo`: starts executing instructions defined by the function `foo`
```
sampsyo: Missing a period (the other bullet in this list has one).

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+```
+...
+call foo
+r: int = retval;
+```
+Here, if `foo` returned 0, then `r` would have the value 0.
+
+### Extending the Interpreter
+The interpreter needed to be extended to implement the above semantics. To model
+a stack frame, I explicitly keep track of the program counter (i.e. which
```
sampsyo:
```suggestion
a stack frame, I explicitly keep track of the program counter (i.e., which
```
Again: https://capra.cs.cornell.edu/styleguide/#egie

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+To return a value, a special variable name is set in the environment in the
+previous stack frame so it can be retrieved by the caller using `retval`.
+
+### Extending the TypeScript Frontend
+For the majority of this, I referred to [Alexa and Greg's implementation](https://github.com/sampsyo/bril/pull/16).
+There were some differences however because I have separate instructions for
+passing arguments to a function. So, whenever a function call was found in the
+AST, the arguments needed to be converted to Bril instructions first, then those
+would be used in a `push`. Additionally, a `retval` would need to be created
+afterwards if in the AST, the function call was assigned to a variable.
```
sampsyo:
```suggestion
afterwards if, in the AST, the function call was assigned to a variable.
```
The *result* of the function call, for example? And it doesn't seem to matter whether or not this is assigned to a variable… for example, you'd want to get the return value of `f` if you did `g(f() + x)` or something, i.e., using the result for something other than an assignment.

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+```
+to Bril would be
+```
+push n
+call foo
+v: int = retval;
+ret v
+```
+Thus we just need to look for `call`s that are immediately and optionally
+followed by `reval`, and immediately followed by a `ret`.
```
sampsyo:
```suggestion
followed by `retval`, and immediately followed by a `ret`.
```

sampsyo: Is it OK for functions to just ignore the result altogether—that is, just to have a `call` immediately followed by a `ret` with no arguments?

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+
+This doesn't take into account more complex cases where there isn't an explicit
+return of a function call, but the value returned only comes from the same
+created. For example:
```
sampsyo: "the same created": missing a word?

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+else.6:
+  ...
+  call foo;
+  result: int = retval;
+endif.6:
+  ret result;
+```
+
+Then we can analyze the CFG backwards to see that indeed we can replace
+the calls to a `jmp`.
```
sampsyo: "replace the calls with `jmp` instructions" (matching plurality)

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+  ...
+  jmp foo;
+endif.6:
+  ret result;
+```
+Note that the extra instructions can simply be removed by a DCE pass, so we don't
+worry about that.
+
+*Unfortunately, I couldn't get this to work properly because my copy propagation
+pass had bugs.*
```
sampsyo: Maybe mention that a little earlier? Just say that this is something you wanted to, but couldn't quite get to work because of copyprop.

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+## Evaluation
+To evaluate that the tail call elimination is working and actually gives us an
+improvement, we benchmark some recursive functions that use tail recursion, and
+show the difference in execution time and memory usage between an optimized and
+unoptimized Bril program.
+
+The table entries show how much change was observed, as a percentage, by doing
+tail call elimination (TCE). For example, an entry of -10% means the optimized program
+used 10% less memory/time than the unoptimized program. An `X` means that the output of the program was too
+big to handle. `n` is the argument passed to the recursive function.
```
sampsyo: Where does this limit originate from? (Did the interpreter crash?)

content/blog/2019-10-23-tail-call-elimination.md
```diff
@@ -0,0 +1,239 @@
...
+|    loop    |  +2.1%   |   +2.2%  |   -10%    |   -29.8%   |
+|  factorial |  +1.1%   |   +5.5%  |   -1.6%   |      X     |
+| mutual_rec |  +1.1%   |   -1%    |   +5.7%   |   +5.07%   |
+
+To get the execution time and peak memory usage, I use `/usr/bin/time -l` (which prints the contents of rusage).
+To make sure the measurements are meaningful, I chose a maximum `n` value so
+that the tests took a few seconds. Here we can clearly see that with large values
+of `n`, the programs with TCE use considerably less memory. However, it is unclear
+whether there is a benefit to the execution time of the program since the values
+vary quite a bit.
```
sampsyo: If numbers vary a lot, it might be a good idea to report the variance or standard deviation or whatever to give a more complete picture of the distribution. (No need to go back and do this for this post, of course.)

sampsyo (changes requested):
Good work! It is interesting that much of the work here went into making the language more assembly-like, with a more explicit calling convention. I just have a few writing suggestions inline.

content/blog/2019-10-23-tail-call-elimination.md
chrisroman:
```suggestion
Importantly, we need the capability to jump to program locations other than the
```

content/blog/2019-10-23-tail-call-elimination.md
chrisroman:
```suggestion
- `call foo`: starts executing instructions defined by the function `foo`.
```

content/blog/2019-10-23-tail-call-elimination.md
chrisroman: Agh, I'll internalize this one day.

content/blog/2019-10-23-tail-call-elimination.md
chrisroman:
```suggestion
the calls with `jmp` instructions.
```

content/blog/2019-10-23-tail-call-elimination.md
chrisroman: If the function isn't supposed to return anything, then this sounds fine to me, since we're immediately returning, so we don't need the current stack frame anymore, meaning we can jump to the function instead of doing a call.

content/blog/2019-10-23-tail-call-elimination.md
chrisroman: Good point. It definitely would have been good to include those things.

content/blog/2019-10-23-tail-call-elimination.md
chrisroman: For the loop version of the program, it runs out of memory:
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

sampsyo (approved):
Looks great!

PR #93, opened 2019-10-25
Efficient Implementation of Self

rachitnigam:


content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+  [subtweets](https://twitter.com/notypes/status/1170037148290080771) his advisor and [annoys tenured professors](https://twitter.com/natefoster/status/1074401015565291520).
+"""
+[[extra.authors]]
+name = "Rachit Nigam"
+link = "https://rachitnigam.com"
++++
+
+[An Efficient Implementation of Self, a Dynamically-Typed Object-Oriented
+Language Based on Prototypes][paper] presents techniques for runtime
+compilation, now more commonly referred to as Just In Time (JIT) compilation,
```
sampsyo:
```suggestion
compilation, now more commonly referred to as just-in-time (JIT) compilation,
```

It's usually not capitalized, but it usually does have hyphens.

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+> We have developed and implemented techniques that double the performance of
+> the dynamically-typed object oriented languages.
+
+
+## The Challenge
+
+The term "dynamic language" is mostly commonly associated with modern scripting
+languages like Python and JavaScript. Self, a much older language developed
+at the famed Xerox PARC labs takes the philosophy of dynamism to a logical
```
sampsyo:
```suggestion
at the famed Xerox PARC labs, takes the philosophy of dynamism to a logical
```

Missing a closer for the parenthetical comma.

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+extreme -- *everything* in self is a message to an object. This includes
+Java-like method calls on object *as well as* control structures like loops
+and conditionals.
```
sampsyo: It might be worth mentioning that JavaScript is directly inspired by Self (and Scheme), which is why there will be some suspicious similarities here.

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+For example, the Self condition `IfTrue:IfFalse:` is a method invocation on
+the boolean object `true` and `false`. This means that a simple conditional:
+
+```
+if (x) 1 else 2
+```
+
+in Self parlance (written with JavaScript-esque syntax) looks like this:
```
sampsyo: The above is a little hard to understand for two reasons:

- It's not clear which of these is supposed to be JavaScript-like. I think it's the second one. But then it's not 100% sure what "language" the first snippet is written in—I think it's just a notional, abstract language, but maybe it's worth mentioning that it's not supposed to be Self or Self-esque JS.
- In JS, at least `if (x) 1 else 2` isn't terribly meaningful because `if` isn't an expression, unlike `x ? 1 : 2`. Maybe that's fine, though—this does make sense in, like, Scala or Rust.

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+
+```
+x.If({ true: () => 1, false: () => 2 })
+```
+
+The runtime behavior of this program is invoking the method `IfTrue` on the
```
sampsyo:
```suggestion
The run-time behavior of this program is invoking the method `IfTrue` on the
```
https://capra.cs.cornell.edu/styleguide/#runtime

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+object `x`, which can be a `true` or `false` but is not required to be those,
+and execute the "thunk" (a function with no argument) corresponding to the true
+or the false branch.  Note that `x` is not required to be `true` or `false`.
+Any Self object can define the `If` method and specify its conditional
+behavior.
```
sampsyo: This paragraph switches a few times between several names for the same method: `If`, `IfTrue`, and `IfTrue:IfFalse:`. Maybe pick one?

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+While the paper goes into the nitty-gritty of object layouts and method
+invocations, the essence of the paper can be summarized as:
+
+>  When type information for an object is available, generate specialized code
+>  and let inlining and compiler optimizations work their magic.
+
+The ideas described in the paper have influenced the design of modern JITs for
+JavaScript which use a lot of [similar features][jit] to improve JavaScript execution
+times.
```
sampsyo: Not just JavaScript compilers—basically every JIT in use today descends from this set of ideas. (Even ones for more static-seeming languages like Java and C#.)

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+from a prototype is as simple as _cloning_ it and setting its parent pointer to
+the prototype. Changing the common behavior of all clones is as simple as
+changing the behavior of a prototype. Since method and field lookups traverse
+the parent hierarchy, clones can also override methods and fields of their
+prototypes.
+
+A naive layout scheme for objects would copy all fields from a prototype and
+end up wasting a lot of space describing potentially shared behaviors. The Self
+compiler minimizes space usage of _clones_ derived from the same prototype by
+using _clone families_. A cloned object only stores its modifiable in its
```
sampsyo: Missing noun after "modifiable."

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+object and points to the clone family for its prototype. If the instance ever
```
sampsyo: I'm not 100% sure what "in its object" means—maybe introduce something like an "allocation" to mean the actual memory for laying out fields, as opposed to "object" which is more abstract?

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+overrides one of its methods, the creates a new clone family to preserve
+semantics and propagate behavior changes to all clones of a prototype.
+
+<img src="without-maps.png" alt="drawing" width="300"/>
+<img src="with-maps.png" alt="drawing" width="300"/>
```
sampsyo: These figures are a little context free—maybe include a sentence or two describing what we're looking for in these pictures?

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+Inlining remains one of the most crucial optimization in modern compilers,
+enabling other optimizations to become more effective. However, inlining
+becomes impossible in presence of dynamic method calls if the type of the
+receiver object is unknown. This is an even bigger problem for Self since
+most of program is a sequence of dynamic calls.
+
+The Self compiler performs two optimizations, message _inlining_ and
+_splitting_ to enable efficient execution. Message _inlining_ works in a similar
+fashion to customized compilation -- If the type of a receiver is know, either
```
sampsyo:
```suggestion
fashion to customized compilation -- if the type of a receiver is known, either
```

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+at initial compilation through a dataflow analysis or at runtime, inline the
+method body at the call location. When type information for an object is lost
+due to control flow splits, _message splitting_ generates specialized code for
+possible receiver types and guards them using type tests. If the type of
+receiver matches, fast code can be executed.
```
sampsyo: To drive this point home, maybe it would be useful to include some example code to illustrate what the program looks like after these optimizations are applied. For example, you could show several imaginary assembly- or C-like listings depicting these versions of a program `x < y`:

- The obvious version, where the code looks up the string `lessThan` in the method dictionary for the object `x` and jumps to the corresponding code with argument `y`.
- The "customized compilation" version, with a type guard and a direct `call` to the `integer.lessThan` function with `y` as the argument.
- The "inlined" version, which still has the type guard but just uses a `lt` machine instruction or whatever to do the comparison—no call necessary.

For bonus points, you could even show imaginary code for the "split" version.

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Programming Environment Support
+
+The Self compiler support _incremental recompilation_ and _source-level
```
sampsyo:
```suggestion
The Self compiler supports _incremental recompilation_ and _source-level
```

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+specializations in a map.
+
+Incremental Recompilation occurs when the compiler observes a change in the
+programming environment and invalidates compiled methods associated with the
+affected data. At compile time, the compiler creates a dependency list encoding
+the information required for specialization. For example, if a method was
+inlined from a clone's parent, updating the parent pointer or the method body
+in the parent should invalidate the specialized code. Since the compiler
+selectively invalidates code, methods that weren't modified can still execute
+with specialized code.
```
sampsyo: This is the weirdest part of Self, and the one that has not stuck with us today: the interactive, live-editable code environment. It's maybe worth pointing out that this is the most anachronistic part of the paper as a result. V8 doesn't have an interactive, persistent, first-class IDE, for example, so it doesn't have to worry about this as much.

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+Benchmarks. The authors both did a straightforward transliteration of Self programs
+(marked `Self`) and also rewrote the benchmarks in idiomatic Self (marked `Self'`).
+
+<img src="eval.png" alt="drawing" width="400"/>
+
+On average, Self was 2-3x faster than the Smalltalk implementation and 4x slower
+than the original C programs.
+
+The authors also describes a new metric to measure the performance of an object
+oriented programming language - _Millions of Messages per Second (MiMS)_ which
```
sampsyo: Elsewhere, you have used ` -- ` to separate thoughts, rather than just a single hyphen. Consider sticking to that or just using commas here.

content/blog/2019-10-28-self/index.md
```diff
@@ -0,0 +1,176 @@
...
+corresponds to MIPS. To compute the MiMS of a specific virtual machine, divide
```
sampsyo: How does it "correspond" to MIPS? The paper says it's "analogous" to MIPS, which I think is more accurate.

sampsyo (changes requested):
Looks pretty good; thanks for putting this together! This PR does contain your Project 2 blog post, in case that's easy to resolve.

Although you don't need to hit every point, I notice that there are a few things from the paper that you didn't cover, some of which are historically pretty important:

- Heap organization to facilitate GC ("segregation").
- Inline word tagging for integer & FP values.
- Bytecode (preceded the JVM).

For at least the second two, which are bog-standard JIT implementation techniques today, it might be worth at least briefly mentioning that this paper covers them—and attempting to put them into a modern context.

rachitnigam:
Thanks for the comments! I've updated the inline comments and added sections on segregation, tagging, and the bytecode. I initially chose to forgo them because I thought they were not contributions of the paper. It is remarkable how many ideas this paper contains!

I'm going to move this to a new PR to remove the commits from my project.

rachitnigam:
Moved to #95.

PR #94, opened 2019-10-27
Project 2: Sparse Conditional Constant Propagation

### LINKED ISSUE #54: Project 2 Proposal: Sparse Conditional Constant Propagation
**What will you do?**
Implement an optimizing pass on Bril programs for sparse conditional constant propagation, as described in [Constant propagation with conditional branches](https://dl.acm.org/citation.cfm?id=103136). This pass will evaluate expressions at compile time when the inputs are determined to be constant, and find unreachable code from known constant branch conditions. These two effects of the optimization work well together as from knowledge of an untraversed control-flow path, you can discover more constant variables, and vice versa. 

**How will you do it?**
I will follow the algorithm to compute constants described in the paper, which involves converting the program to SSA form and then invoking a work-list-based data-flow analysis. The next step will be to modify the program to replace variables with their known constant values, and to remove code determined to be unreachable. Then the program needs to be converted back from SSA form. I may then need to run a pass of dead-code elimination in order to show the full effect of constant propagation.

I will evaluate the effect on program performance by measuring the dynamic instruction counts of programs with and without the optimization applied. If I include DCE in the optimization, I will also include a version optimized by DCE alone in the comparisons in order to properly attribute where the instruction count changes are coming from.
I think that dynamic instruction count is the most fruitful statistic for evaluating Bril program performance in this case. Execution time in the interpreter would most likely see negligible differences for almost any optimization. The static instruction count would see a difference, but it does not represent program speed, and code size reduction is not the goal of this optimization. However, the effect on dynamic instruction count might not be that great either, because Bril has no support for inlined immediate values in operations, and so a `const` instruction will still be required for each constant.

**Team members:**
Just me.

anastos:
closes #54

### DIFF
--- a/content/blog/2019-10-23-sccp.md
+++ b/content/blog/2019-10-23-sccp.md
@@ -54,10 +54,10 @@ definition could be removed as well.
 The example above is clearly contrived to show the capabilities of conditional
 constant propagation. And so, it may not be obvious if and when this
 optimization would truly be beneficial for real programs. Why would a program
-contain code that will never be run? A common answer to this question is that
+contain code that will never run? A common answer to this question is that
 people often put code that is meant for debugging under conditionals. As such,
-when the code is compiled with debugging disabled, all of that code is unneeded,
-and can be removed through this optimization.
+when the code is compiled for production, with debugging disabled, all of that
+code is unneeded and can be removed through this optimization.
 
 However, I think that a more important motivation for this optimization revolves
 around the fact that it is operating on an intermediate language. People would
@@ -73,7 +73,7 @@ benefits even for programs without any obviously constant expressions.
 ## Implementation
 
 Sparse conditional constant propagation was introduced by Wegman and Zadeck in
-[Constant Propagation with Conditional Branches][SCCP] (1991). My
+“[Constant Propagation with Conditional Branches][SCCP]” (1991). My
 [implementation][] of the data-flow analysis is based on the description given
 in that paper. The analysis works on programs in SSA form, and as such I also
 needed to implement transformations on Bril programs to and from SSA form. After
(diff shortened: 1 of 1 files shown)


content/blog/2019-10-23-sccp.md
```diff
@@ -0,0 +1,263 @@
...
+general, less efficient then the original.[^coalesce] However, during
+implementation I realized that, specifically for this optimization, the fully
+general conversion back from SSA is not necessary. This is because constant
+propagation only ever decreases the live ranges of SSA variables, by replacing
+their uses with constants or removing dead code. As such, the live ranges of SSA
+variables that come from the same original variable will never interfere with
+each other and can simply be renamed back to the original variable name. If this
+SSA conversion were to be used for other optimizations that do not share this
+property with constant propagation, the conversion back would need to be
+modified.
```
sampsyo: Cool shortcut!

content/blog/2019-10-23-sccp.md
```diff
@@ -0,0 +1,263 @@
...
+The primary component of the optimization is the constant propagation analysis
+itself. I implemented it according to the worklist algorithm described in the
+paper. Doing so was mostly straightforward. However, the paper seems to not
+mention one point in the algorithm where it is necessary to add to the worklist,
+which took a while for me to figure out.[^visit-phi] The output of the analysis
+is a mapping from variables to elements of a lattice, which can be ⊤ (the
+variable is undefined), ⊥ (the value is unknown), or a value (the variable is a
+constant). Because the program is in SSA form, the analysis only needs one
+lattice element per variable, instead of a lattice element for each variable at
+each program point. The paper aludes to the fact that a constant propagation
```
sampsyo: Indeed; much better.

sampsyo:
```suggestion
each program point. The paper alludes to the fact that a constant propagation
```

content/blog/2019-10-23-sccp.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Evaluation
+
+It has proven difficult to evaluate the effectiveness of this optimization in a
+manner that would accurately reflect its utility. Upon running the optimization
+on several test cases, I have observed that, for programs that are written
+directly by humans in Bril, the optimization tends to exhibit one of two
+behaviors: either the program is effectively unchanged, or it is completely
+evaluated leaving only assignments of constants to variables, and print
+statements with those variables. By running these test cases, however, I have
```
sampsyo: Hmm; maybe it would be better to use a benchmark suite with actual, real inputs instead of just baked-in constant inputs… maybe this "collapse" to constants could be avoided by moving the computation into a function takes the inputs as arguments?

content/blog/2019-10-23-sccp.md
```diff
@@ -0,0 +1,263 @@
...
+
+</div>
+
+For this program, the optimization removed only three instructions, one of which
+was in the loop body (an extraneous assignment to a variable that was never
+used). In general, it is difficult to judge the effectiveness of this
+optimization without the infrastructure of other optimizations and compilers to
+pair it with. In the future, if someone were to implement, say, a function
+inlining optimization, and were to extend this optimization to support function
+calls, I would be interested to see the effect it would have.
```
sampsyo: Indeed!

content/blog/2019-10-23-sccp.md
```diff
@@ -0,0 +1,263 @@
...
+
+--------------------------------------------------------------------------------
```
sampsyo: FWIW, we added some style to footnotes in the CSS for the blog posts such that a visual separator may not be necessary.

sampsyo (changes requested):
Very nice work! This is an extremely clear writeup about a solid optimization. :+1: I have a few low-level writing suggestions within.

sampsyo (approved):
Looks great; thanks!!

PR #103, opened 2019-10-30
Project 2: Global Value Numbering

### LINKED ISSUE #59: Project 2 Proposal: Global Value Numbering
**What will you do?**
We will implement global value numbering (per-function, across basic blocks) for Bril programs.

**How will you do it?**
We first need to implement dominator tree construction and conversion to SSA. We will then implement the value partitioning algorithm from [Briggs, Cooper, & Simpson 1997][]. Finally, we’ll add an extension for constant propagation. 

Our evaluation will consider two metrics (in order of increasing ambition). In particular, we will compare these metrics on the same program before and after our optimization. Will hope to use the TypeScript front end to create and share a large benchmark suite, based in part on porting benchmarks from SPEC and other existing suites.

- Code size (the fewer Bril instructions, the better).
- Runtime performance of programs (either via measuring the interpreted Bril execution time, or by measuring the machine code generated from the existing LLVM backend).

In addition, we hope to extend our property-based testing from Project 1 to generate programs that have redundant subexpressions, in order to fuzz-test our implementation. 

**Team members:**
@gyauney 

[Briggs, Cooper, & Simpson 1997]:http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.36.8877&rep=rep1&type=pdf

avanhatt:
Closes #59


content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+  print x3;
+```
+
+Having unique variable assignments is enormously helpful in implementing optimization---most industrial-strength compilers, including [LLVM][] and [GCC][], rely on it internally. We followed a standard algorithm to convert Bril programs to SSA form. 
+
+The first pass inserts `phi` nodes wherever control flow merges multiple variable definitions. 
+We determine this condition by computing the [_dominance frontier_][domf] (essentially, boundaries in the control flow graph where control flow can recombine) for every basic block. 
+For every block in the dominance frontier of every variable, we insert a `phi` node with as many arguments as there are predecessors to the block. 
+One bug we hit with this part of the implementation was realizing that we only should insert `phi` nodes at the dominance frontier for variables that we assigned to multiple times. 
+For convenience of later analysis, we also augmented the base algorithm to track which basic blocks (or source) each `phi` argument originated from (for the above example, our generated `phi` would be `x3 : int = phi x1 x2 left right;`)
```
sampsyo: Huh, that's a pretty good idea… maybe I should standardize an "SSA Bril" extension that encodes phi-nodes this way.

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+For convenience of later analysis, we also augmented the base algorithm to track which basic blocks (or source) each `phi` argument originated from (for the above example, our generated `phi` would be `x3 : int = phi x1 x2 left right;`)  
+
+The second pass is a recursive renaming algorithm that uses a stack to determine the correct version of a variable name for every point in the program.
+The trickiest part of implementing this algorithm was traversing the basic blocks in the correct order.
+The high level algorithm specifies that you should recursively traverse the children of each block in the dominator tree. 
+However, we found that not only did we ned to traverse the children, but we needed to traverse them _in order_ relative to the control flow graph. 
+In particular, Bril’s ecosystem already had a utility for getting the dominators of every block. To get the dominance tree, we calculated the transitive reduction of this relation. 
+Our implementation additionally extending the dominance tree such that every list of children was ordered relative to a reverse post-order traversal of the control flow graph.
```
sampsyo:
```suggestion
Our implementation additionally extends the dominance tree such that every list of children was ordered relative to a reverse post-order traversal of the control flow graph.
```

…or "extended."

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+
+In later testing (using an example, `test/gvn/constant-folding-tau-example.bril`, ported from the [class notes of Prof. Mooly Sagiv of Tel Aviv University)][tau], we found one minor bug with our SSA implementation. 
+Bril programs do not have a type checker and are aggressively dynamic: it’s perfectly valid to assign to a variable in one branch of a conditional but not the other. 
+The base SSA algorithm does not account for this, so we fail to properly rename in the second step of our algorithm. 
+For now, we work around this case by assigning a default value to to the variable before we branch.
```
sampsyo: Yeah, that's totally the right thing to do—to imagine that every variable is assigned to with a special "uninitialized" value at the entry block. I should probably mention this in future SSA lectures.

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+
+[ssa]: https://en.wikipedia.org/wiki/Static_single_assignment_form
+[llvm]: https://llvm.org/docs/tutorial/OCamlLangImpl7.html
+[gcc]: https://gcc.gnu.org/onlinedocs/gccint/SSA.html
+[domf]: https://en.wikipedia.org/wiki/Dominator_(graph_theory)
+[tau]: http://www.cs.tau.ac.il/~msagiv/courses/pa07/lecture2-notes-update.pdf
+
+## Global value numbering
+
+There are two main approaches to global value numbering, as described in [Value Numbering][], Briggs, Cooper, and Simpson, 1997, both of which ultimately require input code in SSA form.
```
sampsyo: Use quotation marks around titles to make it clear that it's the title of something.

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+Hash-based techniques, like those used in local value numbering, hash operations and find redundant values by comparing to previously hashed operations.
+Partitioning algorithms, on the other hand, divide all operations into equivalence classes.
+
+We chose to implement hash-based value numbering because it allows for copy propagation extensions and operates on the dominator tree computed during conversion to SSA.
+While converting to SSA and performing global value numbering can be performed together in one pass over the original code due to their similarity, we chose to implement them separately for modularity.
+A distinct SSA conversion allows more downstream optimizations to operate on the program's SSA form in addition to allowing testing of both passes independently.
+
+We implemented the following pseudocode, adapted from Figure 4 of the aforementioned paper, so called because it is a 'dominator-based value numbering technique'.
```
sampsyo:
```suggestion
We implemented the following pseudocode, adapted from Figure 4 of the aforementioned paper, so called because it is a "dominator-based value numbering technique."
```

In most :us: style guides, double quotes are the top-level thing, and periods go inside the punctuation.

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+```
+
+We extend our global value numbering to perform constant folding by keeping a mapping from variable names defined as constants to their integer or boolean literal values.
+When we encounter and arithmetic operation, we check whether all operands are constant, and fold (that is, perform the operation and drop in the result) if so.
+The argument constants can then often be removed entirely with a later dead code elimination pass.
+
+When implementing constant folding, we [found a bug][bug] in Bril's existing local value numbering implementation.
+When Bril programs have division by zero, the local constant folding failed with an exception.
+We made the design judgment that compiler passes should not fail on misbehaving code, but instead generate code with the same behavior (for example, the implementation's dynamic error could exist on a branch that never actually executes).
+We modifying this behavior in both the central Bril local value and our new global value numbering to instead bail out on constant folding when encountering a potential exception in the folding step.
```
sampsyo:
```suggestion
We modified this behavior in both the central Bril local value and our new global value numbering to instead bail out on constant folding when encountering a potential exception in the folding step.
```

sampsyo: "Bril local value" -> +numbering?

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+
+[bug]: https://github.com/sampsyo/bril/issues/40
+
+<!-- Difficulties:
+- Recursively visiting blocks in the dominator tree in reverse post order is not enough to ensure that a phi node's arguments have been processed in the absence of backedges. 
+We also had to sort the immediately dominated blocks by their order in the CFG.
+- Can't do constant propagation because Bril operations require registers as operands.
+- `examples/gvn/constant-folding-calpoly-example.bril`: There's a division by zero when running constant folding, but the compiler shouldn't crash!
+We filed an issue with Bril's local value numbering and fixed our implementation.
+ -->
```
sampsyo: I think you've covered all of this in un-commented text.

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+B2:
+  jmp B4;
+B3:
+  jmp B4;
+B4:
+  x2: int = phi v0 w0 B2 B3;
+  z0: int = add u0 x2;
+  ret ;
+}
+```
```
sampsyo: Yay! I really like the idea of reproducing a specific example from the paper.

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+We treat the number of instructions as a proxy for code performance, where fewer lines of code is superior.
+For each of a number of Bril test programs, we report in the following graph:
+1. The number of instructions after converting to SSA, converting out of SSA, and running Bril's existing [trivial dead code elimination][] (TDCE), shown in blue.
+2. The number of instructions after converting to SSA, running GVN, converting out of SSA, and running TDCE, shown in red.
+
+<img src="eval-correctness.png" width="500"/>
+
+### Bigger benchmarks with TypeScript frontend and comparison with LVN
+
+For a slightly more realistic analysis, we ran GVN on the more complex programs afforded by Bril's TypeScript frontend. In particular, none of these programs were written with intentional redundancy like some of our correctness test cases (thought the conversion from TypeScript to Bril does cause some), so they portray a more realistic evaluation of the benefits of value numbering.
```
sampsyo:
```suggestion
For a slightly more realistic analysis, we ran GVN on the more complex programs afforded by Bril's TypeScript frontend. In particular, none of these programs were written with intentional redundancy like some of our correctness test cases (though the conversion from TypeScript to Bril does cause some), so they portray a more realistic evaluation of the benefits of value numbering.
```

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+In addition to the existing test programs from project 1, we also implemented a quadratic formula calculation (`test/gvn/quadratic.ts.bril`), [fizz buzz][] (`test/gvn/fizz-buzz.ts.bril`), and a naive [Sieve of Eratosthenes][] without arrays (`test/gvn/check-primes.ts.bril`).
+All TypeScript programs were first processed by the `ts2bril` frontend.
+We encountered an implementation difficulty: the code generated by the TypeScript frontend sometimes inserts `jmp` instructions after returns in the same basic block.
+We had to manually remove these in order for SSA conversion and dominator tree creation to work.
```
sampsyo: Hmm, yeah… I guess maybe the thing to do in the future would be to have the CFG generator transform `ret`s into jumps to a common exit block that just contains `ret` all by itself, in the case where there are multiple return points.

sampsyo (changes requested):
Fantastic work! Thanks for clearly explaining every step along the way with specific examples. This makes your implementation very easy to believe. I only have a few low-level text edits within.

content/blog/2019-10-23-global-value-numbering/index.md
```diff
@@ -0,0 +1,357 @@
...
+
+If you follow [PL Twitter][pltwitter]™, you may have seen this tweet go by from our very own [Adrian][]:
+
+<img src="twitter.jpg" width="500"/>
+
+What, you may ask, is value numbering, and how is it so mind-blowing? 
+
+The basic premise of [value numbering][vn] is that we can make our code more efficient by removing duplicate computation during compilation.
+In particular, if we assign values to every “unique” piece of computation, we can save on work by only actually running that piece once.
+At every subsequent use, we can then refer to the already-computed value.
```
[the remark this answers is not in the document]
avanhatt: Added a bit more about this in, while keeping the computation reuse example, thanks!

sampsyo (approved):
Excellent! Looks great.

Issue #114, opened 2019-10-31
Project 3 Proposal: Dynamic Memory Safety


dz333:
**What will you do?**
Instrument `getelementptr` instructions with guards to ensure that the sized type accesses are safe. If the guard fails, the program will abort with an exception. This is mostly equivalent to array bounds checking for arrays of statically-known size.

**How will you do it?**
I will use the LLVM type information from arrays and structs of statically known size to determine the maximum appropriate offsets for each of the arguments to `getelementptr`
If the offset arguments are constants I can throw a compile time error; otherwise I will insert a guard to check that the offset arguments are of correct size (i.e. between 0 and the maximum index, inclusively). For any pointer whose element length type is unsized (e.g., `i32*` or `[0 x i32]`) I will assume that it points to a single value.

This approach is (mostly) sound but not complete (with respect to `getelementptr`). I say mostly sound because the only sound approach here is to either collect a lot more information about the allocation site for unsized pointers or to disallow all unsized pointers. In general this also requires a pointer liveness analysis for soundness (i.e., even if a pointer is of sized type it may be aliased by another reference which has already been `freed`).  It is also unsound since LLVM allows for bitcasting, which can convert a sized pointer -> unsized -> incorrect size.

Furthermore, it is incomplete because I assume that "unsized" pointers point to element arrays of length 1. Therefore it may be safe to access with an offset of 2 but the guard will fail.

I will attempt to improve the completeness problem within a function by tracking `alloca` calls and calls to `malloc`. In some cases I can use this to determine the runtime size of the allocation and therefore insert a guard using the appropriate maximum index length. LLVM does have some metadata tags which can be used for exactly tracking this information ([dereferencable](https://llvm.org/docs/LangRef.html#dereferenceable-metadata)).

**How will you empirically measure success?**
I will evaluate using the [parsec](https://parsec.cs.princeton.edu/) benchmark suite and my primary evaluation will be on 1) the completeness of the analysis and 2) the overhead of runtime checking. I will evaluate the former by measuring how much I needed to rewrite the benchmark programs in order for them to complete successfully when running approximate safety guards.
**Team members:**
me

sampsyo:
Excellent! This sounds great to me.

For what it's worth, I wouldn't spend _too_ much time trying to fix the PARSEC programs so they don't trigger false positives. You might instead opt to just log all the possible memory safety violations (rather than crashing on the first one) and just report the counts.

sampsyo:
Closed by #126.

PR #126, opened 2019-11-13
Project 3: Dynamic Memory Safety

dzagieboylo:
Proj 3 submission for issue #114


content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+  [Drew Zagieboylo](https://www.cs.cornell.edu/~dzag/) is a 3rd year PhD student researching Security, Hardware Design, and Programming Languages. He enjoys rock climbing and gaming in his free time.
+"""
++++
+
+# Memory Safety in LLVM
+
+LLVM IR code is not generally memory safe.
+While certain *obviously bad* behaviors are
+disallowed, it is not hard to write code that
+may execute out-of-bounds memory accesses at runtime.
```
sampsyo: "At run time." https://capra.cs.cornell.edu/styleguide/#runtime (This also appears elsewhere in the post.)

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+Before a program might make an out-of-bounds access it has to
+acquire an out-of-bounds pointer. Usually, this means it
+executes a GEP whose result will then later be the argument of a
+load or store operation.
+
+Our approach in this implementation is to prevent the execution of
+any runtime GEP instructions that might lead to
+illegal memory accesses.
+*If a program can never acquire an out-of-bounds pointer,
+it can't violate memory safety.*
```
sampsyo: This is certainly true, but there's an implicit connotation behind it that probably isn't true: that GEPs are the only way to acquire an out-of-bounds pointer. For example, you can [bitcast](https://llvm.org/docs/LangRef.html#bitcast-to-instruction) an arbitrary integer to a pointer that, when dereferenced directly without a GEP, would violate memory safety.

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+between 0 (inclusive) and 10 (exclusive).
+
+To execute this check completely, we need to check that *all* index operands
+are legal. You may have noticed that our prior example actually has **two**
+index operands; we have to check both that `tmp` points to at least one
+integer array of size 10, *and* that `x` is a valid index for such an array.
+
+In general our algorithm for modifying LLVM code is this:
+
+```
```
sampsyo: Probably no need for a code block here—you can use a nice Markdown numbered list, like this:

1. Foo.
2. Bar.

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+1) Initialize the current type to be the type of the first operand.
+2) Initialize current operand to the first index operand.
+3) If possible, insert instructions to check if current operand is in bounds based on current type.
+4) Set current type to the next element type (e.g. if current type is *int[] the next type is int[]).
```
sampsyo:
```suggestion
4) Set current type to the next element type (e.g., if current type is *int[] the next type is int[]).
```

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+5) If there are no more index operands, exit.
+   Else, set the current operand to the next in the operand list and goto (3).
+```
+
+### GEP Checking: A Walkthrough
+
+In this section we'll walkthrough the above example in excruciating detail.
```
sampsyo: "walk through" (with the space) when it's a verb.

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+Based on the possible types that GEP expects there are only a few cases to handle.
+Intuitively, each type represents a container in some way and "indexing" into it
+should get us the type contained by the outer one.
+
+| Type | Next Element Type | Notes |
+|------|-------------------|----------|
+| t*   | t                 | For pointers, the next type is the type being pointed to |
+| [ size x t ] | t          | For arrays, the next type is the array element type |
+| < size x t > | t | Vectors, like arrays, have an element type |
+| struct { f1, f2,...,fn } | fi | i is the index value, LLVM requires this is a compile time constant |
```
sampsyo:
```suggestion
| struct { f1, f2,...,fn } | fi | i is the index value; LLVM requires this is a compile time constant |
```

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+InsertCheck(CurrentOp >= 0 && CurrentOp < NumElements); // (x >= 0 && x < 10)
+
+CurrentOp = <none>;
+//Done!
+```
+
+## Pointer Sizes and Tracking Allocations
+
+In the above examples, we could always tell how big our memory allocations
+were since they were allocated with static sizes. `int tmp[10]` comprises of
```
sampsyo: Just "comprises," not "comprises of."

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+that we can do. LLVM uses [`alloca`](https://llvm.org/docs/LangRef.html#alloca-instruction)
+instructions to allocate local variables. `alloca` takes an argument to
+determine how many elements must be allocated. If we keep track of the
+sizes of local allocations we can infer that the above code is safe if and only if:
+
+```C
+0 <= y < x //we'll assume x > 0 here
+```
+
+In our implementation, we simply keep around a map from allocaitons to their sizes.
```
sampsyo:
```suggestion
In our implementation, we simply keep around a map from allocations to their sizes.
```
(Consider running spell check.)

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+Additionally, we run our transformation as a function pass so
+it doesn't track interprocedural allocations. The main
+reason for this limitation is that, even if we knew
+the original allocation size for all callers, we would
+have to modify the function signature to communicate
+legal index values from the caller.
+This seemed both out of scope for our current project
+and a potentially questionable design decision. Should
+a compiler pass be modifying the signatures of
+potentially every function?
```
sampsyo: Indeed. For what it's worth, I think the sound (e.g., Valgrind or ASan) approach to this wouldn't require modifying function signatures—it would instead keep a gigantic hash table somewhere that maps addresses to size information. So when you try to index a pointer, the instrumentation would invoke the runtime to look up the actual address in this table to find its dynamic size. These hash table loads could be selectively eliminated when you have sufficient static information, but that would be an optimization rather than the primary mechanism as it is here.

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+
+## Soundness and Completeness
+
+Often, when trying to ensure a safety
+property you'd like to show that your results are either *sound* or *complete*.
+In our case, soundness would imply that any LLVM program which uses no
+"type unsafe" features and is compiled with our pass 
+executes *no GEP instructions which would generate out-of-bounds pointers*.
+Completeness, on the other hand, would imply that we can compile
+all programs and allow them to execute.
```
sampsyo: I would actually frame "completeness" in this context _slightly_ differently: Whenever one of our early exits occurred, the program was _actually_ about to do an illegal memory access. In other words, while the tool can "miss" bugs, when it signals a bug, one actually exists. I think this means the same thing but is a little more useful of a definition.

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+Often, these two problems combined, since `prices` may be accessed
+outside the scope at which `%45` is available and we therefore would need
+to modify the program to communicate this allocation information (potentially
+via an extra global variable).
+
+
+### Evaluation: Overhead
+
+We measured runtime overhead in terms of wall clock time purely because it
+was the simplest thing to instrument. In the following graph we report the
```
sampsyo: It's also probably the most relevant bottom-line metric!

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+which reports the mean runtime for both configurations. Error bars on that
+graph represent one standard deviation.
+
+<img src="overhead.png"/>
+
+In the case of the Ferret benchmark, our instrumentation caused the implementation
+to exit prematurely. Since the benchmark is quite large we did not have time to investigate
+why this was; it is possible that Ferret intentionally executes "unsafe" GEP instructions.
+Since our instrumentation did not cause bugs in any of the other implementations we find
+this to be a likely cause, but it does warrant further examination.
```
sampsyo: It would indeed be interesting to investigate… something like Address Sanitizer could help pin down the line of code where a safety violation *might* happen. It's unfortunately not unheard of for benchmarks to be full of UB.

content/blog/2019-11-13-gep.md/index.md
```diff
@@ -0,0 +1,458 @@
...
+
+
+Interestingly, the instrumented Canneal and Streamcluster benchmarks run ever so slightly
+faster, however this result is within the standard deviation and could
```
sampsyo: "faster; however, this" https://capra.cs.cornell.edu/styleguide/#however

sampsyo (changes requested):
Really nice work! This was a very clear and complete description of the problem space and the limitations of your solution. The evaluation was also very clearly described. Awesome!

content/blog/2019-11-13-gep.md/index.md
dz333: Added note pointing to the "soundness and completeness" discussion on why this isn't a complete solution

content/blog/2019-11-13-gep.md/index.md
dz333: Good point, I'll mention that and point to the other group that implemented that style of check.

sampsyo:
Yay; looks great!

PR #128, opened 2019-11-14
Project 3: MemPass

### LINKED ISSUE #111: Project 3 Proposal: Static Analysis for Memory Bugs
**What will you do?**
We will implement a pass that keeps track of use-after-free and double free errors.  We will generate a report of these errors and suggestions to improve the code.
 
**How will you do it?**
We will complete this pass by monitoring code that is allocated and checking when memory segments have been freed. If pointers are freed multiple times, or pointers are referenced after free, we want to make sure that we detect and return those errors as part of our static analysis.

In order to evaluate our success, we’ll run our tool on the BegBunch benchmark suite (https://llvm.org/pubs/2009-07-ISSTA-BegBunch.pdf) and keep track of how many memory leak errors our tool is able to detect.  We can also consider generating our own microbenchmarks to test the memory tool. We also know that @sgpthomas is doing the same project, so we’d be interested in seeing what bugs our different implementations are capable of reporting.

**Team members:**
@sameerlal @splashofcrimson

sameerlal:
Closes #111

### DIFF
--- a/content/blog/2019-11-13-MemPass/index.md
+++ b/content/blog/2019-11-13-MemPass/index.md
@@ -1,0 +1,168 @@
++++
+title = "MemPass"
+extra.author = "Eashan Garg and Sameer Lal"
+extra.bio = """
+Eashan Garg is a senior undergraduate studying CS and English.
+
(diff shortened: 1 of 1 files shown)


sameerlal:
Use #131  instead please

PR #131, opened 2019-11-14
Project 3: MemPass

### LINKED ISSUE #111: Project 3 Proposal: Static Analysis for Memory Bugs

splashofcrimson:
closes #111


content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+```c
+int *ptr, *ptr2;
+ptr = (int *)malloc(4 * sizeof(int));
+ptr[0] = 4;
+free(ptr);
+ptr2 = (int *)malloc(4 * sizeof(int));
+printf("%d", ptr[0]);
+```
+
+Since `ptr` has been freed, the second call to `malloc` (which is of same size as the first) is likely to allocate the free space `ptr` points to. Now, lets assume a malicious attacker managed to store a dangerous payload in this space. Since the program tries to use `ptr` after it has been freed, it might actually end up accessing the attacker's payload, not the original `ptr` data as intended!
```
sampsyo: let's

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+Double free errors occur upon consecutive `free()` call with the same memory address. For instance, consider the following code snippet taken from a Double-Free Toyota ITC test case:
```
sampsyo: "Toyota ITC" won't be familiar to most readers, so consider introducing it and linking to some sort of description.

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+``` c
+int main() {
+	char* ptr= (char*) malloc(sizeof(char));
+	free(ptr);
+	free(ptr); /* Double Free Error! */
+}
+```
+
+Here, `ptr` is allocated and then freed twice.  Generally, double-freeing a block of main memory will corrupt the state of the memory manager, resulting in a write-what-where condition. In this case, a malicious attacker may be able to write arbitrary dangerous code to the memory location that was freed twice.
```
sampsyo: I've never heard the phrase "write-what-where condition"! It might be worth explaining too. Or you can just say that this is a memory safety violation, which everyone should agree is probably bad.

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+int main() {
+    char* ptr= (char*) malloc(sizeof(char));
+}
+```
+
+The consequences of memory leaks is often degraded performance. Memory leaks reduce the available amount of memory in a system, and if not enough space is available, the program may slow down due to thrashing or stop altogether. 
+
+## Introducing MemPass
+
+Here, we present a analysis tool (creatively titled MemPass) for LLVM to detect use after free bugs, double free bugs, and memory leakage. Here's how it works.
```
sampsyo: an analysis tool

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+In practice, we can only detect memory leaks after the program has completely executed, while use-after-free and double-free errors can be detected during runtime and should result in the program halting to prevent subsequent malicious code from executing. We decided to only return warnings for this report to make testing more manageable.
+
+## Design Overview
+As mentioned at the beginning, there is a large breadth of possible memory safety vulnerabilities. In order to tackle a modest subset of these bugs, our strategy of choice is a simple dynamic analysis pass over the LLVM IR.
+
+MemPass inserts instrumentation after relevant memory allocation instructions, recording the relevant addresses. Anytime the program attempts to either access or deallocate memory, MemPass will then check if those addresses are still available for use. If not, the program will throw a warning.
```
sampsyo: Any time

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+With `i8*` pointers, all MemPass needs is the size of the memory chunk that a load or store plans to interact with. While this data is not immediately accessible, LLVM provides a handy DataLayout class. After grabbing the type of the element that the *original* pointer points to, MemPass can extract its size and pass that to our library functions.
+
+Finally, we need to actually check if we are accessing memory that is available as per the allocation table. MemPass takes the difference of the address in question with every pointer address in the allocation table, and compares it to the appropriate sizes. If the address is within the bounds of a chunk of memory, then we are fine. Otherwise, the program will emit a use after free warning, and continue to look for more vulnerabilities.
+
+> **Note**: These pointer address comparisons are somewhat sketchy, there likely may be a better approach to memory access checking that we were unable to find.
```
sampsyo: Comma splice—replace the , with a ;.

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+ret i32 0
+```
+
+First, the program accesses more memory than what was allocated by the programmer through calls to `malloc`. If MemPass compares the address accessed by one of the first `store` instructions with the allocation hash, it wouldn't find the address and emit a false-positive warning.
+
+In reality, programs also allocates stack frame memory with `alloca` instructions. In order to handle these extra allocations, MemPass adds extra instrumentation here, and adds the stack frame addresses/sizes onto the allocation table.
+
+However, there's another problem. Some of the load and store instructions use pointers of arbitrary types. If MemPass doesn't know what the pointer types are, it can't pass those addresses to the runtime library for evaluation.
+
+A solution that MemPass employs is to insert `bitcast` instructions after every load or store instruction, converting the address pointer from its arbitrary type to an 8-bit integer pointer. Since we're only comparing addresses and not the actual values at these addresses, this should work somewhat well. However, the breadth of types in LLVM could allow for some imprecision here.
```
sampsyo: Hmm… I don't quite see what the imprecision is. If you supply the base address (as an `i8*`) and the access size, I think you should have everything you need to check safety? Maybe I'm missing something.

content/blog/2019-11-14-mempass/index.md
sampsyo: TBH I don't see this as sketchy. If your table contains (base address, size) pairs, then it does seem like the natural thing to do with a given access is to look at each pair and see whether the new address is within range for that allocation. Not sure what else you could do!

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Program Termination
+
+On program termination, we need to check the allocation table for any remaining memory that has not been freed. In order to differentiate stack memory that was allocated with `alloca`, any memory allocated with `malloc` is written to a separate file, along with any pointer addresses passed to `free`. Now, MemPass just searches this file and emits any malloced addresses that were not freed. This is compiled into a final report, listing all of the memory safety vulnerabilities (among double free, use after free, and memory leak) that were detected throughout the execution of this program, along with their line numbers.
+
+## Implementation extras
+
+Another possible implementation scheme we considered was the use of a *deallocation* hashmap in addition to the allocation hashmap, to store memory that has been freed. This way, MemPass does not need add instrumentation after `alloca` instructions, it just needs to store memory addresses allocated with malloc. However, every time memory is allocated or deallocated, MemPass must check addresses in the other map to ensure there are no overlaps.
```
sampsyo: Comma splice again at "instructions, it"

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+Both for the proposed framework and our current one, some sort of segmentation tree implementation to store memory bounds as intervals and compare them quickly. However, the overhead of building this tree might not be worth the benefits for small programs.
```
sampsyo: The first sentence seems to be missing a verb?

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+Another implementation idea that would have been much more effective in hindsight would be to find some way invalidate pointers after they are freed. This way, MemPass does not need to add instrumentation after every load and store (the program can exit when accesssing a specific 'invalid' pointer).
```
sampsyo:
```suggestion
Another implementation idea that would have been much more effective in hindsight would be to find some way invalidate pointers after they are freed. This way, MemPass does not need to add instrumentation after every load and store (the program can exit when accesssing a specific "invalid" pointer). 
```

sampsyo: That actually seems pretty hard to accomplish! You'd need to either:

- Find all the places where that pointer exists and update it to zero or something.
- Somehow ensure that all allocations are on separate virtual memory pages, and unmap every page when the allocation it contains is deallocated.

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+| Perm           | 2.322     | 0.186     |
+| Queens         | 8.198     | 0.252     |
+| Quicksort      | 19.060    | 0.240     |
+| RealMM         | 0.446     | 0.013     |
+| Towers         | 2.852     | 0.279     |
+| Treesort       | 5.948     | 0.305     |
+
+> **Note:** Some programs had an almost negligible runtime, which we record as 0.000 in the charts.
+
+MemPass's overhead can be anywhere from a 100x slowdown to over 1000x in the case of Bubblesort. This can be attributed to the fact that MemPass adds instrumentation not only after `malloc` and `free` calls, but also after every `load` and `store`. The benchmarks in question (especially FloatMM, IntMM, and RealMM) do many matrix computations, which end up blowing up the performance cost of our instrumentation significantly. In hindsight, restricting instrumentation to only `malloc` and `free` would significantly reduce the overhead of MemPass. As explained before, we could find some clever way to invalidate pointers once they are freed, thus causing the program to exit when trying to access an invalid location in memory.
```
sampsyo: I don't think 100x is crazy at all for a heavyweight memory instrumentation pass. You can compare to the overheads for Valgrind for example.

You might also find that the overheads would be lower if you ran your pass after some standard `-O2` optimizations, which would reduce the number of loads & stores by keeping some things in SSA variables.

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Benchmark Testing: Correctness
+We evaluated the correctness of MemPass on subsets of the Toyota ITC and SARD-100 benchmark tests. The Toyota benchmark tests consist of a family of memory tests, and two test suites that we used are `Double Free` and `Memory Leak` tests.  The Sard-100 benchmark tests are similar, and we used the `cwe-401-memory-leak`, `cwe-415-double-free` and `cwe-416-use-after-free` suites.
```
sampsyo: Again, it would be a good idea to link to the benchmark suites. Also, stay consistent about whether you capitalize "SARD."

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+>**Note**: We wanted to use the Invalid Memory Access (Use After Free) test set in Toyota as well, but we kept running into segmentation faults with the tests and struggled to debug them.
+
+In general, the`Double Free` and `cwe-415-double-free` benchmark tests consist of normal double free errors, freeing in constant/variable if statements, freeing in a function, freeing in conditional while loops, and freeing in for loops.
```
sampsyo:
```suggestion
In general, the `Double Free` and `cwe-415-double-free` benchmark tests consist of normal double free errors, freeing in constant/variable if statements, freeing in a function, freeing in conditional while loops, and freeing in for loops.
```

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+The `Memory Leak`, `cwe-401-memory-leak`, and `cwe-416-use-after-free` benchmark tests consist of a series of tests such as allocating memory without freeing, allocating in conditional statements, freeing based on function return values, allocating memory in mutually recursive functions and various branching scenarios.
+
+A comprehensive description of each test case can be found on the respective pages: [Toyota ITC](https://github.com/regehr/itc-benchmarks) and [SARD-100](https://samate.nist.gov/SRD/testsuite.php#sardsuites).
+
+### Benchmark Results
+In [Detection of Security Vulnerabilities in C Code using Runtime Verification](https://nikolai-kosmatov.eu/publications/vorobyov_ks_tap_2018.pdf), the authors provide benchmark test results for E-ACSL, Google Sanitizer, and RV-Match on both the Toyota-ITC and SARD-100 test suites.
```
sampsyo: Please add a little more context here to explain what the percentages are—what is divided by what?

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+| Defect Type          | E-ACSL | Sanitizer | RV-Match |
+|----------------------|--------|-----------|----------|
+| Dynamic Memory Tests | 94%    | 78%       | 94%      |
+
+Unfortunately, the above statistics are not that useful, as our LLVM pass is only able to target double free, memory leaks and use-after-free errors. Still, within the Dynamic Memory tests, we considered the `Double Free` and `Memory Leak` suites of tests. `Double Free` contains 12 cases, and `Memory Leak` contains 18. Over these test suites we achieved the following results:
```
sampsyo: I still don't quite understand why these statistics are "not all that useful," but perhaps I would if I knew what the percentages meant or what "Dynamic Memory Tests" are.

content/blog/2019-11-14-mempass/index.md
```diff
@@ -0,0 +1,292 @@
...
+
+> **Note:** Our implementation attempts to achieve some sense of completeness, as we only emit a warning if a program tries to interact with memory that was not allocated to begin with. However, this dynamic analysis comes at the cost of soundness, especially since we only interact with one execution path at a time. In addition, due to casting errors we can end up with some false positives in certain cases.
+
+### Dynamic vs Static Analysis
+
+While a dynamic analysis is certainly interesting and useful, it's difficult to analyze certain memory bugs such as memory leaks, since they cannot be detected until the program terminates. In addition, a dynamic analysis can only check individual executions of a program, and therefore might miss bugs with programs that have a large number of inputs.
+
+A static analysis that uses some sort of use-def chain would be another interesting way to triage these vulnerabilities, and would relax much of the overhead that our method produces. In addition, it would be able to analyze all possible execution paths at once, and therefore complete a more 'sound' analysis of the various bugs that may be present in the program.
```
sampsyo:
```suggestion
A static analysis that uses some sort of use-def chain would be another interesting way to triage these vulnerabilities, and would relax much of the overhead that our method produces. In addition, it would be able to analyze all possible execution paths at once, and therefore complete a more "sound" analysis of the various bugs that may be present in the program.
```
Use double quotes.

sampsyo (changes requested):
This is really nice work! It's an ambitious implementation that covers lots of cases, and you did a great job finding reasonable buggy benchmarks to run. Awesome!

sampsyo (approved):
Looks great; nice work!

PR #132, opened 2019-11-14
Project 3: Run-time execution profiling

### LINKED ISSUE #97: Project 3 Proposal: Run-time execution profiling
**What will you do?**
We are planning to collect run-time information including number of LLVM instructions, basic blocks and memory instructions being executed. We can also look into expensive instructions like multiplication and branches. We also propose to count the number of times each function is called.

**How will you do it?**
Profiling at run-time can be tricky due to change in control flow, hence LLVM compiler cannot tell the dynamic behavior at compilation time. We can however, do a basic block analysis where we have a single entry and exit point implying that the instructions within a basic block will execute. In order to collect run-time statistics, we can introduce a profiling function in each basic block of the IR code using IRBuilder. The profiling function will modify global variables (counters for the statistics being measured) atomically to ensure well defined behavior even in multi-threaded applications.

One idea could be to profile benchmarks with and without optimization directives like o3. We would expect optimized benchmarks to have less LLVM instructions or less expensive operations than the unoptimized version.
Another thing to do would be to run standard c based profilers like gprof which gives us the number of times a function is called and compare it with our output.

**Team members:**
@elainejiang @tissue3

neiladit:
closes #97


content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Design Overview
+
+The LLVM architecture looks like this:
+
+<img src="llvm.jpg" style="width: 90%">
+
+
+
+IR represents intermediate language. We only need to add one pass that changes the existing IR to another IR (which includes the desired properties like profiling information), following instructions to add an LLVM pass in our instructor, [Adrian](<https://www.cs.cornell.edu/~asampson/>)'s [blog post](<https://www.cs.cornell.edu/~asampson/blog/llvm.html>). We also took help from the github repo [atomicCounter](<https://github.com/pranith/AtomicCounter/blob/master/AtomicCountPass/AtomicCount.cpp>) where it counts the number of all atomic instructions in the program atomically!
```
sampsyo: Capitalize "GitHub."

sampsyo: What does it mean to count atomics "atomically"? Does it mean that the increments to the global counter variables are atomic updates?

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+We start from [ModulePass](<http://llvm.org/docs/WritingAnLLVMPass.html#the-modulepass-class>) and then access functions inside a Module, Basic Blocks per function and instructions per Basic Block to obtain all information. At the beginning and ending of each module, we also call custom method `initialize` to create global variables in IR  and `finalize` to print global variables. The structure of our atomic counting pass looks like this:
```
sampsyo: Try to be a little more consistent about your capitalization. If I were you, I'd capitalize "BasicBlock" and not use a space when referring specifically to [the LLVM class](https://llvm.org/doxygen/classllvm_1_1BasicBlock.html) but _not_ capitalize "basic block" and use a space when you're referring to the general concept.

sampsyo: call _a_ custom method

sampsyo: I'm a little confused about why this is referring to "your" "atomic counting" pass. The overview doesn't say that you're counting atomic ops, although the open-source code you referred to does.

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+    virtual bool runOnModule(Module &M); //when there is a Module
+    virtual bool runOnFunction(Function &F, Module &M); //called by runOnModule
+    virtual bool runOnBasicBlock(BasicBlock &BB, Module &M); // called by runOnFunction
+      
+    bool initialize(Module &M); //create global variable
+    bool finialize(Module &M); //print global variable
+      
+    void createInstr(BasicBlock &bb, Constant *counter_ptr, int num);
+      
+    vector<string> atomicCounter; //keep global variable names for profiling. e.g. instr counter
```
sampsyo: How does a `vector<string>` suffice? How do you know which variable is which in the vector?

Also, why not use actual `Value*` or `GlobalVariable*` pointers instead of strings?

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+  };
+```
+
+
+
+## Implementation
+
+### Basic Execution Information
+
+We first create global variable `llvmInstrAtomicCounter` and `basicBlockAtomicCounter` with [GlobalVariable](<https://llvm.org/doxygen/classllvm_1_1GlobalVariable.html#a3ef813d6bda7e49e31cb6bf239c4e264>) constructor.
```
sampsyo: Hmm; this seems to differ from the code listing above, which doesn't list these variables.

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+```c++
+new GlobalVariable(M, I64Ty, false, GlobalValue::CommonLinkage, ConstantInt::get(I64Ty, 0), atomicCounter[i]);
+```
+
+ Then in each basic block we obtain pointer to the global variable names with [getOrInsertGlobal](<https://llvm.org/doxygen/classllvm_1_1Module.html#abd8f7242df6ecb10f429c4d39403c334>) method. After getting the instruction number in each block, we create atomic addition with [AtomicRMWInst](<https://llvm.org/doxygen/classllvm_1_1AtomicRMWInst.html#abf7e0649c7f272cc49165e579be010a5>) constructor. The complete code looks like this:
```
sampsyo: obtain _a_ pointer

sampsyo: It doesn't seem necessary to re-run "getOrInstertGlobal" while iterating over each basic block… you can keep a reference to it and reuse that.

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+    // create atomic addition instruction
+    createInstr(bb, bbCounter, basic_block);
+    createInstr(bb, instrCounter,instr);
+    
+    return true;
+}
+
+```
+
+Finally, we print the profiling results by first create a function `printf` by specifying input argument and output argument types. Then transverse the `main` function. Right before `return` of the last block in `main`, create string variable corresponding to variables we are tracing and get the pointer to that string variable. Also, get the number variable by first get the pointer pointing to the global variable and then create a load instruction to get the actual pointer to the value. Finally, we can call printf function with input arguments.
```
sampsyo: - by first _creating_
- Then _traverse_ the
- _we_ [or some other noun] create string _variables_ corresponding
- Also, _we_ get
- by first _getting_
- and then _creating_
- we can call _the_ `printf` function

sampsyo: I'm a little lost about the high-level purpose of this work. Why are there string variables involved? Why isn't the template argument to `printf` just a constant string?

The specific mechanics of exactly how you generated the instructions are less important than making it clear what the instrumented program does and why.

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+```c++
+    RBuilder<> Builder(M.getContext());
```
sampsyo: IRBuilder?

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+    createInstr(bb, instrCounter,instr);
+    
+    return true;
+}
+
+```
+
+Finally, we print the profiling results by first create a function `printf` by specifying input argument and output argument types. Then transverse the `main` function. Right before `return` of the last block in `main`, create string variable corresponding to variables we are tracing and get the pointer to that string variable. Also, get the number variable by first get the pointer pointing to the global variable and then create a load instruction to get the actual pointer to the value. Finally, we can call printf function with input arguments.
+
+```c++
```
sampsyo: No need to _both_ indent _and_ have Markdown code fences.

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+    }
+    return true;
+}
+```
+
+
+
+### Expensive Operations Information
+
+The whole part follows the same flow as basic execution information except we need to distinguish the instruction type and increment corresponding counter on block basis. Therefore in each `runOnBasicBlock` method, we need the following lines:
```
sampsyo: - I'm not sure what "The whole part" refers to… maybe "this part"?
- on _a_ block basis

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+```c
+auto &functionList = M.getFunctionList(); // gets the list of functions
+for (auto &function : functionList) { //iterates over the list
+    Value *atomic_counter = new GlobalVariable(M, I64Ty, false, GlobalValue::CommonLinkage, ConstantInt::get(I64Ty, 0),  function.getName()+".glob"); // create a global variable, name it based on the function name
+} 
+
+```
+
+Next we want to insert atomic counters at the start of each function definition. This ensures that irrespective of multiple return points in the function we can always increment the counter for it at the beginning. We start with the entry basic block in the function which is given by the iterator `F.begin()`.  To insert it at the top of the basic block we use `getFirstNonPHI()` which returns the first instruction, which is not a PHI node. We insert an atomic add instruction similar to other profiling instructions.
```
sampsyo: the first instruction _that_ is not a PHI node (no comma)

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+Next we want to insert atomic counters at the start of each function definition. This ensures that irrespective of multiple return points in the function we can always increment the counter for it at the beginning. We start with the entry basic block in the function which is given by the iterator `F.begin()`.  To insert it at the top of the basic block we use `getFirstNonPHI()` which returns the first instruction, which is not a PHI node. We insert an atomic add instruction similar to other profiling instructions. 
+
+
+
+
+
+
+## Hardest Parts
+
+1. For people who are new to LLVM, instructions are hard to find and follow. Searching on google can help if you know what you're looking for. It's difficult to get helpful information unless our search is confined to existing functions. Even though LLVM documentation is pretty exhaustive, it has too many functions to go through and the lack of examples can be off putting. Tutorials and existing backbone codes on Github can be really handy in these scenarios, which we took advantage of. It not only helped us implement specific functions like `printf` but also establish a structure to our IR pass.
```
sampsyo: - capitalize Google
- _your_ search?

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+
+
+
+
+
+## Hardest Parts
+
+1. For people who are new to LLVM, instructions are hard to find and follow. Searching on google can help if you know what you're looking for. It's difficult to get helpful information unless our search is confined to existing functions. Even though LLVM documentation is pretty exhaustive, it has too many functions to go through and the lack of examples can be off putting. Tutorials and existing backbone codes on Github can be really handy in these scenarios, which we took advantage of. It not only helped us implement specific functions like `printf` but also establish a structure to our IR pass. 
+2. String manipulations: I am not sure if this is the right term to use, but LLVM seems to have 2 string types - Twine and StringRef. `getName` on a function returns a StringRef. In order to make a custom name I perform `F.getName()+"name"` which returns a Twine. But the function `getGlobalVariable` only accepts StringRef. Twine has a function `str` which can be used to convert it into string. Even though this is a straight forward solution, it ended up taking time to figure out the problem and looking into the documentation of these classes.
```
sampsyo: - straightforward

sampsyo: Maybe this is already clear, but there's an important difference between `Twine` (which is an actual string object containing actual characters, like a C++ `std::string`) and `StringRef` (which is just a pointer to characters stored elsewhere, like a C `char*`). When you delete a `Twine`, any other pointers to the same data become invalid. When you delete a `StringRef`, nothing really happens and no character memory is freed.

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Evaluation and results
+
+### Gcov profiling tool
+
+To validate and benchmark our profiling results, we use `gcov` testing tool which can be used as a profiler to give performance statistics. We first compile the code using gcc flags required for `gcov` : `gcc -fprofile-arcs -ftest-coverage foo.c` . Now we run `gcov` with relevant flags to give us statistics to compare with our profiler:
+```
+gcov -b -c -f foo.c
+```
+We ran the [Phoenix](https://github.com/kozyraki/phoenix) benchmark suite and used `gcov` to profile statistics of function calls. The makefile initially had optimization flag -O3 but this might lead to incorrect numbers of function calls due to optimizations, hence we compile without any flags.
```
sampsyo: "Hence" is not a conjunction, like "however": https://capra.cs.cornell.edu/styleguide/#however

Use "so" instead or start a new sentence at "hence."

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+#### Kmeans
```
sampsyo: Any particular reason you picked Kmeans as the one benchmark to study in more detail, or did you just pick one arbitrarily? (It might be worth clarifying this in the text.)

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+parse_args: 1
+generate_points: 2
+add_to_sum: 23000
+find_clusters: 23
+get_sq_dist: 230000
+calc_means: 23
+dump_matrix: 1
+main: 1
+```
```
sampsyo: It's at least worth mentioning in the text that the numbers match. :smiley:

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Results
+
+The following results is for sequential execution of the benchmarks. We have reported the instruction counts in the following table:
```
sampsyo: results _are_

content/blog/2019-11-13-llvm-profiling/index.md
```diff
@@ -0,0 +1,336 @@
...
+
+| Benchmark | llvm instruction count |basic block count | multiplication count | memory operation count | branch operation count|
```
sampsyo: Capitalize LLVM.

sampsyo (changes requested):
Thanks for the detailed writeup! Everything was very clear, for the most part. One thing I wished for in this report (although there's no need to add it now) was some sort of performance measurement—it would have been nice to know how much the slowdown cost, especially in a multithreaded context.

Also, here's one high-level writing suggestion: I don't think most of the code listings from your implementation add much to the report. You can always link to the code on GitHub if people are really interested, but a simple description of what's actually happening is quite a bit easier to ingest than reading the C++ code—which generally bears out what the reader was expecting to see anyway.

sampsyo:
One other thing I forgot to mention: please use Title Case in the title of your post.

content/blog/2019-11-13-llvm-profiling/index.md
neiladit: Picked arbitrarily. Mentioned this in a line. Thanks!

content/blog/2019-11-13-llvm-profiling/index.md
tissue3: I don't get your question. These are just name of global variables. We can add value by name at each basic block, load values and print them out at the end of main function.

content/blog/2019-11-13-llvm-profiling/index.md
tissue3: That's in initialize function. We didn't post it in the blog.

content/blog/2019-11-13-llvm-profiling/index.md
tissue3: By "atomic counting" we meant counting instructions atomically, _not_ counting atomic ops.

content/blog/2019-11-13-llvm-profiling/index.md
tissue3: I feel I understand your question on `vector<string> atomicCounter;` now! Yes, we could have done that. Thanks for your advice.

sampsyo (approved):
Looks great; thanks for the updates!

PR #157, opened 2019-12-01
blog post: Equivalence Modulo Inputs

rolph-recto:


rolph-recto:
looks like this is also pulling commits for the loop perforation blog post. should i fix or is this fine?

PR #168, opened 2019-12-18
Project 4 Evaluating the Costs of Physical Addressing

### LINKED ISSUE #147: Project 4 Proposal: Generating Code for Noncontiguous Memory Allocation
**What will you do?**
This project is targeted at compilation for a system *without* virtual address translation. We assume an architecture where processes are allocated physical memory addresses (and memory protection is implemented orthogonally; e.g., using tagged memory).

The main challenges for this kind of compilation are instructions which assume physically contiguous memory addresses and use this assumption for address calculation. Some primary examples are:
 - Stack allocation and accesses
 - Large object allocation (i.e., when the memory allocator returns a large contiguous *virtually addressed* region made up of multiple non-contiguous *physical pages*)
 - Dynamically Linked Libraries & Repositionable (? the word people normally use for this is escaping me) code

In this project, we will focus on the first two problems.

Our goal is to measure the overhead of the runtime code required to generate correct code. Ideally, this overhead should be less than the overhead of typical address translation. However, we will likely not evaluate the performance in that case since disabling address translation is actually quite hard to set up on typical CPUs + OSs.

**How will you do it?**

 - Stack allocation

We will statically compute the size of each function's stack frame. Then, we will modify the stack prologue and epilogue to check if the currently allocated stack page has enough space for the next frame. If not, it will call a custom stack allocation function, which we will implement. Similarly, at the end of the function, the epilogue will deallocate the current stack frame.
Some stack frames have dynamic size requirements. If possible, we will insert instructions to compute the dynamic stack frame size during the function prologue.

 - Large Objects

Heap Allocations (or stack frames) whose size is greater than 1 page (4KB) require significant recompilation. We believe this is better suited to source level transformation rather than modifying code generation. While modifying any real language to support the required transformation would be prohibitively time consuming, we can write a C++ object which implements array allocation and dereferencing using a page table like structure. In this way, we can manually recompile some benchmarks to use these array-like objects instead of actual arrays. This can provide us with some proof of concept numbers for overhead incurred by this technique.

Our first evaluation will focus on stack overheads. We will measure the runtime performance of a number of microbenchmarks both designed to stress stack allocation and some which we expect to have almost no overhead due to very little stack allocation. In addition, we will measure the performance overhead on a number of standard C benchmarks.

Next, we will report the number of large (> 4KB) allocations to summarize how accurate our overhead measurements are. In the face of few large allocations, our stack results are likely to reflect accurate upper bounds on performance overheads.

Lastly, using the previous evaluation, we will choose a small number of benchmarks which actually do allocate and access large arrays frequently. Then we will manually modify their source to use arrays with fixed size allocations and measure the overhead of using these kinds of arrays rather than true contiguous memory. 

This evaluation is investigative and meant to determine if these overheads are, in practice, less than that of using hardware accelerated virtual addressing. Ideally, we would measure these numbers on a machine with physical addressing, but that is a difficult experimental setup to establish (e.g., for all intents and purposes you can't use Linux).

**Team members:**
@rachitnigam if he is interested?

dzagieboylo:
Resolves #147


content/blog/2019-12-18-novm/index.md
```diff
@@ -0,0 +1,317 @@
...
+There are many different proposals (and even some usable implementations)
+of _tagged memory_ architectures, where physical memory locations are associated with
+_tags_ that control how those locations can be accessed by software.
+Some examples include: the [CHERI capability architecture](https://www.cl.cam.ac.uk/research/security/ctsrd/cheri/);
+the [PUMP processor for software-defined metadata](https://dl.acm.org/citation.cfm?id=2694383);
+and the [secure information flow CPU, Hyperflow](https://dl.acm.org/citation.cfm?id=3243743).
+Instead of relying on a process' inability to address memory,
+these designs use hardware to efficiently check whether or not a memory access
+is allowed by the system's security policy. In these designs,
+the protection provided by virtual addressing is either mostly or completely redundant.
```
sampsyo: FWIW, the other technique generally bandied about (aside from tagged memory) is some way of _decoupling protection from translation_. Here's [a recent paper](https://dl.acm.org/citation.cfm?id=3080209) that has that flavor. The rough idea is that the enforcement mechanism for memory protection might look different—and be quite a bit cheaper—than traditional page tables & TLBs if it didn't also have the responsibility of telling you where to find the memory.

content/blog/2019-12-18-novm/index.md
```diff
@@ -0,0 +1,317 @@
...
+the OS needs to find a large enough space inside the memory to allocate. Due to the
+presence of [fragmentation](https://en.wikipedia.org/wiki/Fragmentation_(computing)),
+it is possible that no such region exists. In that case, `malloc` returns `0` and,
+in all likelihood, the program explodes.
+
+Remember that such fragmentation was present
+with virtual addressing as well, but the OS could stitch together various fragmented segments
+to form a single virtual allocation. Therefore, programs should strive to allocate memory in fixed-size chunks;
+essentially, they should assume that the OS can only allocate them pages of physical memory
+and it's _their job to stripe datastructures across them_.
```
sampsyo: It probably goes without saying that it's now the application's job to do paging (i.e., moving data between disk and RAM) in low-memory conditions too.

content/blog/2019-12-18-novm/index.md
```diff
@@ -0,0 +1,317 @@
...
+
+# Experimental Setup
+
+To evaluate the impact of software changes required in lieu of virtual addressing,
+we ran experiments with the following configurations. First, we ran all of our tests
+on a computer with 8 Intel i7-7700 CPUs clocked at 3.60GHz, with 32GB of physical memory, running Ubuntu 16.04.
+Secondly, we followed the [guidlines](https://llvm.org/docs/Benchmarking.html) provided by llvm to reduce
```
sampsyo: Capitalize LLVM.

content/blog/2019-12-18-novm/index.md
```diff
@@ -0,0 +1,317 @@
...
+Finally, we assumed that some reasonable amount of stack could be pre-allocated contiguously,
+even on a physically addressed machine. We chose 32KB since that was approximately the smallest
+sized stack required to execute the benchmarks normally.
+
+Unfortunately, we could not actually execute any tests using physical addressing,
+since there is no reliable method for allocating physical memory in user space.
+While there are several [proposals](https://blog.linuxplumbersconf.org/2017/ocw/system/presentations/4669/original/Support%20user%20space%20POSIX%20conformant%20contiguous__v1.00.pdf) for how to implement these
+features, they aren't currently supported in Linux. Therefore, our results
+are overhead measurements that represent worst-case performance; we don't actually
+expect any of our tests to result in speedups.
```
sampsyo: It does seem like you could get part of the way there by allocating a few hugepages and then "mallocing" from that pool…

content/blog/2019-12-18-novm/index.md
```diff
@@ -0,0 +1,317 @@
...
+gcc, but it essentially implements the algorithm described above, modulo some tricks
+for making the common case fast and maintaining its own small free list for stack pages.
+
+## Overhead of Stack Checking
+
+We evaluated the performance impact of using split stacks on two microbenchmarks
+designed to be bottlenecked on function calls, which respectively do and do not
+trigger run-time memory allocations.
+The first microbenchmark was a naive program to compute the 50th
+number in the fibonnacci sequence without memoization; this did not require
```
sampsyo: Capitalize Fibonacci. (Also find/replace below.)

content/blog/2019-12-18-novm/index.md
```diff
@@ -0,0 +1,317 @@
...
+We should note that the StreamCluster benchmark actually _sped up_ when using split-stack;
+likely this is some sort of memory alignment effect à la [Mytcowicz et al.](https://dl.acm.org/citation.cfm?id=1508275).
+In any case, we should probably consider this impact to be negligible.
+
+Of these benchmarks, only Ferret actually required dynamically allocating stack space.
+Each of these allocations was 618 KB, which is a potential concern. It is unclear, in a real
+system using only physical addressing, whether or not allocations of this size would
+be frequently servicable or not. I hypothesize that real systems with many gigabytes or terabytes
+of memory with even severe fragmentation will be able to regularly respond to allocations
+in the kilobyte range; however, proving this is future work.
```
sampsyo: It does seem like such a futuristic system would probably want to hand out memory to applications in largish chunks. 4 KB in particular seems like too small an increment to allocate to a modern application at a time. I can see this working by handing out biggish chunks, even when the application doesn't need them, and then letting userspace manage the fragmentation within those chunks.

content/blog/2019-12-18-novm/index.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Custom Array Implementation
+
+We implemented our `Array` using a tree-based datastructure that mimics
+the functionality of page tables in a virtual address environment.
+We support up to three levels of allocation, where the final
+level contains data and all previous levels contain pointers to
+other pages. In the following diagram, the **L1** page contains
+pointers to **L2** pages, which contain pointers to the **L3** pages.
+Each **L3** page contains the actual object array data.
```
sampsyo: You might be interested to read about ["z-rays"](http://www.cs.utexas.edu/~mckinley/papers/zrays-pldi-2010.pdf).

content/blog/2019-12-18-novm/index.md
```diff
@@ -0,0 +1,317 @@
...
+and no **L2** or **L3** pages will be allocated.
+
+## Evaluating Tree Overheads
+
+We had to modify the Blackscholes benchmark slightly to replace the calls
+to `malloc` to use our C++ `Array` objects. This only involved modifying
+the array allocation and deallocation since we overloaded the `[]` operator
+for normal array dereferences.
+Blackscholes only uses pointer arithmetic for allocation purposes so we didn't
+need to modify any other instructions.
```
sampsyo: Overall, this exercise seems to imply that this idea is probably best suited to languages with nontrivial runtimes—like Java or .NET or whatever, where you can imagine transparently swapping in different implementations of the basic standard-library data structures that everybody uses.

sampsyo (changes requested):
This was really cool, Drew! Nice work on the implementation, and also great work making the findings very clear in the writeup. I think I only have one little writing suggestion within (and a few random thoughts on the direction).

sampsyo (approved):
Yay!

dz333:
Thanks for all the pointers to papers! And I agree pretty much spot on with your comments

PR #177, opened 2019-12-21
Project 4: Finding Redundant Structures in Data Flow Graphs

### LINKED ISSUE #141: Project 4 Proposal: Data flow graph coverings
**What will you do?**
We will find snippet coverings of data flow graphs (DFGs). That is, given a program's DFG we want to find subgraphs ("snippets") that share structure (up to some definition of snippet equivalence and cost model).

**How will you do it?**
We will be working on programs in LLVM IR because it is an in-memory representation of a DFG.

1. As a first step, we will use a fixed, manually crafted set of small template snippets. The problem formulation then becomes: given the set of snippets and an input program, find the components of the program (at the IR level) that match some snippet. If snippets overlap, we can only chose one.

2. Then, the more interesting step is to generate snippets from the source program. Given a source program and some bounds (how many snippets or nodes per snippet, for example), what snippets successfully cover the most code? We will analyze the performance of this with variants on a greedy algorithm that takes the "best" snippet, for various components of "best." Some ideas for finding snippets:
	- A fixed sampling of code snippets generated from some training set
	- Naive sampling of subgraphs from the source program itself
 	- Variance-maximizing sampling: every time you select a sample, down-weight future samples by similarity to the newly selected one.

3. We will also explore alternate cost-models for mapping one DFG onto another.

4. We will also explore abstracting components of DFGs. More precisely: for each subgraph, we will evaluate the effect of collapsing connected components on the cost of matching with snippets.

Using LLVM allows us to use a large amount of existing benchmarks and applications. We may also evaluate on randomly-generated DFGs.

We plan to use the following single-program metrics: 
1. Static code coverage: the fraction of a program's code that can be assigned to a snippet (at the IR-instruction level).
2. Dynamic code coverage: the fraction of instructions from an execution on a fixed input that are assigned to a snippet (to capture whether we successfully find snippets in looping structures).
2. For each subgraph size (from one vertex up to the number of vertices in the DFG), the fraction of possible subgraphs snippets that are found in the DFG.

As a first step, we'll evaluate these metrics on small benchmarks (such as selected PARSEC benchmarks, or emBench). For a stretch goal, we'd also like to try running on a larger application (for example, maybe even some of LLVM itself!).

**Team members:**
@orichardson @avanhatt

avanhatt:
Closes #141


content/blog/2019-12-18-dfg-cover/index.md
```diff
@@ -0,0 +1,383 @@
...
+To consider this as a general learning problem, imagine that there's some
+underlying distribution $\mathtt{Programs}$ of programs that people write; we
+can now cast our work as a solution to the optimization problem of finding:
+
+$$ \arg\max_{\mathcal H}\left( \mathop{\mathbb E}\limits_{\mathcal G\sim \texttt{Programs}}~ \mathcal S_{\mathcal H}(\mathcal G) - \text{Cost}(\mathcal H) \right)$$
+
+where $\text{Cost}(\mathcal H)$ is the additional cost incurred by choosing to
+accelerate the subgraph stencil $\mathcal H$, which is higher for larger
+subgraphs. In this learning analogy, finding common DFGs for a particular
+collection of programs is training data.
```
sampsyo: Hmm, I'm not _quite_ following the analogy… is the training data DFG/program pairs? Where would the training data come from? Maybe it's just too late at night but for some reason I wasn't quite able to follow what's going on in this whole learning-based formulation…

content/blog/2019-12-18-dfg-cover/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+Rather than solve this optimization problem in closed form, we optimize for
+the heuristics of finding candidate subgraphs based on specifying (1) the size
+of each subgraph, and (2) the number of subgraphs we can choose. These two
+heuristics are effectively regularization knobs that could be used in future
+work to automate the entire optimization.
```
sampsyo: Huh, I also don't get the connection between these parameters and regularization. Is it really regularization if you're just simplifying the problem (i.e., making it tractable) by making some assumptions? Is there some danger of overfitting that I'm missing here?

content/blog/2019-12-18-dfg-cover/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Implementation strategy
+We first instrument an LLVM module pass that writes out a JSON representation of
+the DFG. Our Python module then explores candidate subgraphs using a combination
+of our heuristics, and the out-of-the-box graph isomorphism tooling.
```
sampsyo:
```suggestion
of our heuristics and the out-of-the-box graph isomorphism tooling.
```
Not an independent clause.

content/blog/2019-12-18-dfg-cover/index.md
```diff
@@ -0,0 +1,383 @@
...
+<img src="nettle-sha256-cropped.png" width=100%/>
+
+<img src="nettle-sha256-cropped-zoomed.png" width=100% style='border:2px solid black;'/>
+
+### Stencil generalization
+
+We also explored generating stencils from one benchmark and testing how well they generalized to the other benchmarks.
+The three-node stencils generated and chosen from `minver`:
+1. `fcmp` &rarr; `select` &larr; `fsub`
+2. `getelementptr` &larr; `pointer` &rarr; `getelementptr`)
```
sampsyo: Extra closing parentheses?

sampsyo (changes requested):
This was a pleasure to read. Nice work distilling out a tractable, interesting problem out of a very large possible design space. I also appreciated the way you combined the "breadth" evaluation with several Embench apps with some case-study details.

For some reason, I didn't quite understanding the "learning analogy" bit in the middle. It sounds cool but I couldn't quite connect the parts of the proposed analogy to what you actually did. As I said in a comment within, it may be because I have been reading 6120 reports for too long… but maybe there's something to clarify there.

I'd be glad to talk you y'all about next steps if you're interested after you return from break. (Happy break!)

content/blog/2019-12-18-dfg-cover/index.md
orichardson: We're regularizing the search space by artificially imposing a cost for larger or more sub-graphs. Just like $\ell^1$ regularization, we are explicitly adding a preference for short descriptions, to avoid overfitting to a single program. The example of overfitting is the one given above: if we just set $\mathcal H := \mathcal G$, then we've overfit to the particular program we were using to try to infer general stencils.

content/blog/2019-12-18-dfg-cover/index.md
orichardson: Sorry, I will expand this to make it clearer. The training data is a program DFG. The hypothesis class consists of possible collections $\mathcal H$ of snippets; the evaluation is the coverage of the program. You can test on your training data (see how well your snippets work for the program they were fit to) or on held-out data (generate snippets from one program, and observe their coverage on another). 

The point is that we can either stipulate the most useful snippets by hand, or learn them by example.

sampsyo: Got it; that helps! Thanks!

content/blog/2019-12-18-dfg-cover/index.md
sampsyo: Excellent; this also helps me clarify. An overfit solution would just encode exactly what the program wants to do as one giant instruction. Makes sense!

sampsyo (approved):
Looks great, y'all!

Issue #146, opened 2019-11-21
Project 4 Proposal: Atom Autoscheduler


g-c-c:
**What will you do?**

We will develop a Halide autoscheduling pass for accelerating 4DSTEM analysis. Four-dimensional Scanning Transmission Electron Microscopy (4DSTEM) lets you see atoms better than ever before. Sadly, there's a lot of data which existing image analysis is not built for. A single image is several gigabytes; at this rate a single experiment will amount to petabytes of data.

Optimizing data flow with Halide has the additional complexity of dimensionality. Each STEM "image" is a 2D real-space array of 2D fourier-space arrays. Sometimes these are gathered with extra dimensions of time or rotation. To make experiments and image processing feasible, we will implement a pipeline for running existing algorithms in Halide, with automatic scheduling of operations.

**How will you do it?**

Our first task is to have Halide successfully schedule 4DSTEM algorithms. Halide works great for normal 2D pictures, yet some types of 4DSTEM image analysis must look at the full 4D or 5D dataset. We will start with a toy dataset of the right dimensionalities, then scale to real dat.

Given the many analysis modules, we will initially choose a simple algorithm for Halide to work with. Should scheduling proceed successfully, more complex problems may be solved.

**How will you empirically measure success?**

As our main goal is to attain reasonable data processing speeds, we will measure the speedup. We will test a variety of 4D image analysis algorithms with our custom-constructed Halide scheduler. We will test the speedup gain on a variety of architectures.

**Team members:**
@rolph-recto @hackedy

sampsyo:
Sounds cool! I think the biggest risk is that Halide might not actually be able to express the algorithms you want. So please write those programs first to make sure you can do something.

About "a variety of architectures": I think one hardware target is fine, as long as it's not a single-core CPU. Just compare the speed of a hand-scheduled implementation to the one your automated scheduler generates.

I would also be interested to hear any thoughts y'all have about how you will search for schedules. Randomly? Exhaustively? Heuristically? Solver-aided-ly?

sampsyo:
Closed in #182.

PR #182, opened 2019-12-22
Project 4: An autoscheduler for Halide

rolph-recto:
corresponding to https://github.com/sampsyo/cs6120/issues/146


content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
++++
+title = "An Autoscheduler for Halide"
+[extra]
+bio = """
+  Ela is a fourth-year graduate student. They should fill their bio here!
```
sampsyo: They should indeed, if they want!

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Schedule Transformers
+
+We define transformers over schedule trees.
+We use these to traverse the search space of schedules.
+
+- **Split** - split a function's variable into two. For example, we can split
+  a function's `x` variable into `x_inner` and `x_outer`.
+  Combined with *reorder*, *split* can represent schedules that *tile*
+  computations.
```
sampsyo: It makes sense that something like this would be useful for tiling, but the description is still a bit vague—maybe add one more half-sentence to explain what `x_inner` and `x_outer` would represent in this example?

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+
+Given that `f` is realized in a 512x512 box and a schedule where `g` is
+computed in total before computing `f`, the extent of `g` is 513x513.
+
+Computing the extent in which functions will be computed is hard in general,
+but since Halide makes the simplifying assumption that all extents are
+rectangular (as opposed to, say, polytopes in the polyhedral model),
+there is a simple method for doing this:
+we only need to check the maximum and minimum points of the caller functions
+and check the arguments to the callee.
```
sampsyo: Seems like this also relies on an implicit assumption that the index "propagation" is linear? For example, you might have a function like `f(x) = g(x == 50 ? 1000 : x)` or something, and if you knew the extent for `f` was 0–100, you'd miss the extent of `g` including one far-out point. But maybe this is implicated by the "rectangular" assumption.

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+We measure runtime and memory usage using `gprof`.
+
+For the experiments, we set the weights for execution features as follows:
+
+Group       | Weight 
+------------|---------------
+mem         | 0.1
+loads       | 0.5
+stores      | 0.5
+arith ops   | 1.0
```
sampsyo: Interesting! As long as we're guessing, I would have guessed load & store costs to be _more_ than that for arithmetic ops. I guess it depends heavily on which level of the memory hierarchy the data ends up fitting into…

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+```
+g(x,y) = sqrt(cos(x) + sin(y));
+f(x,y) = g(x + 1,y) + g(x,y) + g(x + 1,y + 1) + g(x,y + 1);
+```
+
+Function     | Runtime (ms) | Peak heap usage (bytes)
+-------------|--------------|------------------------
+f (DEF)      | 87.72        | 0
+f (OPT)      | 13.48        | 0
+g (DEF)      | 0.00         | 0
```
sampsyo: Did this benchmark really run in zero time?

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+```
+blur_x(x,y) = input(x - 1,y) + input(x,y) + input(x + 1,y) / 3;
+blur_y(x,y) = blur_x(x - 1,y) + blur_x(x,y) + blur_x(x + 1,y) / 3;
+```
+
+Function     | Runtime (ms) | Peak heap usage (bytes)
+-------------|--------------|------------------------
+blur_y (DEF) | 12.70        | 0
+blur_y (OPT) | 19.02        | 0
+blur_x (DEF) | 0.00         | 0
```
sampsyo: This one too—perhaps some number is getting dropped?

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+f (OPT)      | 12.18        | 0
+
+
+### Discussion
+
+Note that for the DEF configuration, only the output functions have runtimes
+associated with them since all called functions are inlined.
+
+The autoscheduler performs rather poorly relative to the default schedule.
+While it successfully makes space-runtime tradeoffs (e.g. `f` in Benchmark 1),
```
sampsyo:
```suggestion
While it successfully makes space-runtime tradeoffs (e.g., `f` in Benchmark 1),
```

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+allowing the computation of a function to run much faster by saving
+intermediate results, it runs more slowly and uses more memory than the
+default schedule across all benchmarks.
+
+We believe the poor performance of the autoscheduler has two main causes:
+
+* **Wrong feature weights.** The feature weights for the cost model
+  are chosen by fiat; if these were learned instead given a set of
+  training data, then more the weights can probably better capture the
+  execution profile of schedules.
```
sampsyo: Not that you should do this now, obviously, but you could imagine evaluating how well your current cost model correlates with real running times by just executing lots of the configurations you explored…

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+
+* **Missing execution features**. There are some execution features not
+  captured in the current cost model that probably has a significant effect
+  on performance.
```
sampsyo: Indeed; that Halide auto-scheduler paper you linked to above uses _lots_ of features.

content/blog/2019-12-18-halide-autoscheduler/index.md
```diff
@@ -0,0 +1,359 @@
...
+  Most importantly, the cost model does not reason about locality.
+  Because of this, the autoscheduler sometimes generates schedules with
+  loop order that has poor locality
+  (e.g. a function being traversed in column-major order instead
```
sampsyo:
```suggestion
  (e.g., a function being traversed in column-major order instead
```

sampsyo (changes requested):
Cool thing! It's too bad that the results were not obviously better, but I agree with your diagnosis that cost modeling and featurizing might be the main culprits. Sadly, I think both of these things wouldn't be _too_ hard to explore and perhaps even correct, so it's too bad that the time for the project was too short. It's hard to justify keeping working on this because it seems like it might be difficult to publish, but if you are looking for distractions from holiday festivities this winter, maybe a little more exploration and hacking could make the results Actually Good™. In any case, positive results are not necessary for a successful 6120 project, so nice work.

content/blog/2019-12-18-halide-autoscheduler/index.md
rolph-recto: Inlined functions have 0 runtime; see Discussion section

sampsyo: Aha; I see! It might be worth calling that out explicitly next to the table, or using an "N/A" or something instead of a zero to make this obvious.

content/blog/2019-12-18-halide-autoscheduler/index.md
rolph-recto: This function specifically can't be expressed because we assume arguments are drawn from a grammar of "simple" arithmetic expressions --- `+`, `-`, `*`, `/`, variables and constants. Will add text to make this more explicit.

sampsyo (approved):
Looks good! @g-c-c, there is still a gap for your bio. If you want to fill one in, please open a separate PR.

PR #200, opened 2020-09-02
fix task link

EclecticGriffin:
This just fixes the task link on the lessons page since it was missing the lesson perma-link.


sampsyo:
Good catch; thanks!!

PR #202, opened 2020-09-02
Michael Roberts taking path profiling

mrob94:
This is the last available slot that isn't late in the semester, when I anticipate being super busy.


sampsyo:
You got it! I absolutely love this paper; it is understated but so clever.

PR #211, opened 2020-09-10
Claim a paper for discussion

JASONews:
Talked with Hanchen. We are going to present the paper for safe parallelism together. Thanks!


sampsyo:
Sounds good!

PR #243, opened 2020-12-12
Fix paper link in Lesson 1

tbmreza:
The semanticscholar page directs to http://www.cs.virginia.edu/~techrep/CS-2001-12.pdf, which is apparently 404 now.


sampsyo:
Thanks!!

PR #245, opened 2020-12-19
Add blog post for type and effect system

### LINKED ISSUE #227: Project Proposal: Type and Effect System
**What will you do?**
I hope to implement a simple interpreted language with a type and effect system. To my understanding, this has two potential facets: a static system, where "effects" are a kind, and a runtime augmentation, where an effect is encoded as metadata on a value. The static system will allow for more careful programming (e.g. explicit marking of side effects) as well as enabling smarter optimizations (e.g. marking pure/idempotent function calls). The runtime support would allow for things like algebraic effect handlers, where functions can return to their callers with special values and be resumed.

This project was partially inspired by [1](https://www.janestreet.com/tech-talks/effective-programming/), [2](http://web.cs.ucla.edu/~todd/research/tldi09.pdf), and [3](https://www.ccs.neu.edu/home/amal/course/7480-s12/effects-notes.pdf), with additional inspiration from the question of to how compilers might be able to take advantage of types (and effects).

**How will you do it?**
I plan on implementing the static system first; namely, the language's frontend and typechecker. Then, I want to demonstrate the benefits by implementing simple optimization passes that take advantage of the side effect information. As a later goal, I'd like to also extend this system with user-defined effects and runtime support.

I'm still fleshing out the actual language specification, but I will start simply with expressions over types like `int` and `bool` and control flow statements with the two side-effect statements `output` and `input` to write to `stdout` and read from `stdin`. This should be sufficient to demonstrate some optimizations based on idempotence and reordering. Then, I want to implement support for runtime effects, allowing dynamic detection of what effects occurred -- this would use some sort of pattern matching and a special `return` construct. A stretch goal would be to implement a memory extension with arrays and explore some of the memory-based effects discussed in the above sources. 

As this project hinges on a new language, there won't be existing benchmark suites or statistics to compare against. I plan on writing my own benchmarks for the purposes of testing language correctness and optimization performance, but further suggestions on how to evaluate the project's success are very welcome.

**Team members:**
Just me

kq-li:
closes #227

### DIFF
--- a/content/blog/2020-12-18-type-and-effect-system.md
+++ b/content/blog/2020-12-18-type-and-effect-system.md
@@ -1,0 +1,93 @@
++++
+title = "A Semi-Practical Type-and-Effect System"
+[extra]
+bio = """
+  Kenneth Li is a first-semester M.Eng. student who is interested in functional
+  programming. He sings bass in the 
+  [Cornell Hangovers](http://www.hangovers.com/).
+"""
+[[extra.authors]]
+name = "Kenneth Li"
++++
+
+# Introduction
+Side effects in programming are any observable effects caused by executing some code beyond its return values. For example, the simple `print` functions inherent in most languages cause the side effect of printing to the standard output stream. However, because side effects can be observed from the outside, such as by the program user, then to enforce determinism of programs the order of side-effect-causing instructions must be the same. However, side-effect-free instructions, which we will refer to as "pure", can be rearranged arbitrarily or even precomputed, so long as the code results in the same output. Unfortunately, in most modern programming languages, it is difficult to know what side effects are caused by what code, making it difficult for compilers to reorder instructions. In this report, we explore a way to encode these side effects in a static manner.
+
+In order to perform side-effect-aware static analysis, we introduce the idea of a _type-and-effect system_, which extends the notion of types to encode side effect information. For example, the OCaml function `print_int` has type `int -> unit`, but in a type-and-effect system, the signature might look more like `print_int : (int -> unit !output)`. The following section explores one implementation of such a system, available [here](https://github.com/kq-li/type-and-effect-system).
+
+# Language
+The language designed for this project operates over four types -- `unit`, `int`, `bool`, and one-argument functions `t1 -> t2` -- and four effects -- `input`, `output`, `read`, and `write`. In addition, the language includes control flow statements such as conditionals and loops. Finally, the language supports higher-order and anonymous functions.
+
+The standard library for the language includes, along with standard unary and binary operations, four side-effect-causing functions:
+- `scan : (unit -> int !input)` reads an integer from standard in
+- `print : (int -> unit !output)` writes an integer to standard out
+- `load : (int -> int !read)` reads an integer from the specified integer memory location 
+- `store : int -> (int -> unit !write)` writes an integer to the specified integer memory location
+
+Below is a sample program, which takes in a number as input and computes its square:
+
+```
+x : int !input = scan ();
+y : int = mul x x;
+_ : unit !output = print y;
+```
+
+Notice the special assignment syntax: every expression must be assigned to a variable (with `_` as a dummy) and every effect of an assigned expression must be encoded exactly in the type annotation. These requirements, which are enforced by the typechecker, guard against user error: it is impossible to cause a side effect without knowing about it.
+
+All functions are values, and use the lambda-calculus-inspired syntax `\x : t . s` in which `x` is the argument, `t` is its type, and `s` is a statement to execute. In addition, there is a short form `\x : t . e` where the statement is replaced by an expression to be immediately returned.
+
+Below is a sample program demonstrating functions and their properties:
+
+```
+add5 : int -> int = add 5;
+twice : (int -> int) -> int -> int = \f : int -> int. \x : int. f (f x);
+print_thrice : (int -> unit !output) = \x : int. {
+  _ : unit !output = print x;
+  _ : unit !output = print x;
+  _ : unit !output = print x;
+}
+_ : unit !output = print_thrice (twice add5 2);
+```
+
+# Implementation
+The language is written entirely in OCaml, using Jane Street Core as standard library, ocamllex as lexer-generator, and Menhir as parser-generator. The provided implementation includes a typechecker, an interpreter, and a sample optimization pass. The `Or_error` monad and its corresponding `let%bind`/`let%map` monadic syntax is used throughout the implementation to capture and thread through errors.
+
+## Typechecker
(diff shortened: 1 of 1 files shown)


content/blog/2020-12-18-type-and-effect-system.md
```diff
@@ -0,0 +1,93 @@
...
+
+In order to perform side-effect-aware static analysis, we introduce the idea of a _type-and-effect system_, which extends the notion of types to encode side effect information. For example, the OCaml function `print_int` has type `int -> unit`, but in a type-and-effect system, the signature might look more like `print_int : (int -> unit !output)`. The following section explores one implementation of such a system, available [here](https://github.com/kq-li/type-and-effect-system).
```
sampsyo: FWIW, "we introduce" sounds like you invented the idea. (I know you didn't mean to, but people often say stuff like "we introduce a foo" to mean "this paper is about a new foo we made" in papers.) Linking to/citing some existing work on type-and-effect systems might be cool too!

content/blog/2020-12-18-type-and-effect-system.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+Notice the special assignment syntax: every expression must be assigned to a variable (with `_` as a dummy) and every effect of an assigned expression must be encoded exactly in the type annotation. These requirements, which are enforced by the typechecker, guard against user error: it is impossible to cause a side effect without knowing about it.
+
+All functions are values, and use the lambda-calculus-inspired syntax `\x : t . s` in which `x` is the argument, `t` is its type, and `s` is a statement to execute. In addition, there is a short form `\x : t . e` where the statement is replaced by an expression to be immediately returned.
+
+Below is a sample program demonstrating functions and their properties:
+
+```
+add5 : int -> int = add 5;
```
sampsyo, line 46: Presumably `add` is defined somewhere else?

content/blog/2020-12-18-type-and-effect-system.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Interpreter
+The interpreter takes a parsed program AST as input and executes the program, relying on the typechecker as a preprocessing step to better enforce correctness. Starting from an empty map from variable names to their values, each statement of the list of top-level statements comprising the entire program is executed in order, with the resulting context being threaded from one statement to the next. 
+
+The most difficult part of this module was correctly evaluating lambdas and closing around the variables needed, especially in partial application situations. To that end, `Lambda (x, t, s)` had to be extended in the code to `Lambda (x, t, ctx, s)`, with `ctx` representing the context of the closure upon creation. Then, partial application could be implemented as nested lambda expressions, and evaluating one layer of a lambda stack would encode its argument into the context of the next layer.
```
sampsyo, line 67: Sounds about right. Just to be clear here, though, there wasn't anything particularly effecty about this, right? It's how you would do it if you were implementing an interpreter for a language without effects too?

content/blog/2020-12-18-type-and-effect-system.md
```diff
@@ -0,0 +1,93 @@
...
+
+The standard library provides a mapping from function names to function types, which is used as the starting type context for typechecking. This had the unfortunate constraint of disallowing `print` from supporting a variable number of arguments; varargs support would require an extra wrapper around types in the type context or a new `Any` type, neither of which were appealing options for this project.
+
+## Interpreter
+The interpreter takes a parsed program AST as input and executes the program, relying on the typechecker as a preprocessing step to better enforce correctness. Starting from an empty map from variable names to their values, each statement of the list of top-level statements comprising the entire program is executed in order, with the resulting context being threaded from one statement to the next.
```
sampsyo: I'd be interested to hear a little more about what you mean by "better enforce correctness."

content/blog/2020-12-18-type-and-effect-system.md
```diff
@@ -0,0 +1,93 @@
...
+
+The most difficult part of this module was correctly evaluating lambdas and closing around the variables needed, especially in partial application situations. To that end, `Lambda (x, t, s)` had to be extended in the code to `Lambda (x, t, ctx, s)`, with `ctx` representing the context of the closure upon creation. Then, partial application could be implemented as nested lambda expressions, and evaluating one layer of a lambda stack would encode its argument into the context of the next layer. 
+
+## Optimizer
+The optimizer included with the language is very simple, demonstrating only one of the possible optimizations. In essence, relying on the typechecker to properly enforce the effect annotations, the optimizer is able to collapse pure expressions that have fixed inputs, essentially performing a combination of inlining and constant propagation. This can be seen in the `fact.txt` sample program, which automatically precomputes the factorial of 5 after an optimization pass. A further dead code elimination pass, not implemented here, could drastically reduce program size in this case.
```
sampsyo: Link to that example program?

content/blog/2020-12-18-type-and-effect-system.md
```diff
@@ -0,0 +1,93 @@
...
+
+For each of the small benchmark programs in the `test/basic/` subdirectory, an optimized version will run in the `test/optimize/` subdirectory. The comparison script `test/compare.sh` will diff the results from these two runs; at the time of writing, the results of this diff are as follows:
+- `fact.txt` is improved from 24 executed statements to 4
+- `func.txt` is improved from 8 executed statements to 6
+- `func3.txt` is improved from 20 executed statements to 8
+- all other program execution counts were unchanged
+
+Though not deep, this level of empirical evaluation seemed appropriate considering the scope of the optimization: any memory manipulation, output, or input (e.g. variable input) would result in no optimizations at all.
```
sampsyo: Always use a comma after "e.g." https://capra.cs.cornell.edu/styleguide/#egie
(This also occurs again below.)

sampsyo (changes requested):
Cool! Even if you had to scale back your scope a bit, this is still a solid exploration of how to implement a type checker and some simple optimizations in the presence of effects. Seems useful!

content/blog/2020-12-18-type-and-effect-system.md
kq-li: yep! added a bit in reference to the extensions section that talks about effects at runtime

content/blog/2020-12-18-type-and-effect-system.md
kq-li: yep, clarified in the stdlib section

content/blog/2020-12-18-type-and-effect-system.md
kq-li: makes sense!

content/blog/2020-12-18-type-and-effect-system.md
kq-li: rephrased this -- "correctness" is too loaded a term, haha

sampsyo:
Yo—just a quick reminder to use "request a review" when you're done making changes to remind me to take a look.

sampsyo (changes requested):
Looks great; thanks!

sampsyo (approved):

PR #248, opened 2020-12-19
Final Project: Relational Algebra Compiler

### LINKED ISSUE #225: Project Proposal: Relational Algebra Compiler
**What will you do?**
The way most existing SQL engines work is by first converting SQL to a tree of [relational algebra](https://en.wikipedia.org/wiki/Relational_algebra) operators and then interpreting that tree. Interpreting means that each operator has a produce function. When you call it, the operator will call a produce function from its child operators which return a batch of tuples (or empty if it's done), process that batch (filter by some condition, execute a map function, join data, etc.), and then return the processed batch.

While easy to implement from a system point of view, this leads to a lot of overhead as you have a function call boundary at each operator. Instead, by generating actual code, we minimize the overhead of this and keep values in registers for as long as possible.

For example, this would be what is executed for the interpreted version of `op = Filter(Scan(T), x>5)`
```
output = []
while (true) {
  batch = op.produce()
  if (batch.empty()) break;
  output.append_all(batch)
}

GT.evaluate() {
   l = left.evaluate()
   r = right.evaluate()
   return l > r;
}

Filter.produce() {
   batch = child.produce()
   output = []
   for (x in batch) {
      if (expr.evaluate(x)) {
        output.append(x)
      }
   }
   return output;
}

Scan.produce() {
  return next batch of T
}
```
This is really backward. The following equivalent code will be much faster as it avoids all these virtual calls and keeps values in registers for longer.
```
output = []
for (x in T) {
   if (x > 5) {
     output.append(x)
   }
}
```
Our goal is to for a given relational plan, generate the code for the second version, compile and execute it. 

**How will you do it?**
Broadly, the implementation will follow these papers:
https://www.vldb.org/pvldb/vol4/p539-neumann.pdf
https://db.in.tum.de/~kersten/Tidy%20Tuples%20and%20Flying%20Start%20Fast%20Compilation%20and%20Fast%20Execution%20of%20Relational%20Queries%20in%20Umbra.pdf?lang=en

Rather than a single produce interface to "pull" tuples from the child operators, you use a produce/consume interface. Produce tells the operator to generate code that iterates over it's result while consume allows you to "push" tuples to that operator. This ends up generating tight loops around the pipelines of the program.

While a lot of paper 1 and paper 2 focus on strategies to reduce compilation time (immediately generate LLVM, generate custom IR, restrict LLVM compilation passes, etc.), I'll skip that and directly generate C++ code. My goal is to ideally use the data structures outlined by the second paper: proxy structures that generate code to reduce the complexity of the code generator, wrapper types, linking directly with the database types so I can have a maintainable code base for extending after this course.

I'll implement an in-memory column store as the storage back end.

My goal is to first get this working for just tables with integer types and then progressively add in other types like dates (just end up being ints), strings (ideally use dictionary encoding to just keep them as ints), etc.

I'll limit joins to equi-joins for now and then use cross products for more wild join conditions.

Depending on how many types the system supports, I'll take an existing benchmark suite like TPC-H or JOB and then filter them down to the compatible queries and run them or prune out the features that I don't support and run it.

I'll plan on finding a similar database that is interpreted and execute that.

**Team members:**
Just me.

ankushrayabhari:
closes #225


content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
++++
+title = "Compilation-based Execution Engine for a Database"
```
sampsyo: Probably capitalize "Based."

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+name = "Ankush Rayabhari"
++++
+
+My project was to implement a compilation-based execution engine for a database. The code is located [here](https://github.com/ankushrayabhari/kushdb/).
+
+## Introduction
+
+Existing database systems initially convert a SQL query to a tree of relational algebra operators and then interpret that plan. In the past, this wasn't a bottleneck as the main cost of processing data was reading and writing data to disk. Since the amount of CPU time spent processing data was negligible compared to the disk access time, the overheads of interpretation didn't play a large role.
+
+Nowadays, databases have access to faster storage and large amounts of main memory which means that the CPU time spent processing data is much more valuable. To reduce the overhead of interpretation, we can directly compile trees of operators down to binary and then execute them. This projects explores implementing such a compiler.
```
sampsyo: Might be worth linking to/citing some sort of canonical reference on query compilation? Like, is [this one](https://15721.courses.cs.cmu.edu/spring2018/papers/03-compilation/shaikhha-sigmod2016.pdf) any good?

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+  output = []
+  for (t in T) {
+    if (t[x] > 5) {
+      output.add(t)
+    }
+  }
+  return output
+}
+```
+This should be much faster since there are no function calls (let alone virtual ones) and since tight loops are generated, values are kept in registered for as long as possible.
```
sampsyo: "in registered" -> registered, or in registers

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+## High Level Design
+
+### Producer/Consumer Interface
+
+To generate code for an operator tree, we introduce produce and consume functions for each operator. Produce  generates a loop over the tuples in an operator's result. Consume is called by an operator's child surrounded by a loop over its result tuples. Effectively, produce signifies to "pull" tuples from an operator while consume "push"es them. Expressions have a produce function that generate code for calculating that expression.
```
sampsyo: generate -> generates

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+  // This is called in a loop over the child's result
+  output << "if (";
+    cond.produce()
+  output << ") {"
+    parent.consume()
+  output << "}"
+}
+```
+
+It's worth taking some time to appreciate how elegant this is and check that it does indeed do what we want. I was positively awestruck when I first came across it.
```
sampsyo, line 122: 😄

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+Filter.produce() {
+  // To generate a loop over its result, generate a loop over its child's result.
+  child.produce();
+}
+
+Filter.consume() {
+  // This is called in a loop over the child's result
+  output << "if (";
+    cond.produce()
```
sampsyo, line 115: Just checking, do conditions have `produce` functions too? Or should this be `evaluate`?

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+    parent.consume()
+  output << "}"
+}
+
+GroupBy.consume() {
+  output << "insert t into HT[group by cols of t]"
+}
+```
+
+## Implementation
```
sampsyo, line 168: Maybe a couple of sentences of overview here would help clarify what you've actually built. Did you implement a whole DBMS? Just a query engine? If so, how does the query engine act on actual data? A little context here would be useful.

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+### Operator Tree
+
+This is fairly standard. We introduce a base Operator and Expression class that is inherited by respective implementations. Each one can hold any number of child operators or expressions.
+
+The only thing of note is that operators need to keep track of the schema that it outputs. For example, a table scan would have some or all the underlying table's columns in its schema. A filter operation would have the schema of its child. A join would have the schema of the concatenation of both its inputs. A group by would have the schema of whatever aggregation functions are defined on it. More generally, we associate an array of expressions as the output schema where each expression computes a function of the input schema values.
```
sampsyo: "that it outputs" -> that they output?

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+- The format for all types except for TEXT consists of tightly-packed values of the underlying type, much like an in-memory array. No other metadata such as cardinality is stored.
+- For TEXT types, the format used is as follows. The first 4 bytes represent an unsigned 32-bit integer that contains the cardinality (number of strings) inside the file. Then, for each string, we store 8 bytes. The first 4 bytes are an unsigned 32-bit integer that contains the string length. The second 4 bytes are an unsigned 32-bit integer that contains the byte offset of the string's data in the file. After this metadata array, we store all the strings in a packed format.
+
+To avoid having to implement a buffer pool manager and on-demand paging, I mmap the column files and rely on OS paging to handle this.
+
+### Code Generation
+
+We generate C++ code directly to simplify the implementation. All the types above are native C++ types that we can reuse in the implementation. For example, we can use int16_t, int32_t and int64_t for 16, 32 and 64-bit signed integers respectively. Code is generated within a compute function, `extern "C" compute() {...}`.
```
sampsyo: Maybe use Markdown backticks for those C type names?

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+As mentioned in the design, we use the producer/consumer interface to generate the body of this compute function. Each operator in the tree needs to have a produce/consume function on it. While initially I developed these as plain functions that simply take in an operator as input, I quickly realized that the produce and consume function calls on the same operator needed to share state. For example, the hash join produce function outputs a hash table declaration while the hash join consume function relies on that same variable.
+
+Because state needs to be shared, we could attach these extra state variables onto the relation algebra operators. However, for better separation of concerns, we create a wrapper translator object around each operator and then organize them into the same tree. This allows for per-operator state sharing without having to modify each of the plan classes. We use the visitor pattern to generate these translator wrappers.
```
sampsyo: "relation" -> relational

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+```
+CREATE TABLE lineitem_01 AS (SELECT l_returnflag, l_linestatus, l_quantity, l_extendedprice, l_discount, l_tax, l_shipdate FROM lineitem);
+```
+Then, you can rewrite the query to use this newly created `lineitem_01` table. This way, Postgres and our system read the same amount of data from disk.
+
+I also verify correctness by checking that the query results are identical in Postgres.
+
+### Setup
+
+I'm running this on a Macbook Pro, Late 2013 with an i7-4750 HQ and 8GB of RAM. This is running Ubuntu 20.04 under the Windows Linux Subystem.
```
sampsyo: Macbook -> MacBook
Windows Linux Subystem -> Windows Subsystem for Linux

It might be relevant to report whether you're using WSL or WSL2, since they are pretty different from each other (and pretty different from "real" Linux)…

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+For both systems, I took total running times on the client. THis means it includes time to print the result rows were printed but I redirected stdout to /dev/null. While this does include time to execute query planning for Postgres, Postgres reports that it takes less than a millisecond for these on all queries so I've omitted values beyond the decimal place to account for the loss of accuracy.
```
sampsyo: "THis"

sampsyo: Use Markdown backticks for `/dev/null`?

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+![phase1](compile.jpg)
+
+At such low table sizes, the time spent compiling actually dominates. Note that the time spent compiling is fixed so as the database size grows larger, the yellow portion of the graph will increase and the grey portion will stay constant. This means the ratio of time spent compiling to time spent executing will decrease on larger databases.
+
+However, this is still unacceptable on smaller data sets where Postgres outperforms the compilation version. In future work, I discuss some ways that this can be fixed.
+
+## Personal Experiences
+
+The largest barrier I had to overcome in working on this project was understanding all the pieces of the puzzle. I needed to understand each portion of the system enough to implement it.
```
sampsyo: Maybe expand a tiny bit on exactly what those "portions" are?

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Future Work
+
+While this generates extremely high performant code, the time spent compiling is extremely large. The following strategies are useful in reducing this:
+- Directly generate LLVM. This removes any overhead for file I/O to output the C++ and parsing, type checking and converting that C++ to LLVM.
```
sampsyo: FWIW, the conventional wisdom is that LLVM might not be a great fit for most JITs, at least before the highest tier, because it is built to prioritize optimal code over compilation time. So many JITs use simpler code generators such as [DynASM](https://luajit.org/dynasm.html), [B3](https://webkit.org/docs/b3/), or [Cranelift](https://cranelift.readthedocs.io/en/latest/). The same probably applies to query compilers—after all, it is basically a JIT.

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+- Control which optimization passes run on this. For example, most of the overhead comes from dead copies that get inserted into the code. One GVN/DCE pass can get rid of these. Auto-vectorization would also prove beneficial here.
+
+As always, there are a myriad of SQL features that I have not implemented as part of this project. Adding more features would allow me to test this more thoroughly.
+
+Another aspect to explore is how best to parallelize the generate code. Existing techniques break the code up into the materialization boundaries and then use simple strategies to parallelize it.
```
sampsyo: "generate" -> generated

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+Finally, to simplify the codebase, we can take advantage of operator overloading to define proxy types. For example, if you have `int`s a and b, writing a * b in your source code generates code that multiplies a and b. One more layer of indirection will simplify this. For example, if you have `proxy[int]`s a and b, writing a * b in your source code generates code that generates a program that multiplies a and b. This technique allows you to write code that behaves like the compilation version but appears like the interpreted version.
```
sampsyo: Probably use Markdown code backticks for all the code here, such as `a` and `b` and `a * b`.

sampsyo: FWIW, this technique is approximately [LMS](https://scala-lms.github.io).

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Sources
+
+I used the following sources as references:
+- [NoisePage](noise.page)
```
sampsyo: This link does not go where you think it goes (missing protocol).

content/blog/2020-12-18-db-compiler/index.md
```diff
@@ -0,0 +1,293 @@
...
+- https://www.vldb.org/pvldb/vol4/p539-neumann.pdf
+- https://db.in.tum.de/~kersten/Tidy%20Tuples%20and%20Flying%20Start%20Fast%20Compilation%20and%20Fast%20Execution%20of%20Relational%20Queries%20in%20Umbra.pdf?lang=en
+- https://www.cs.purdue.edu/homes/rompf/papers/tahboub-sigmod18.pdf
```
sampsyo: Rather than just listing URLs, it would be more helpful to link to this stuff in context. Like, what did you get out of learning about NoisePage, and what part did it help you with? Link to it from that section by saying "this idea was inspired by NoisePage" or something.

If there is no more context to be given, and you really do want some miscellaneous context-free bullet points at the bottom of your post, at least use actual citations when you're referring to a paper—i.e., give the title and authors instead of just the URL.

sampsyo (changes requested):
This is very very cool! There's a lot of pedagogical value here in demonstrating the basics of how to bootstrap a minimal query compiler from scratch. I like that it's "disembodied" from a proper full-scale DBMS so you get to focus on the compilation aspect. I hope you put this on social media or whatever as a good first step for people interested in the topic!

content/blog/2020-12-18-db-compiler/index.md
ankushrayabhari: Yep. Mentioned that above.

content/blog/2020-12-18-db-compiler/index.md
ankushrayabhari: I've linked some slides there.

content/blog/2020-12-18-db-compiler/index.md
ankushrayabhari: Gave a little more context here.

content/blog/2020-12-18-db-compiler/index.md
ankushrayabhari: Integrated sources like you mentioned.

sampsyo (approved):
Looks awesome. Nice work.

PR #251, opened 2020-12-20
Final Project : Automatic Static Memory Management

### LINKED ISSUE #232: Project Proposal: Automatic Static Memory Management
**What will you do?**
Implement a version of bril that allows for automatic static memory management.

**How will you do it?**
Based on the Cyclone paper and in class discussion, I would have all pointers be parameterized by a region. A region specifies the scope in which a pointer is valid. At the end of a region, all allocations created in that region are freed and can no longer be accessed. I think regions need a lattice like hierarchy, where some regions are above others in the hierarchy and thus memory allocated higher up in the lattice can have pointers to it stored in memory from regions below it. It might look something like 
```
R1  R2
 \  /
  R3
```
Allocations in R1 can be used in allocations of R3 but not in allocations of R2. You can always move down the lattice.

Any allocations in a region must dominate all successor end blocks of the region. For instance, the following would not be allowed:
```
// There is some region R
if cond {
    // Control flow through this block is optional so it dominates itself and no other blocks.
    // I'm hoping that specifying the region for the pointer as R is unnecessary(inferred).
    x : ptr<int><R> = alloc 5;
}
// Region R will end when there is no more code or at all ret statements in the function
// We don't know if x should be freed or not thus the type checker will reject this program.
```

A region is implicitly created at the start of each function which will dominate all exits for the function.
```
func f(){
    // There is some region R initialized
    x : ptr<int> = alloc 5; // I infer that x belongs in region R as the most recently initialized region
    if cond {
         // As the previous allocation dominates this region, I know to free x
        ret
    } else {
        y : ptr<int> = alloc 5; // I infer that y belongs in region R as the most recently initialized region
        // y also dominates this ret and no rets are left un-dominated so the alloc to y is valid

        // As the previous allocation dominates this region, I know to free x
        // I can also free y
        ret
    }
}
```

I think there is an argument for programmers to be able to define regions explicitly which would require implementing an additional operator for bril. I would like to start without this first and only add it if necessary.

Each function argument that has a pointer type is also given a region. If left implicit, I will infer each region as higher in the lattice than the region created for the function body but not above any other function argument.
```
f(x : ptr<ptr<int>> /*implicit <R1>*/, y: ptr<int> /*implicit <R2>*/) {
    // R3 is initialized implicitly for the function
}
R1  R2
 \  /
  R3
```

You can be explicit about regions in your function arguments to try and do something like store y in x or x in y. 
```
f(x : ptr<ptr<int>><R1>, y: ptr<int><R1>) {
    // R3 is initialized implicitly for the function
    *x = y
}
```
Then when you try to call this function, the pointers passed in will be checked that they have the same region. This might be a little to restrictive, it might be more useful to just enforce that R2 is above R1 in the lattice so that y can always be stored in x but not the other way around. If this is useful, then I would need to add some region parameterization on f so that R1<- R2.

If you want to have an array of memory, it seems like you must be able to have multiple pointers into that block of memory. This is to allow programmers to use `ptradd` to create pointers to be used as arguments for `load` calls. This isn't a problem in the cyclone paper because regions as a whole are freed when their scope ends with all the memory inside of them. However, this seems to require some significant runtime implementation or I would need to statically initialize a bunch of heap memory to start and then replace `alloc` with pointer addition to the correct location in that region heap. 

I was kind of hoping to not have to do this and would instead like to propose the following addition to the type checker. I want to introduce a modified version of borrowing/ownership where each memory location has an owner, the pointer that is initialized from the `alloc` call. This owner can only be defined once so that I don't lose the memory address. At the end of the region, `free` is called on all pointers that are owners. An owner can be used to create more pointers to that block of memory through `id` and `ptradd`. These pointers do not own the memory and are in a sense "borrowing" the address to it. These non-owner pointers can be used like normal bril pointers except that free will never be called on them.

There would be a pass over the bril json that type checks the program containing the above type system extension and translates it to the bril type system with free memory calls appropriately inserted.

I think the above is sufficient to create a restricted subset of bril that is checkable and somewhat useable. Then depending on expressiveness, I would add features like explicit creation of regions and R1<-R2 parameterizations.

I would show how the current benchmark suite can be translated to this flavor of bril, that those programs are memory safe, and that they produce identical results to their manually managed versions.

I would in some way show that common memory issues like memory leaks, use after free and double free are not possible with a correctly typed program in this version/extension of bril. This seems like something I should be able to prove from the type system or have a runtime implementation that can check for these issues.

Performance isn't a focus of this project but bril programs that result from this static memory management pass should not have significant performance penalties. It might be interesting to in some way compare bril programs with this extension against standard bril programs that are garbage collected. That is, how does the time it takes to type check and insert free calls compare against just letting the garbage collector handle it. However, this is not a focus of the proposal.

**Team members:**
@Pat-Lafon

Pat-Lafon:
Closes #232

### DIFF
--- a/content/blog/2020-12-19-asmm/index.md
+++ b/content/blog/2020-12-19-asmm/index.md
@@ -10,18 +10,88 @@ name = "Patrick LaFontaine"
 
 ## Motivation: Compiler-led Garbage Collection
 
-The [memory extension][memex] for the Bril programming language provides manually-managed memory for the allocation of arrays. Manual memory management, when the programmer is in charge of explicitly freeing their memory, provides full control to the programmer at the cost of exposing them to difficult-to-debug memory bugs. The common bugs are leaking memory by forgetting to call ```free```, attempting to ```free``` memory more than once, or attempting to use a memory location after it has been freed. A popular way to handle these issues is to have a run-time system in charge of deciding when to ```free``` memory. These run-time garbage collectors allow programmers to easily work with memory, but often at the cost of worse performance, poor latency with "stop the world" pauses or significant memory usage. One alternative to this is to put the compiler in charge of deciding when to free memory. This handles all of the memory management logic, at compile time, for the programmer, hence "Automatic Static Memory Management". The main downside of this approach is that this static analysis, like all compiler analyses, must be conservative which can limit the expressiveness of the programming language.
+The [memory extension][memex] for the Bril programming language provides manually-managed memory for the allocation of arrays. Manual memory management, when the programmer is in charge of explicitly freeing their memory, provides full control to the programmer at the cost of exposing them to difficult-to-debug memory bugs. The common bugs are leaking memory by forgetting to call `free`, attempting to `free` memory more than once, or attempting to use a memory location after it has been freed. A popular way to handle these issues is to have a runtime system in charge of deciding when to `free` memory. These run-time garbage collectors allow programmers to easily work with memory, but often at the cost of worse performance, poor latency with "stop the world" pauses or significant memory usage. One alternative to this is to put the compiler in charge of deciding when to free memory. This handles all of the memory management logic, at compile time, for the programmer, hence "Automatic Static Memory Management". The main downside of this approach is that this static analysis, like all compiler analyses, must be conservative which can limit the expressiveness of the programming language.
 
-## A Region-Based type extension for Bril
+## A Region-Based Type Extension for Bril
 
-This work, based on the [cyclone] language, seeks to bring region-based memory management to the Bril [memory extension][memex]. Each pointer to memory exists within the static scope of assigned memory called a region. Regions are hierarchical such that they can be nested within each other. Data that is allocated inside of a region can only be used within that region or inner-nested regions. When the scope of a region ends, all of the allocations that belong to that region are freed. In this way, the compiler statically rules out the previously stated, common memory bugs by enforcing when and for how long an allocation can be used for. For Bril, this comes in the form of an extension to the Bril Json representation and a static analysis tool that accepts this extension of Bril and converts it to standard Bril with manual memory management inserted into the program.
+This work, based on the [Cyclone] language, seeks to bring region-based memory management to the Bril [memory extension][memex]. Each pointer to memory exists within the static scope of assigned memory called a region. Regions are hierarchical such that they can be nested within each other. Data that is allocated inside of a region can only be used within that region or inner-nested regions. When the scope of a region ends, all of the allocations that belong to that region are freed. In this way, the compiler statically rules out the previously stated, common memory bugs by enforcing when and for how long an allocation can be used for. For Bril, this comes in the form of an extension to the Bril JSON representation and a static analysis tool that accepts this extension of Bril and converts it to standard Bril with manual memory management inserted into the program.
 
 ## Extension Design
+
+An example program using this extension with annotations is as follows:
+
+```json
+{
+  "functions": [
+    {
+      // A new region field for functions
+      "region" : "T",
+      "instrs": [
+        {
+          "dest": "v",
+          "op": "const",
+          "type": "int",
+          "value": 4
+        },
+        {
+          "args": [
+            "v"
+          ],
+          "dest": "bp",
+          "op": "alloc",
+          "type": {
+            // Each ptr type is now a tuple of the inner type and the region its in
+            "ptr": {"type": "bool", "ownership":"Owner", "region": "T"}
+          }
+        },
+        {
+          "args": [
+            "bp"
+          ],
+          "dest": "bp2",
+          "op": "id",
+          "type": {
+            "ptr": {"type": "bool", "ownership":"Borrower", "region": "T"}
+          }
+        },
+        {
+          "dest": "b",
+          "op": "const",
+          "type": "bool",
+          "value": true
+        },
+        {
+          "args": [
+            "bp2",
+            "b"
+          ],
+          "op": "store"
+        },
+        {
+          "args": [
+            "bp2"
+          ],
+          "dest": "b",
+          "op": "load",
+          "type": "bool"
+        },
+        {
+          "args": [
+            "b"
+          ],
+          "op": "print"
+        }
+      ],
+      "name": "main"
+    }
+  ]
+}
+```
 
 The Type enum in the [bril_rs] library is modified as follows:
 
 ```rust
-// I've removed the parts of this enum used in serializing/deserializing to json.
+// I've removed the parts of this enum used in serializing/deserializing to JSON.
 pub enum Type {
     Int,
     Bool,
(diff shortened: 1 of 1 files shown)


content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
...tension][memex] for the Bril programming language provides manually-managed memory for the allocation of arrays. Manual memory management, when the programmer is in charge of explicitly freeing their memory, provides full control to the programmer at the cost of exposing them to difficult-to-debug memory bugs. The common bugs are leaking memory by forgetting to call ```free```, attempting to ```free``` memory more than once, or attempting to use a memory location after it has been freed. A popular way to handle these issues is to have a run-time system in charge of deciding when to ```free``` memory. These run-time garbage collectors allow programmers to easily work with memory, but often at the cost of worse performance, poor latency with "stop the world" pauses or significant memory usage. One alternative to this is to put the compiler in charge of deciding when to free memory. This handles all of the memory management logic, at compile time, for the programmer, hence "Automatic Static Memory Management". The main downside of this approach is that this static analysis, like all compiler analyses, must be conservative which can limit the expressiveness of the programming language.
```
sampsyo: FWIW, just single backtick suffices to mark inline code in Markdown:

```
when to `free` memory
```

sampsyo: Normally, it's "runtime system" rather than "run-time system." But certainly "run-time garbage collector." https://capra.cs.cornell.edu/styleguide/#runtime

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+## A Region-Based type extension for Bril
```
sampsyo: Probably use Title Case consistently in headings?

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+This work, based on the [cyclone] language, seeks to bring region-based memory management to the Bril [memory extension][memex]. Each pointer to memory exists within the static scope of assigned memory called a region. Regions are hierarchical such that they can be nested within each other. Data that is allocated inside of a region can only be used within that region or inner-nested regions. When the scope of a region ends, all of the allocations that belong to that region are freed. In this way, the compiler statically rules out the previously stated, common memory bugs by enforcing when and for how long an allocation can be used for. For Bril, this comes in the form of an extension to the Bril Json representation and a static analysis tool that accepts this extension of Bril and converts it to standard Bril with manual memory management inserted into the program.
```
sampsyo: Capitalize Cyclone. Also, JSON.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Extension Design
+
+The Type enum in the [bril_rs] library is modified as follows:
```
sampsyo, line 91: I think it would be more helpful to just give the JSON representation instead of the Rust data type. People might not be familiar with the internals of your Rust library design, whereas using the JSON notation will match the Bril docs exactly. (And it will help clarify the separation between the abstract, implementation-independent language design and the Rust implementation.)

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+    PointerRegions {
+        pointer_type : Box<Type>,
+        ownership : Option<Ownership>,
+        region: String,
+    },
+    ///
+}
+```
+
+The new addition is a pointer type which is annotated with the region it exists in and a concept of whether a pointer owns the memory it is pointing to. In Bril, every ```id```, ```load```, ```alloc```, and ```ptr_add``` can create a new pointer into memory. However, not all pointers are created equal. Pointer's
```
sampsyo: Pointers (no apostrophe)

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+created by ```alloc``` are the original pointers into the zeroth index of memory and are treated like the "Owners" of that memory. It is these pointers that need to be freed when the scope of their region ends. Pointers created by ```id```, ```load```, and ```ptr_add``` are merely memory borrowers with views into heap memory. They still need to be statically checked that they aren't being used after the region scope ends but do not need a corresponding ```free``` instruction like memory owners do. What is important about this typing is that both pointer types can exist within a program depending on whether you want the region to be inferred or explicitly stated. Whether a pointer owns its memory is trivial to decide and it is not expected that programmers included this annotation. As will be touched on in the evaluation section, this opt-in use of explicit regions makes converting valid programs to this extension simple.
```
sampsyo: "Whether a pointer owns its memory is trivial to decide" actually seems pretty surprising to me… this always depends on the provenance of the pointer? What if a pointer is passed into a function as an argument or is returned from a function as return value? Does the provenance analysis need to be interprocedural then?

Or, more simply, what if the CFG is complicated enough that there is no single reaching definition of a given pointer, and one comes from `alloc` while another comes from `id`?

Finally, what happens if you store an "owning" pointer to memory and then get it back with a `load`? Does it cease to be an owning pointer?

The standard linear-types way to deal with this is that you maintain an invariant that there is exactly one pointer to a given region at any point in time. So if you `ptr_add` based on `p` to produce `q`, then `p` is destroyed and now `q` owns the region. I can't quite tell if this is what you're after.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Current Limitations on Programs
+
+Outside of the type system additions, there are a few restriction put onto Bril Programs for them to be valid for this extension:
```
sampsyo: a few restrictions

sampsyo: Lower-case "programs."

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+created by ```alloc``` are the original pointers into the zeroth index of memory and are treated like the "Owners" of that memory. It is these pointers that need to be freed when the scope of their region ends. Pointers created by ```id```, ```load```, and ```ptr_add``` are merely memory borrowers with views into heap memory. They still need to be statically checked that they aren't being used after the region scope ends but do not need a corresponding ```free``` instruction like memory owners do. What is important about this typing is that both pointer types can exist within a program depending on whether you want the region to be inferred or explicitly stated. Whether a pointer owns its memory is trivial to decide and it is not expected that programmers included this annotation. As will be touched on in the evaluation section, this opt-in use of explicit regions makes converting valid programs to this extension simple.
+
```
sampsyo, line 193: It's not quite clear yet, by the end of this section, what the type annotations look like. Maybe a very small example program showing a program that produces a memory-related type error would help clarify?

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+## Current Limitations on Programs
+
+Outside of the type system additions, there are a few restriction put onto Bril Programs for them to be valid for this extension:
+
+- Any variable that owns it's memory(returned from ```alloc```) can not be overwritten during the course of the program. This is to ensure that a program doesn't make multiple allocations using the same variable name but whereas the static analysis can only free the variables once.
```
sampsyo: its memory

sampsyo:
```suggestion
- Any variable that owns it's memory (returned from ```alloc```) can not be overwritten during the course of the program. This is to ensure that a program doesn't make multiple allocations using the same variable name but whereas the static analysis can only free the variables once.
```

sampsyo: Is it correct to summarize this as saying that `alloc` instructions must be in SSA, but other stuff need not be? (Makes me wonder why we don't go "full SSA" instead…)

[edit] Reading below, it sounds like the restriction is even tighter than that: `alloc` must be *dynamically* assigned only once? Clarifying this would be helpful.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+- Currently, programmers cannot declare smaller regions within the function scope. Currently, each function has a region and each pointer argument has it's own region. The next step would be to allow the programmer to create their own scoped regions. This was set asider during the project proposal process but would be interesting further work.
```
sampsyo: "its" (I'm gong to stop pointing out it's/its errors now… please audit them throughout.)

sampsyo: "set asider"

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+- Every Owning pointer must dominate the exit block(s) of it's region. In this way, you cannot conditionally allocate or free memory in a Bril Program. A given allocation of memory must exist in the same variable no matter which block it exits the region from. A borrowed pointer can be freely used as long as it is invalidated once the region scope ends.
+- The programmer can specify when two arguments to a function should have the same region but they are unable to declare one region as nested within another. This would allow finer grain control over what input regions are acceptable for a function can instead the coarseness of, "Either they are the same region or they aren't".
+- A pointer returned from a ```Call``` value instruction is assumed to own the memory it is returning. This is to make a clear distinction of who's job it is to free memory so that different functions don't double free the same memory at the end of their regions. This is more of an implementation limitation to make inference easier than a technical limitation.
+
+## Implementation of the analysis
```
sampsyo: Other sections are Title Case, so this one should be too.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+- Any variable that owns it's memory(returned from ```alloc```) can not be overwritten during the course of the program. This is to ensure that a program doesn't make multiple allocations using the same variable name but whereas the static analysis can only free the variables once.
+- Currently, programmers cannot declare smaller regions within the function scope. Currently, each function has a region and each pointer argument has it's own region. The next step would be to allow the programmer to create their own scoped regions. This was set asider during the project proposal process but would be interesting further work.
+- As a consequence of the first two, a programmer cannot currently allocate memory in a loop. The ideal way to do this would to have the loop body be one region so that at the end of each loop iteration any memory allocated during that iteration is freed. I attempted to infer small regions just for this use case and found it to be both difficult infer and confusing to use. It can be very unclear where the analysis has chosen to start and end a region which makes for a pour experience when the programmer can't tell when a memory location goes out of scope.
```
sampsyo: "pour experience"

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+- Every Owning pointer must dominate the exit block(s) of it's region. In this way, you cannot conditionally allocate or free memory in a Bril Program. A given allocation of memory must exist in the same variable no matter which block it exits the region from. A borrowed pointer can be freely used as long as it is invalidated once the region scope ends.
```
sampsyo: By "the pointer must dominate," do you mean "the `alloc` instruction must dominate"? (Dominance is defined in terms of code locations, not values.)

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+- The programmer can specify when two arguments to a function should have the same region but they are unable to declare one region as nested within another. This would allow finer grain control over what input regions are acceptable for a function can instead the coarseness of, "Either they are the same region or they aren't".
```
sampsyo, line 220: Showing how this attribute of function arguments is expressed in the language would be helpful.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+- A pointer returned from a ```Call``` value instruction is assumed to own the memory it is returning. This is to make a clear distinction of who's job it is to free memory so that different functions don't double free the same memory at the end of their regions. This is more of an implementation limitation to make inference easier than a technical limitation.
```
sampsyo: who's -> whose

sampsyo: I dunno, I think this is not just an implementation limitation… it is pretty fundamental to shift from an intraprocedural analysis to an interprocedural one.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Implementation of the analysis
+
+The analysis is broken up into 5 stages. First the Bril program is broken up into a control flow graph of basic blocks. This a straight forward phase which involves inserting extra labels which will show up in the resulting code. An inference pass is then run on each function to add in any region annotations that may not have been included. The function is given a region name if the programmer does not specify one. This is also when the program identifies the owner pointers from the borrowed pointers. From this point, regardless of user input, the control flow graph for each function is now fully annotated and consistent. Because this inference needs to be conservative, each argument that a program takes in is assumed to be a part of its own region. This is to avoid some kind of leakage bug where a pointer of short lived memory is stored in a long living array and it ends up leaving a gap after it is free.
```
sampsyo: The middle of this process could use more detail:

> An inference pass is then run on each function to add in any region annotations that may not have been included. [...] This is also when the program identifies the owner pointers from the borrowed pointers.

This is kind of the "meat" of what you have done in this project, so I think it deserves some more expansion. The other steps in this chain, such as constructing a CFG, are generic and therefore not terribly germane. But we want to know what these analyses actually do! Is it a data flow analysis? If so, what are the rules? If not, how did you traverse the CFG? Are there any pitfalls that could trip up the analysis if you weren't careful?

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+Once all of the pointer types are fully annotated, the program can be quickly checked for some of the current limitations. For instance, when you call ```alloc```, we can check that all of the exit blocks that can be reached from this block are domination by the current block. This is to ensure when we come through with the next phase, any call to ```alloc``` is valid and needs to be free at each of the exit blocks dominated by this. We go through one more time and remove all of the region annotations to bring the Bril extension back to standard Bril code.
```
sampsyo: "are domination by the current block"

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Evaluation
+
+Surprisingly, the inference part of this tool is strong enough that all four of the current benchmarks that use the memory extension can be converted to this extension of Bril merely by removing the ```free``` calls. The inference pass is able to annotate the rest in these case, ```eigth-queens.bril, fib.bril. mat-mul.bril, and sieve.bril```. The other benchmarks in the current suite do not use the memory extension but they can still be handed to and returned from this pass with no change. Three out of four of these benchmarks have the exact same total dynamic instruction count between the current benchmark and running this bril extension benchmark through the analysis. ```Sieve.bril``` was slower but this was due to the reconstruction of the program from the control graph being suboptimal and is not a performance issue with the extension itself.
```
sampsyo: Don't put the code backticks around English:

```
`eigth-queens.bril`, `fib.bril`, `mat-mul.bril`, and `sieve.bril`
```

Also, the first benchmark is misspelled. Also, don't capitalize filenames at the beginning of sentences.

sampsyo: Instead of just saying "slower," can you give the actual instruction counts?

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+With the exception of the ```mat-mul.bril``` benchmark, these benchmarks are not a very difficult evaluation. It may just be that full flexibility in memory management is hard to keep track of anyways so users are already restricting their use and these benchmarks are representative of some standard use. To compensate for this, I've gone and included the interpreter memory test cases to get fuller coverage of some of the edge cases that could occur. For instance, ```alloc_many.bril``` is a go example of a program that would fail to check in this bril extension.
```
sampsyo: "a go example"

sampsyo: Capitalize Bril.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+Alloc_many.bril
```
sampsyo: Probably don't need this, since you just told us what file this is.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+  br loop .end .lbl;
+.end:
+  print count;
+}
+
+```
+
+Notice how this program loops over the loop body some `max` number of times. Each time, the program allocates and frees some memory. Because the same variable destination is being used each time, this would not check. The allocation would need to be moved to the loop header and either freed immediately or at block `.end`. If there were user defined regions that you could write something like:
+
+Alloc_many.bril.asmm
```
sampsyo: Probably delete this too.

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+```
+@main {
+  v: int = const 1;
+  max: int = const 1000000;
+  count: int = const 0;
+.lbl:
+  start_region "loop_body"
```
sampsyo, line 281: Some discussion of how this might work would be helpful… would these need to be lexically scoped, somehow?

content/blog/2020-12-19-asmm/index.md
```diff
@@ -0,0 +1,119 @@
...
... in two parts. First by using all of the memory tests/benchmarks, this bril extension will have as much coverage/correctness as the interpreter itself. The more logical argument some from the limitations that have been enforced that remove complex control flow like branches and loops for ```alloc``` calls. This means that you can look at the path of the dominator tree from the root to exit leaf nodes to find all of the owner pointers created for this region. Also by not including user created regions, this simplifies the region hierarchy with the current function at the bottom and each of it's arguments has a region different from the function one that is above it in the hierarchy. Another important part around function calls is that function arguments are always of a borrowed pointer type and the return pointer is always an owned pointer type. Therefore, you cannot duplicate a function pointer on accident and try ot free it twice. This gives a clear divide, it is the callers job to do the memory management for any pointer it gives or receives from the callee. It is a bit unclear if a programmer could give purposely wrong annotations to violate the soundness of this transformation.
```
sampsyo: Capitalize Bril.

content/blog/2020-12-19-asmm/index.md
sampsyo: No need to do this now, but I'd be interested to know how much "later" this pass inserts `free`s than the manually managed versions. Moving `free`s later means a larger maximum memory footprint.

content/blog/2020-12-19-asmm/index.md
sampsyo: "try ot"

sampsyo: "the callers job"

sampsyo: > It is a bit unclear if a programmer could give purposely wrong annotations to violate the soundness of this transformation.

That's pretty worrisome! :smiley: Some clarity on what you mean by this would be helpful…

In particular, does your compiler actually enforce that returned values are owned pointers? And presumably it doesn't free them?

sampsyo (commented):
Interesting project! The type system design with a latent distinction between "owned" and "borrowed" pointers causes some interesting problems. I'd be interested in more details on how the analyses work.

sampsyo (changes requested):
Oops; forgot to mark as "request changes."

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: I'll add a sentence about passing a pointer into a function. I treat this as an implicit `id` call so a function can only take a pointer that borrows memory. A pointer returned from a function is always owned specifically to avoid having to do an interprocedural analysis. This seemed like a reasonable limitation/tradeoff.

I think this is where some of my constraints come into play. Having an allocated pointer dominate all of its exit's and not allowing a reassignment to that variable should avoid this.

The last two are similar. I'm not destroying any pointers. Any time you are creating a new pointer from an old one, the new one will be a borrow. In this way, you can't store an "owning" pointer into memory because we are not destroying the original pointer. I think if I allowed that I would need some kind of runtime mechanism/destructor every time I want to free memory that can contain pointers. I don't want to "destroy" pointers.

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: I think dynamically assigned once is slightly less restrictive than SSA so I'll add an example showing clarifying this.

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: Yes

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: Returned values are freed by the caller, not the callee and must be owned so that part is enforced. I will provide an example that I'm still unsure with how to handle that motivates this worry. In general though, I've tried to design the extension to be straightforward to infer. I would much rather a programmer just rely on the inference step to handle annotations rather than writing them directly.

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: I would probably need more memory-extension programs to look at to determine if there is an impact. The current benchmark suite puts all of their free calls at the end so there is no difference in the memory footprint.

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: `alloc` dominates all of their exit blocks, `start_region` would have to dominate all of the `end_region` instructions. I will add a small blurb.

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: Well, an implementation limitation in the sense that I can just add another pass after the intraprocedural one that all it does is check for this case. I didn't consider this to be that difficult, though I decided to avoid this to have a clear story of who frees memory and to hopefully have a stronger case for correctness. Being able to return a pointer that "borrows" memory instead of owning it seems to introduce a lot of possible bugs I need to check for.

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: Makes sense.

content/blog/2020-12-19-asmm/index.md
Pat-Lafon: Sure

sampsyo (approved):
Looks good; thanks!

PR #252, opened 2020-12-20
wds68 final project: algebraic types in bril

### LINKED ISSUE #229: Project Proposal: Add algebraic types to Bril
### What will you do?

I will add algebraic data types, specifically n-tuples and variants (tagged unions), to the Bril specification/brili. I will also implement a type-checker for these types, a simple garbage collector, and a translator from a subset of Rust syntax to Bril.

### How will you do it?

The tuples will be structurally typed and the variants will be nominally typed, so there will need to be a section of the bril file for variant type declarations. For example, types could look like `(int, bool, ptr<int>)` or `type list = nil of () | cons of (int, ptr<list>)`.

I will add instructions for constructing tuples, accessing the nth element of a tuple, constructing a variant, and accessing the value of a variant.

I will implement a type checker pass. This is necessary because brili does not check types and I want to be able to verify that bril programs are correct with respect to types given the extra complexity I will be introducing.

Both tuples and variants will be stack-allocated. I expect, however, that variants will be most useful in combination with pointers from the memory extension in order to avoid copying lots of data. In addition, I suspect that it will be difficult to manually free memory while using things like linked lists idiomatically, so to this end I plan to also implement a simple garbage collector. (I may be able to use the garbage collector from lesson 10.)

Typescript supports tuples but not variants (only Java-style enums), so I don't think extending the ts2bril translator to support algebraic data types will be particularly helpful. Instead, I am considering creating a translator for a subset of Rust syntax (rs2bril?) because Rust supports tuples and enums (which work like variants). This translator will help a lot with testing.

### How will you empirically measure success?

I will measure the success of the brili extensions, the type checker, and the Rust translator by running sample programs that utilize the full range of added expressivity, such as programs operating on linked lists and trees, and verifying the correctness of their behavior in brili.

I will measure the success of the garbage collector by checking that memory usage is bounded in the test programs described above

### Team members:

Me! (Will, wds68)

### Notes:

I am aware of some similar past/current projects for [adding record types](https://github.com/sampsyo/cs6120/issues/25), [adding structs and garbage collection](https://github.com/sampsyo/cs6120/issues/224), and [type checking](https://github.com/sampsyo/cs6120/issues/3). While I may take inspiration from these projects, I think that the proposed work discussed here is different in a number of ways such that it is not redundant: the first project added only record types, not variants; the structs/garbage collection project plans to compile Bril to LLVM rather than staying in Bril; and the type checking project did not incorporate type checking for algebraic data types.

Calsign:
closes #229

### DIFF
--- a/content/blog/2020-12-20-bril-algebraic-data-types/index.md
+++ b/content/blog/2020-12-20-bril-algebraic-data-types/index.md
@@ -1,5 +1,5 @@
 +++
-title = """Bril extension for algebraic data types"""
+title = """Bril Extension for Algebraic Data Types"""
 [extra]
 bio = """
 Will is a senior undergraduate student interested in programming language implementation. He also enjoys autonomous robotics and music peformance."""
@@ -24,7 +24,7 @@ completed each of the following:
    [`bril-rs/src/main.rs`](https://github.com/Calsign/bril/blob/algebraic-types/bril-rs/src/main.rs)
 
 Additionally, I added support for the new instructions to the
-TypeScript, Ocaml, and Rust libraries as required to do the above.
+TypeScript, OCaml, and Rust libraries as required to do the above.
 
 The code is available on the [`algebraic-types`
 branch](https://github.com/Calsign/bril/tree/algebraic-types) of my
@@ -40,8 +40,8 @@ other types (a Cartesian product). Commonly, the two case is called
 a pair and the zero case is called unit.
 
 A sum type is a type that at any given time may contain any of one or
-more types, where which type a given value has may be inspected at
-runtime (a tagged union). Each type in a sum type is referred to as a
+more types, where which type a given value has may be inspected at run
+time (a tagged union). Each type in a sum type is referred to as a
 constructor. Note that the zero case is uninstantiable, so we do not
 allow it.
 
@@ -84,10 +84,10 @@ Here are some examples of types that are supported by my extension:
  - `sum<product<int, int>, bool>`
  - `product<sum<int, bool>, int>`
 
-Note that while sum types with duplicate constructors (e.g. `sum<int,
+Note that while sum types with duplicate constructors (e.g., `sum<int,
 int>`) can be declared with the above type definitions, as we will
 see, it is not possible to differentiate between different
-constructors of the same type at runtime, so such sum types are not
+constructors of the same type at run time, so such sum types are not
 useful.
 
 I considered adding a subtyping rule for sum types that would permit
@@ -178,7 +178,7 @@ Here is a sample Bril program that uses tuples:
 The `construct` instruction is used to construct sum types. It takes
 as its sole argument the value to use. The semantics of the
 instruction dictate that constructor to use is determined
-automatically by examining the runtime type of the argument.
+automatically by examining the run-time type of the argument.
 
 The destination must have a sum type and the type of the argument must
 correspond to one of the constructors of that sum type.
@@ -242,9 +242,9 @@ d: sum<t1, ..., tn> = destruct v .l1 ... .ln
 ```
 
 A common use case is some way of representing an empty, or unit,
-value. For this it is recommended to use the empty product type,
-i.e. `product<>`. For example, to represent an optional `int`, one
-could use the following type: `sum<int, product<>>`.
+value. For this it is recommended to use the empty product type, i.e.,
+`product<>`. For example, to represent an optional `int`, one could
+use the following type: `sum<int, product<>>`.
 
 I debated whether to even include a powerful instruction like
 `destruct` at all. I could have instead offered an instruction to get
@@ -528,7 +528,7 @@ more confident in the correctness of the Rust compiler.
 I do think that it would have been preferable to make the `construct`
 instruction require explicitly specifying the index of the constructor
 to construct. This approach would have been better for a number of
-reasons: it would have eliminated the dependency on runtime knowledge
+reasons: it would have eliminated the dependency on run-time knowledge
 of types; it would have made the Rust compiler's job easier; it would
 have made the code more explicit and readable; and it would have made
 it possible to use variants with multiple constructors that carry the
--- a/content/blog/2020-12-20-bril-algebraic-data-types/index.md
+++ b/content/blog/2020-12-20-bril-algebraic-data-types/index.md
@@ -1,0 +1,538 @@
++++
+title = """Bril extension for algebraic data types"""
+[extra]
+bio = """
+Will is a senior undergraduate student interested in programming language implementation. He also enjoys autonomous robotics and music peformance."""
+[[extra.authors]]
+name = "Will Smith"
+link = "https://github.com/Calsign/"
++++
+
+The result of my final project was the addition of support for product
+types (i.e., tuples) and sum types (i.e., variants) to Bril. I
+completed each of the following:
+
+ - Defined four new instructions: `pack`, `unpack`, `construct`, and
+   `destruct`
+ - Added support for these instructions to `brili`
+ - Added support for these instructions to `bril2txt`/`bril2json`
+ - Created a type checker with support for these new instructions;
+   written in OCaml, located in
+   [`algebraic-types/type-check`](https://github.com/Calsign/bril/tree/algebraic-types/algebraic-types/type-check)
+ - Created a Rust compiler for turning a subset of Rust code into
+   Bril; written in Rust, located in
+   [`bril-rs/src/main.rs`](https://github.com/Calsign/bril/blob/algebraic-types/bril-rs/src/main.rs)
+
+Additionally, I added support for the new instructions to the
+TypeScript, Ocaml, and Rust libraries as required to do the above.
+
+The code is available on the [`algebraic-types`
+branch](https://github.com/Calsign/bril/tree/algebraic-types) of my
+fork of the Bril repo.
+
+The remainder of this blog post is devoted to documentation, design
+choices, and evaluation.
+
+## Design: algebraic types
+
+A product type is a type that acts as a block containing zero or more
+other types (a Cartesian product). Commonly, the two case is called
+a pair and the zero case is called unit.
+
+A sum type is a type that at any given time may contain any of one or
+more types, where which type a given value has may be inspected at
+runtime (a tagged union). Each type in a sum type is referred to as a
+constructor. Note that the zero case is uninstantiable, so we do not
+allow it.
+
+Product and sum types are represented in Bril as new kinds of
+parameterized types. In JSON, they are as follows:
+
+ - Product: `{"product": [t1, ..., tn]}`
+ - Sum: `{"sum": [t1, ..., tn]}`
+
+In text, they are as follows:
+
+ - Product: `product<t1, ..., tn>`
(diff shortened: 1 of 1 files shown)


content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
++++
+title = """Bril extension for algebraic data types"""
```
sampsyo: Please use Title Case for blog posts, for uniformity with other posts.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+ - Added support for these instructions to `bril2txt`/`bril2json`
+ - Created a type checker with support for these new instructions;
+   written in OCaml, located in
+   [`algebraic-types/type-check`](https://github.com/Calsign/bril/tree/algebraic-types/algebraic-types/type-check)
+ - Created a Rust compiler for turning a subset of Rust code into
+   Bril; written in Rust, located in
+   [`bril-rs/src/main.rs`](https://github.com/Calsign/bril/blob/algebraic-types/bril-rs/src/main.rs)
+
+Additionally, I added support for the new instructions to the
+TypeScript, Ocaml, and Rust libraries as required to do the above.
```
sampsyo:
```suggestion
TypeScript, OCaml, and Rust libraries as required to do the above.
```

sampsyo: Dang, that's a lot of implementation! 👏

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+
+```ocaml
+type intlist =
+  | Nil
+  | Cons of int * intlist
+```
+
+Unfortunately, it is not clear to me how to easily support recursive
+types without named sum types, let alone potential other complications
+with recursive types that I did not consider.
```
sampsyo, line 77: It's cool; I still think doing sums & products independently was plenty of work. :smiley: A nice way to add recursion without names would be to add explicit mu-types, even if that's a little silly.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+
+Here are some examples of types that are supported by my extension:
+
+ - `product<>`
+ - `product<int, bool>`
+ - `sum<int, product<>>`
+ - `sum<product<int, int>, bool>`
+ - `product<sum<int, bool>, int>`
+
+Note that while sum types with duplicate constructors (e.g. `sum<int,
```
sampsyo: Always use a comma after "e.g." https://capra.cs.cornell.edu/styleguide/#egie

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+int>`) can be declared with the above type definitions, as we will
+see, it is not possible to differentiate between different
+constructors of the same type at runtime, so such sum types are not
+useful.
+
+I considered adding a subtyping rule for sum types that would permit
+using, for example, `sum<int, bool>` in places that require `sum<int,
+bool, product<>>`. I decided to leave this out because it seemed
+excessive for an IR and difficult to implement.
```
sampsyo, line 96: I agree with your decision here.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+```
+
+### Sum types
+
+#### `construct`
+
+The `construct` instruction is used to construct sum types. It takes
+as its sole argument the value to use. The semantics of the
+instruction dictate that constructor to use is determined
+automatically by examining the runtime type of the argument.
```
sampsyo: "run-time type," with a hyphen, probably. https://capra.cs.cornell.edu/styleguide/#runtime

sampsyo: I'm just a touch concerned about the semantics depending on the run-time type. It would be equivalent, I hope, for the "slot" to be determined by the static type… and of course, it's an error for the run-time type to deviate from that.

Also, while I understand why you did it this way, it makes me wonder about the trade-offs with an explicit version that would require nastier syntax changes. That is, a version with an explicit (static!) index argument. That would let you use stuff like `sum<int, int>` productively. But there is no place in the current Bril syntax for having concrete numbers as arguments to functions (only variables, which wouldn't do, of course).

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+type. The semantics of the instruction dictate that the control flow
+jumps to the correct branch based on the argument and that the
+destination value gets assigned the value within the constructor.
+
+The destination will in fact not have the same sum type as the
+argument; it will have the type carried by one of the constructors,
+which is different for each label. This is confusing because the
+destination is annotated with a type that it cannot have, but I chose
+this approach because the Bril interpreter needs to know the available
+constructors in order to select the correct label to jump to.
```
sampsyo, line 224: Yeah, this is pretty weird! I wonder what a way around this would have been. One option would be to do the "unsafe projection" thing. So you'd say "get me the `int` from this `sum<int, float>`" and it would give you an int or just crash. And there would be a separate instruction to say "which variant are you?" that would presumably return an integer.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+
+The text format of the `destruct` instruction is as follows:
+
+```
+d: sum<t1, ..., tn> = destruct v .l1 ... .ln
+```
+
+A common use case is some way of representing an empty, or unit,
+value. For this it is recommended to use the empty product type,
+i.e. `product<>`. For example, to represent an optional `int`, one
```
sampsyo: i.e.,

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+could use the following type: `sum<int, product<>>`.
+
+I debated whether to even include a powerful instruction like
+`destruct` at all. I could have instead offered an instruction to get
+the index of the constructor used for a value and another instruction
+for casting a sum type down to one of the constructor-carried
+types, using a series of branch instructions to replicate the behavior
+of `destruct`.
```
sampsyo, line 254: Aha, you had the same idea as me. :smiley: Leaving my comment in there above for posterity.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+
+The reason that I opted for the `destruct` approach is that it makes
+it possible, and not terribly difficult, to type-check the usage of
+sum types in a Bril program. With the alternative approach, it is not
+clear to me that it is even possible to perform type-checking of
+things like the casting instruction described above. `destruct` does
+seem rather large and clunky for an IR, especially compared to the
+other instructions (it's like a whole bunch of `br` instructions in
+one!), but I think the ability to perform type-checking makes it
+worthwhile.
```
sampsyo, line 264: Well, you could type-check it, but you couldn't guarantee that the program did bad projections. Which is maybe appropriate for an IR but not for a higher-level language? Interesting design question.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+
+Let bindings must have type annotations attached, for example:
+
+```rust
+let x: i32 = 42;
+```
+
+This restriction comes from the fact that all Bril value instructions
+require a type. It would be possible to add type inference to the
+compiler, but that would have been a lot more work.
```
sampsyo, line 415: Yeah. I guess what you'd actually want is to hook into the actual Rust compiler's AST and reuse its type checker.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+The compiler also does not support arbitrary expressions in `println!`
+macros, but that is because `syn` does not fully parse macro
+contents. Also note that `println!` macros do not use format strings;
+they aren't actually Rust `println!` macros, behind the scenes they
+just transform something like `println!(a, b, b)` into `print a b c`
+in Bril.
+
+### Sample
+
+Here is a sample Rust program (`safe_division.rs`) that can be compiled to Bril:
```
sampsyo, line 434: Pretty amazing that this works.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
```diff
@@ -0,0 +1,538 @@
...
+
+### Reflection
+
+I think that using a powerful `match` instruction was the right
+decision. Having the ability to perform type-checking made me a lot
+more confident in the correctness of the Rust compiler.
+
+I do think that it would have been preferable to make the `construct`
+instruction require explicitly specifying the index of the constructor
+to construct. This approach would have been better for a number of
```
sampsyo, line 530: Yeah, probably so.

sampsyo (changes requested):
Super cool; this is a heroic amount of implemented things. And a very clear writeup! I really liked the way you clearly presented the language extensions with examples, and the discussions of language design trade-offs was super interesting.

If you're interested, maybe we can consider merging at least some of this stuff in to the mainline Bril repository. In particular, product types seem totally good to go, but we might want to reconsider a thing or two about sums. But at least the OCaml type checker could be super nice for other purposes too.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
Calsign: This is exactly what I proposed at the bottom. I am already using a static integer as an argument for the `unpack` instruction. Is this OK? Alternatively, I could add an optional `index` field to value instructions.

content/blog/2020-12-20-bril-algebraic-data-types/index.md
Calsign: So if I revise the `construct` instruction to work this way, would that be good to go for a merge?

Also, is there any interest in the Rust compiler in the main repo? (I know @evanbenadler's proposed a more comprehensive implementation - perhaps that one would be better?)

sampsyo: I think so. I'm still a little puzzled by what to do about the corresponding destructor too… a "safe" version is nice indeed, but the fact that the instruction lies about its type is troubling. I think it might make sense to merge it anyway and think about a different representation later.

sampsyo: About the Rust-to-Bril compiler: why not?! Seems like fun.

sampsyo (approved):
Looks good!

PR #258, opened 2021-04-17
Fix typo and format comments & codes

Marcythm:


sampsyo:
Looking good; thanks!

PR #259, opened 2021-04-26
Updated scviol link

jakevossen5:
Old link was dead ( looks like the repo changed), I found the the file in the new place and updated it. I made it a permanent link, which means that it will point to that version of the file forever but wont break, let me know if you want me to update it to a plain old link.


sampsyo:
Awesome; this is a very good idea. Thanks!!

PR #267, opened 2022-01-25
Update content.toml

gsvic:
Paper claim


sampsyo:
You got it; thanks for claiming, @gsvic!

PR #269, opened 2022-01-25
adding name to schedule

ayakayorihiro:


sampsyo:
Awesome; you got it, @ayakayorihiro!

PR #270, opened 2022-01-25
Paper presentation for Nikita is added

barabanshek:
I would like to present the paper "A Type and Effect System for Deterministic Parallel Java" as I'm familiar with the Java compiler/runtime and I'm generally interested in the very challenging world of concurrency and memory consistency models


sampsyo:
Sounds great; you got it!

PR #276, opened 2022-01-25
Update content.toml

anshumanmohan:


sampsyo:
Great! It’s yours, @anshumanmohan. This paper is such a delight to read.

PR #278, opened 2022-01-25
add charles paper

charles-rs:
title


sampsyo:
Sounds great! This paper should be fun.

PR #281, opened 2022-01-28
Proposed change for presentation date

ayakayorihiro:
I just realized that a paper deadline that I may be aiming for is on March 17 (previously thought it was the week before), so as much as I would have loved to lead the discussion on alias analysis, I think it's best for me to reschedule. Sorry @andrewb1999 for the false alarm! And @5hubh4m I hope you're cool with me joining you :slightly_smiling_face:

### DIFF
--- a/data/content.toml
+++ b/data/content.toml
@@ -62,7 +62,7 @@ lesson = 10
 
 [[classes]]
 title = "Alias-Based Optimization"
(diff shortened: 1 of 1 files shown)


sampsyo:
Sounds good to me, assuming it's OK with @5hubh4m!

PR #291, opened 2022-02-04
Move bril links from master to main

andrewb1999:
Prevents the github popup saying bril's default branch moved from master to main.


sampsyo:
Awesome; thank you very much!

PR #300, opened 2022-03-02
[Blog] Add Efficient Path Profiling Blog

zzzDavid:
This PR is to add the blog of discussion on the paper _Efficient Path Profiling_.


content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+[[extra.authors]]
+name = "Niansong Zhang"
+link = "https://www.zzzdavid.tech"
+[[extra.authors]]
+name = "Jiajie Li"
+link = "https://tonyjie.github.io/"
++++
+
+<p align="center">
+<img src="https://res.cloudinary.com/dxzx2bxch/image/upload/v1646243822/cs6120/2_ugrqoi.png" alt="alt_text" title="image_tooltip" style="zoom:25%;" />
```
sampsyo: It's possible to include your image right here instead of using a third-party hosting service. Just make a directory instead of a single `.md` file, like this: https://github.com/sampsyo/cs6120/tree/2020fa/content/blog/2020-11-09-modern-gc

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+</p>
+
+
+## INTRODUCTION
```
sampsyo: I think the `<h1>` is plenty of emphasis: just use title case, like "Introduction" instead of "INTRODUCTION."

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+
+<p align="center">
+<strong> Left: A Control-Flow Graph (CFG) labeled with edge frequency. Right: Two possible path profiling results with the same edge frequency as the left figure.  </strong>
+</p>
+
+For example, the two path profiling results in the table correspond to the same edge frequency on the left. However, they identify two different most executed paths. Therefore, using edge frequency to estimate path frequency is unreliable. It is easy to find out that path profiling results could determine edge/block profiling results, while the opposite direction is not. 
+
+Also, given an edge profiling result, a commonly used heuristic to select a heavily executed path follows the most frequently executed edge out of a basic block, which identifies path ACDEF in the above example. However, this path is not the most frequently visited in both Prof1 and Prof2, showing the inaccuracy of edge/block profiling. 
+
+The naive way to count how many times each path executes is to exhaust every path in the CFG and label each one. However, this quickly becomes intractable as the CFG becomes large, or the CFG contains cycles with infinite paths.
```
sampsyo:
```suggestion
The naive way to count how many times each path executes is to exhaust every path in the CFG and label each one. However, this quickly becomes intractable as the CFG becomes large, or the CFG contains cycles and therefore infinite paths. 
```

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+
+Also, given an edge profiling result, a commonly used heuristic to select a heavily executed path follows the most frequently executed edge out of a basic block, which identifies path ACDEF in the above example. However, this path is not the most frequently visited in both Prof1 and Prof2, showing the inaccuracy of edge/block profiling. 
+
+The naive way to count how many times each path executes is to exhaust every path in the CFG and label each one. However, this quickly becomes intractable as the CFG becomes large, or the CFG contains cycles with infinite paths. 
+
+In short, the motivation is to use an efficient path profiling method instead of traditional block/edge profiling to improve the accuracy without increasing too much effort. To have a preview, the overhead of path profiling is 30.9%, while the overhead of edge profiling is 16.1%. (You can decide whether this overhead is acceptable for you!)
```
sampsyo: > You can decide whether this overhead is acceptable for you!

Good point! 😊

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+The intuition of the path encoding algorithm is to count the number of paths along “the other” edges and increment by the number of paths. 
+
+<p align="center">
+<img src="https://res.cloudinary.com/dxzx2bxch/image/upload/v1646243822/cs6120/3_dmiy5c.png">
+<p>
+<p align="center">
+<strong> The path encoding generation algorithm </strong>
+</p>
+
+Therefore, we count the number of paths from leaf to node. In other words, we visit vertices in the reverse topological order.
```
sampsyo: By "from leaf to node," do you mean "from the node to any leaf"? Or do you mean "from the root to the node"? In general, in a directed tree, there are no paths from leaves to anywhere.

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+<p align="center">
+<img src="https://res.cloudinary.com/dxzx2bxch/image/upload/v1646243822/cs6120/3_dmiy5c.png">
+<p>
+<p align="center">
+<strong> The path encoding generation algorithm </strong>
+</p>
+
+Therefore, we count the number of paths from leaf to node. In other words, we visit vertices in the reverse topological order. 
+
+[Discussion]: based on the discussion in class, we think the time complexity of this algorithm can be represented as both O(n*n) and O(e). Because every edge is only visited once.
```
sampsyo:
```suggestion
[Discussion]: based on the discussion in class, we think the time complexity of this algorithm can be represented as both O(n*n) and O(e) because every edge is only visited once. 
```
Seems about right to me.

sampsyo: Also, I'd recommend using bold or something to mark your "discussion" sentences, not square brackets… otherwise, the Markdown parser will think you're trying to make a link. Like this:

```md
**Discussion:** Foo bar…
```

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+Figure (a) has two back edges I->A and inner E->B. We add dummy edges A->B and E->I, and remove the back edge E->B. Note that A is the entry block, and I is the exit block in this CFG. So the dummy edges we add are the edge from the entry block to the destination node of the back edge, and the edge from the source node of the back edge to the exit block. Through this transformation, general CFGs are downgraded to the previous CFG problem we already solved. 
+
+<p align="center">
+<img src="https://res.cloudinary.com/dxzx2bxch/image/upload/v1646243822/cs6120/5_ajv4lx.png" alt="alt_text" title="image_tooltip" style="zoom:50%;" />
+</p>
+<p align="center">
+<strong>Removing back edge and creating dummy edges for general CFGs</strong>
+</p>
+
+[Discussion]
```
sampsyo: As above.

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+
+
+### Limitations
+
+
+#### Is it efficient enough?
+
+We need to run edge profiling first and determine where we will put our instrumentation. Only after that, can we run path profiling. This additional overhead makes readers think that the Ball-Larus method would not be as efficient as it claims in the paper. 
+
+Some other works also raise questions on it: Kumar et al. [1] claim that methods like Ball-Larus algorithm take significant time to identify profile points in the program, cannot be used in dynamic (JIT) compilation.
```
sampsyo:
```suggestion
Some other works also raise questions on it: Kumar et al. [1] claim that methods like the Ball-Larus algorithm take significant time to identify profile points in the program, so they cannot be used in dynamic (JIT) compilation. 
```

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+### Limitations
+
+
+#### Is it efficient enough?
+
+We need to run edge profiling first and determine where we will put our instrumentation. Only after that, can we run path profiling. This additional overhead makes readers think that the Ball-Larus method would not be as efficient as it claims in the paper. 
+
+Some other works also raise questions on it: Kumar et al. [1] claim that methods like Ball-Larus algorithm take significant time to identify profile points in the program, cannot be used in dynamic (JIT) compilation. 
+
+We think the phase of identifying instrumentation could be accelerated by only analyzing a tiny and representative sample of the entire program because we only need a rough idea for the edge profiling result. There is also an interesting idea to adopt static program analysis methods like “constant propagation” to know some deterministic value range of certain variables to know additional information about edge profiling without any runtime analysis.
```
sampsyo: It's a good point that the initial edge profile is really just "nice to have" rather than critical for the path profiling technique's correctness/success. So it would be fine to have a "quick and dirty" approximate edge profile; the only thing that will suffer is the efficiency (not the correctness) of the eventual path profile.

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+
+2. Software test coverage
+
+    Quantify the adequacy of a test dataset by profiling a program and reporting unexecuted statements or control-flow. 
+
+
+
+#### Research work
+
+For functions with many paths, allocating an array entry for all program paths is prohibitively expensive, if not feasible [2]. Vaswani et al. [1] propose preferential path profiling (PPP): separating interesting paths from other paths and assigning a set of unique and compact numbers to these interesting paths. This enables the PPP implementation to record path information in an array instead of a hash table.
```
sampsyo: Since this is a blog post, maybe just link to the papers instead of (or in addition to?) using numbered citations like `[1]`.

content/blog/2022-02-22-profiling.md
```diff
@@ -0,0 +1,208 @@
...
+
+
+#### Real use cases
+
+Ball-Laraus used to have an [implementation](https://opensource.apple.com/source/llvmCore/llvmCore-3418.0.80/lib/Transforms/Instrumentation/PathProfiling.cpp.auto.html) in-tree in LLVM. But it was unmaintained and was removed in [2013](https://github.com/llvm/llvm-project/commit/ea564946251eb425fafde97e6398ec52a9ff6bf8). 
+
+
+#### Impact
+
+This idea of instrumentation has also inspired applications outside the compiler world. Sanghi et al. [3] uses similar packet execution path profiling method to detect anomaly data packets in data plane systems. Break Dancing [4] uses path tracing in software to mimic Intel CPU’s hardware branch tracing capabilities for ARM and AMD CPUs. Magklis et al. [5] uses dynamic path profiling to decide when to let CPU scale voltage and frequency to save power and boost efficiency.
```
sampsyo: Nice work tracking down some subsequent research that built on this stuff!

sampsyo (changes requested):
Really nice work, both of you! This turned out great overall.

I don't think it's something you need to change, but I think this post _also_ would have been just fine with a less detailed summary of the paper's technical detail. I mention this mostly so that future blog-post authors will know that they can spend less time on this part—it's OK to summarize the paper only at a high level and somewhat briefly, leaving the details to the paper itself. That way, the paper can focus primarily on commentary/discussion/context that is all new and not contained in the paper.

I have a few low-level suggestions within!

content/blog/2022-02-22-profiling.md
zzzDavid: Good point, thanks for pointing that out! I left a word out, it should be "from the leaf nodes to the root node"

content/blog/2022-02-22-profiling.md
zzzDavid: I changed to using both, clicking on the numbered citations will jump to the papers

zzzDavid:
I believe I've resolved all of the required changes. Thanks for the detailed comments and suggestions!

sampsyo (approved):
All looks good to me! Nice work!

PR #332, opened 2022-05-19
[Blog] Add blog for paper "A Type and Effect System for Deterministic Parallel Java"

barabanshek:


content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+[[extra.authors]]
+name = "Nikita Lazarev"
+link = "https://www.nikita.tech/"  # Links are optional.
+[[extra.authors]]
+name = " Susan Garry"
++++
+
+### Paper review
+
+["A Type and Effect System for Deterministic Parallel Java", Robert L. Bocchino, Vikram S. Adve, Danny Dig, Sarita V. Adve, Stephen Heumann, Rakesh Komuravelli, Jeffrey Overbey, Patrick Simmons, Hyojin Sung, and Mohsen Vakilian. OOPSLA 2009](http://dpj.cs.illinois.edu/DPJ/Publications_files/DPJ-OOPSLA-2009.pdf)
```
sampsyo: This is a somewhat awkward way open a long post… maybe you want to say something about why you're listing the title of the paper here? Or just leave it to where you eventually mention the paper in the first proper section?

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+  Susan Garry is an undergraduate student at Cornell's CS department.
+"""
+[[extra.authors]]
+name = "Nikita Lazarev"
+link = "https://www.nikita.tech/"  # Links are optional.
+[[extra.authors]]
+name = " Susan Garry"
++++
+
+### Paper review
```
sampsyo: I don't think you need this heading. (People can tell it's a paper review by reading the first bit.)

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+[extra]
+bio = """
+  [Nikita Lazarev](https://www.nikita.tech/) is an PhD student at Cornell's ECE department;
+  Susan Garry is an undergraduate student at Cornell's CS department.
+"""
+[[extra.authors]]
+name = "Nikita Lazarev"
+link = "https://www.nikita.tech/"  # Links are optional.
+[[extra.authors]]
+name = " Susan Garry"
```
sampsyo: Extra space.

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
++++
+
+### Paper review
+
+["A Type and Effect System for Deterministic Parallel Java", Robert L. Bocchino, Vikram S. Adve, Danny Dig, Sarita V. Adve, Stephen Heumann, Rakesh Komuravelli, Jeffrey Overbey, Patrick Simmons, Hyojin Sung, and Mohsen Vakilian. OOPSLA 2009](http://dpj.cs.illinois.edu/DPJ/Publications_files/DPJ-OOPSLA-2009.pdf)
+
+### Introduction
+
+Multithreading is notoriously difficult to reason about and error-prone. This happens due to potential race conditions that can be arise when two threads try to access the same region in memory simultaneously. Traditional approaches using locks (e.g. mutexes) can cause a deadlock and are also difficult to reason about and debug.
```
sampsyo: The term for that is "data race," not "race condition" (which means something more general and more vague). But I also assume you meant to say "at least one of which is a write," which is a necessary condition for a data race (two reads are fine).

sampsyo: Add a comma after "e.g." https://capra.cs.cornell.edu/styleguide/#egie

(This occurs several places throughout the post, including with "i.e.")

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+In this paper, the authors propose making concurrency in Java safe and lock-free by using a type and effect system that involves partitioning and annotating Java programs. They introduce novel mechanisms that expand on the capabilities of previous type and effect systems, allowing for concurrent operations on recursive data structures via **region nesting**, concurrent array operations via **index-parameterized arrays** and **subarrays**, and concurrent writes to the same location in memory via **commutativity annotations**. They develop a new extension for Java called DPJ (Deterministic Parallel Java) which additionally comes with the formal verification functionality. The paper demonstrates how DPJ can be used to implement realistic algorithms and verify their thread-safeness.
```
sampsyo: This paragraph starts by talking about safety in general, but that's not what makes this paper different: the focus is first and foremost on *determinism*. Everything else flows from that. To the extent that they avoid data races or other safety properties, it is *only* because it is necessary to provide determinism. So it is a somewhat bold statement saying that, if only parallel programs could be made deterministic, they would be easier to understand. You would still have other problems, but at least you would not have the problem of running the same program twice and getting different results. I think it's important to orient the technical discussion based on that fundamental goal.

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Basic Idea
+
+The paper introduces a set of annotations for unsafe memory accesses. The systems defines two main classes of annotations:
```
sampsyo: Not entirely sure what "for unsafe accesses" means… first of all, what is "unsafe"? And second, what do we mean by "for"? For avoiding them?

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+* **regions** - the programmer names portions of the heap and assigns each object and variable to a region of the heap;
+* **effect annotations** - denotes what regions a method reads and writes from, e.g. ```reads <region-list> writes <region-list>```, where ```<region-list>``` is a list of regions as defined above.
```
sampsyo: A single backtick suffices for inline code in Markdown. Like this:

```
e.g., `reads <region-list> writes <region-list>`
```

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+The code below shows an example of a class definition (a Tree class for a tree data structure) with annotations. The example is taken from the original paper, and we added more comments in it to show the functionality.
```
sampsyo, line 28: Maybe one or two sentences _after_ the example could really help clarify the high-level intuition and takeaways that the reader is supposed to get from reading the example?

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+         // overal effect: reads Links writes R
+         right.mass = mass;
+       }
+    }
+ }
+```
+
+### Region Nesting
+
+Since many datastructures are naturally recursive, DPJ comes with the functionality to express recursiveness of memory regions. This is called **region nesting**. The listing below shows how the aforementioned program can be extended with region nesting:
```
sampsyo: It might be useful to briefly mention what problem this is solving. That is, why is a flat set of regions not expressive enough, or not safe enough, or not easy enough to use? Without that motivation, this concept can sort of feel like it comes out of nowhere.

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+         // overal effect: reads Links writes L
+         left.mass = mass;
+         // overal effect: reads Links writes R
+         right.mass = mass;
+       }
+    }
+ }
+```
+
+The above code results in the following recursive region nesting as shown in the figure below. Here, the mass region M is different for every nested/recursive node.
```
sampsyo: Again, a little extra commentary telling the reader what they're supposed to learn by looking at the code and picture would be helpful. We can align these things, but what is the message here? Is there some lesson we should be learning about the system by looking at these examples?

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+<p align="center">
+<img src="figure_1.jpg" alt="alt_text" title="image_tooltip" width="500" />
+</p>
+
+### Subtyping
+
+The DPJ type system supports subtyping to better express nested regions. A summary of the subtyping rules is shown below:
+
+* Two regions are nested within each other, R1 ≤ R2,  iff R1 is contained within R2, e.g. P:M ≤ P;
```
sampsyo: You could make life a little easier for yourself (and your text look a little better) by setting `latex = true` in the front-matter configuration. Then you could do this:

```
Two regions are nested within each other, $R_1 \leq R_2$,  iff $R_1$ is contained within $R_2$, e.g., $P:M leq P$;
```

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+The above code results in the following recursive region nesting as shown in the figure below. Here, the mass region M is different for every nested/recursive node.
+
+<p align="center">
+<img src="figure_1.jpg" alt="alt_text" title="image_tooltip" width="500" />
+</p>
+
+### Subtyping
+
+The DPJ type system supports subtyping to better express nested regions. A summary of the subtyping rules is shown below:
```
sampsyo: Once again, it might be nice to tell the reader why they should care about subtyping. What situation does it arise in? What problem does it solve? Why is it interesting? A little of this framing can go a long way.

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+* Two regions are nested within each other, R1 ≤ R2,  iff R1 is contained within R2, e.g. P:M ≤ P;
+* C<R1> is a subtype of C<R2> iff every region of R1 is in R2 (R1 ⊆ R2); examples:
+C<P:M> is a subtype of C<P:*>
+C<P:L:*:M> is a subtype of C<P:*:M>
```
sampsyo: Some of these asterisks have turned into italics formatting in Markdown. Using math mode can help avoid this, of course. But you will want to preview your post using Zola to make sure it looks right.

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Supporting Concurrency in Arrays
+
+In order to allow safe concurrency in arrays, the authors define two novel capabilities: (1) **index-parameterized arrays** and (2) **subarrays**.
+
+##### Index-Parameterized Arrays
+
+Index-Parameterized Arrays allow to describe concurrent operations on different elements of an array. The problem seems to be trivial for arrays of primitive types, however, it is way harder for arrays of **mutable references**. The idea behind index-parameterized arrays is simple: we need to parameterize each object reference in the array by the array index's integer expression. For example, the aforementioned ```TreeNode``` class should be parameterized as follows ```TreeNode<Root:[10]>``` for the element ```10``` in the array. The compiler then will check (with a symbolic integer expression solver) that all array accesses in the program going to the element parameterized by expression [i] happen through the array element ```i```. If this holds, we get formally proved safe concurrent access to all the distinct elements in the array, even with mutable references.
```
sampsyo: > for arrays of primitive types, however, it is way harder

This is a comma splice. Replace the first comma with a semicolon or start a new sentence there.

https://capra.cs.cornell.edu/styleguide/#however

sampsyo: TBH I don't quite get why the problem is trivial for arrays of primitives… isn't mutating the array *itself* a problem (like `a[100] = 5` for an `int[]` array)?

sampsyo: I got a little lost toward the end of this paragraph when you start talking about `I`. I'm not sure what this symbol stands in for, nor what "parameterized by expression [i]" means. Maybe this needs a little more detail?

(Also, I'm always wary of saying "the idea is simple!" about technical stuff. It's not always as simple as it seems! And you can make your reader feel dumb if they don't agree that it's simple.)

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
```diff
@@ -0,0 +1,123 @@
...
+class IntSet<region P> {
+  void add(int x) writes P { ... }
+  add commuteswith add;
+}
+```
+defines that two invocations of `add` are atomic and either order of invocations produces the same result. The commutativity property is not getting checked by the compiler and should be verified independently with formal analysis.
+
+### Evaluation
+
+The authors evaluate their type and effect system on a set of benchmarks and show that it indeed allows to express concurrency that provides near-ideal linear scaling of performance with the number of executors. Some algorithms do not scale linearly due to certain limitations of DPJ such as impossibility of re-shuffling for array elements. Additionally, while these techniques allow some realistic concurrent algorithms to be expressed in this language, threads which write to the same location in memory must make use of the commutativity annotations, and the correctness of these annotations must be verified by the programmer.
```
sampsyo: This is one place where some specifics could be helpful. There is absolutely no need to recapitulate the entire evaluation, of course! But mentioning in a little more detail why _one_ benchmark scaled more poorly in DPJ could help make a discrete point about its strengths vs. weaknesses.

sampsyo (changes requested):
Hi, y'all! While this is a reasonable summary of the content of the paper, there are two things I found myself wishing for: (1) a description of why the reader should care about the technical concepts you introduce, and (2) commentary _about_ the paper that goes beyond summary _of_ the paper. I don't think you need to go crazy adding this stuff now, but I thought I would mention it in case this kind of assignment comes up in the future: it can be more useful to focus on stuff you bring to the table than on recapitulating the technical concepts.

Anyway, let me know when you've had a chance to take a look at some of the inline comments (and preview the post using Zola) by using the "request a review" feature on the GitHub pull request page!

sampsyo:
Hi! I know it's been a while, but would you be interested in wrapping this up so we can publish it?

barabanshek:
Hi Adrian,

Omg I completely forgot about it. Sure, let me work on it this week!

content/blog/2022-05-19-A_Type_and_Effect_System_for_Deterministic_Parallel_Java/index.md
barabanshek: I think, what they mean is that when we have an array of values (not references), and, say, multiple workers accessing these elements, determinism is guaranteed if these workers access different indexes, which is "easy" to check. I'm not exactly sure how it works, but I can imagine it's trivial for simple kernels, based on the loop bound conditions.

But for arrays of references, unique traversal is not enough, as the references (and nested references deeper in the recursion)  might change as well.

barabanshek:
OK, Updated! Please please, take a look; if you have any other suggestions, I will be happy to address them!

sampsyo (approved):
Looks awesome; thanks for revising!!

barabanshek:
Sure!!

PR #334, opened 2022-05-19
Xic Objects

### LINKED ISSUE #306: Project Proposal: Xic Objects and Automatic Dynamic Garbage Collection
# Project Proposal: Xic Objects and Automatic Dynamic Garbage Collection

## What will I do?

I will implement objects in my [Xic compiler](https://github.com/Yasgur99/xic.git) from CS 4210 and implement automatic automatic dynamic garbage collection for the language. If there is enough time, the scope can be extended to implement object based optimizations.

### Implementing Objects

There are a few relevant parts to this side of the project. First, I must modify the syntax of the language to support object-oriented-ness. I will have to decide on a syntax for this new language construct. Next, I will have to implement type checking for objects. This obviously includes defining a set of inference rules to use in type checking.

I will reach out to Professer Myers to see if he has inference rules for syntax and semantics, because I'm pretty sure implementing object orientedness is a project that has been assigned in the past. It would be much easier to use syntax and type rules that are already defined since the goal of this project is how to implement objects based on how they behave, not coming up with a way how they should behave.

Once I have added support for objects in the AST, I will have to translate them to IR. This will be a significant part of my project. How will I represent objects in IR form? This will include stuff like object layout and method dispatch.

### Implementing Garbage Collection

I will implement automatic dynamic garbage collection. I propose to implement a simple tracing collector that can serve as a baseline collector. Then I will extend on this collector by implementing fancier collectors. I plan to implement a generation garbage collector. This part will require me to dive deeper into the literature and experiment with different implementations of generational garbage collectors.

### Implementing Object Oriented Optimizations

I expect that the scope defined above will keep me busy for the duration of the semester. However, if I do happen to finish early, I will explore one (or a few) object oriented related optimizations.

## How will I evaluate success?

The main goal of this project is to evaluate the success of garbage collection. Implementing objects is a prerequisite to this evaluation. I will evaluate the performance of the generational garbage collector against the baseline tracing collector that I will also implement.

In evaluating, I would measure multiple things such as the total time it takes for the program to run, the amount of time it took for just the collector to run, and the impact of the garbage collector under different loads.

Evaluating the total time it takes for the program to run can be measured by measuring the total time it took for the program to run with the respective collector enabled. Evaluating the time it took for just the collector to run could be calculated by comparing the run time with collectors enabled versus the time it takes for no collection at all to happen. To measure the impact of the collector under different loads, I can come up with programs that use large amounts of memory in specific ways.

## Group

I do not have a partner at the moment, but if someone wants to join me in what I plan to work on, I would love to form a group!

**EDIT:** @orkosinha and I are partners for this project

orkosinha:
Closes #306


content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+name = "Orko Sinha"
+link = ""
+[[extra.authors]]
+name = "Michael Maitland"
+link = "https://michaelmaitland.com"
++++
+Building a compiler for Xi is the main project in [CS 4120](https://www.cs.cornell.edu/courses/cs4120/) at Cornell. The final assignment, now discontinued, is to extend the compiler to support feaures of the [Xi++ Language Specification](https://www.cs.cornell.edu/courses/cs4120/2019sp/project/oolang.pdf) such as object-oriented programing. Our project is an implementation of these features on Michael's compiler created for CS 4120. Our project is [here](https://github.com/Yasgur99/xic/tree/oo).
+
+# Lexing
+For lexing, there are 5 new syntax elements that we need to handle, listed below:
```
sampsyo, line 13: I think the Lexing and Parsing sections would be a lot more exciting if you were to just show a short snippet of code that uses this extension. What does the Xi++ code actually look like?

I also recommend combining Lexing and Parsing. The lexing stuff is so straightforward that it barely deserves mentioning. Better yet if you just skip over the parsing stuff as quickly as possible… knowing the syntax is useful, but I'm not sure there's much to learn from hearing about the parser itself.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+2. `extends`
+3. `this`
+4. `new`
+5. `.` operator
+
+These correspond to 5 new token types corresponding to each keyword in the language. The existing JFlex file created for the original project is extended to support these features.
+
+# Parsing
+The parser handles the new semantics and allow parsing of the following:
+1. Parsing of `Interface` that contain zero or more `use` statements and zero or more class declarations. A class declaration is the name of the class, optionally the class name it extends, and zero or more function declarations. `ClassDecl` is created as a new AST node to represent this.
```
sampsyo: It's hard to visualize what `use` statements look like without seeing examples.

sampsyo: TBH, I don't think the names of classes in your AST are all that relevant. Your reader will not be following along with your Java code, so names like `ClassDecl` are at the wrong level of abstraction for the reader to care about them.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+3. Parsing of `Exp` with the `new` and `this` keyword.
+4. Parsing of `MethodCall`s which take the form `o.f()` or `this.f()`
+5. Parsing of `FieldAccess` which take the form of `o.x` or `this.x`
+
+Additionally, two AST nodes are added to handle Interfaces and Classes.
+1. `InterfaceBody`
+2. `ClassBody`
+
+Previously, an `Interface` and `Program` contained the function declarations and definitions respectively. Previously, the parser would add to the lists directly when it came across a declaration or definition. Now that we are also handling class declarations and definitions, the parser has `fdecl_or_cdecl_star` and `fdefn_or_cdefn_plus` nonterminal to account for this. These two new nonterminals correspond to the two new AST nodes. This does not have a big input on the rest of the compiler as we just need to add a `getBody()` call in certain places that expected the function declarations and definitions to be accessed directly from the `Program`.
```
sampsyo: I get the general idea that you needed to change these nonterminals, but TBH it is pretty hard to understand exactly how without showing the actual syntax you're trying to parse.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+We also add AST nodes `FieldAccess`, `MethodCall`, and `New`.
```
sampsyo: Again, I don't think the actual class names are at the right level of abstraction for readers of this blog post.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+# Symbol Tables
+
+We extend the old symbol table that just included entries for function declarations into a symbol table that includes both function declarations, class declarations and object fields
```
sampsyo: Missing a period.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Implementation
+
+The old symbol table was represented as a `Map<String, FunctionDecl>`. Now, we introduce a class `SymbolTable` that really serves as a wrapper. It has has fields `Map<String, FunctionDecl>` and `Map<String, ClassDecl>`. By using a wrapper, we can pass around the `SymbolTable` as a single object and we shift some of the operations on the map into our new abstraction.
```
sampsyo: This is a perfect example of something that's a little too close to your code. I think the above summary is just fine: "our symbol table used to just include function declarations; now it has to include class declarations as well." The actual Java realization is not pertinent enough to include.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+This change required us to change the `SymbolTableManager` to operate on `SymbolTable` and do logic related to merging class symbols.
+
+It also requires a renaming of `FunctionCollector` to `SymbolCollector` which extends the logic of collection function to also collect classes. We use the same visitor pattern where all `ClassDecl` nodes tell the visitor to extract their declaration out of them. All other nodes that are not `ClassDecl`s do nothing when the visitor visits them.
+
+The `getValidPrograms()` function now returns a `ValidProgram` and takes one less argument. Previously it took an argument that the function mutated since it already had a different return value. I decided to create the `ValidProgram` class which lets us store both the old return value and this list of function declarations in one object. However after extending with class symbols we need a list of `FunctionDecl` and `ClassDecl`. A new object to represent this is `ProgramSymbols` which is really a wrapper over both lists.
```
sampsyo: The number of arguments to a specific function in your implementation is only really relevant to someone who has already worked with your codebase.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+# Desugaring
+The Xi++ specification states that fields and methods in a class are brought into scope automatically. To supplement this, we refer to fields as `this.x` and methods `this.f()`.  However, as per the language specifiction, we also need to support `x` and `f()` respectively from the examples prior. This is syntactic sugar. The `ThisAugmentor` created is a visitor that converts unqualified field accesses and method calls to qualified ones. Simply, it takes `x` or `f()` into `this.x` or `this.f()` respectively.
```
sampsyo: This is a good example of something at the right level of abstraction! It just tells us the problem and the solution rather than the nitty-gritty Java implementation concerns.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+## `ThisAugmentor`
+This class keeps track of the current class (since we do not allow nested classes yet, but when we do we will just keep track of stack of classes), and a stack of contexts. When we enter a new context which occurs when we see a new block of code, new function, or enter a new loop block. Any declarations or function arguments add a new binding to the current context. Then isField function can check to see if the binding is in a context. If it is, we know its not a field because the variables in the context would shadow all the fields. But if its not in any of those contexts, then we can lookup the fields in the current class context.
```
sampsyo: Knowing that you used a stack-of-contexts approach is useful. Knowing that you have an `isField` utility function is not.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+For all FExprs that are MethodCalls, qualified or unqualified must add this as the first argument.
+
+### `Var`
+
+We must lookup to see if the Var is a field and if it is we wrap it in a FieldAccess
+
+### `SimpleDecl`, `ExtendedDecl`
+
+When we see a decl we must add a new binding in the variable context. This is needed in case there is anything that should shadow the scope of a field. I have not handled this shadowing with function arguments. I wonder if the Var part of the FunctionDecl adds this to the context actually.
```
sampsyo: I wonder this too! (Is this an errant TODO?)

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+Methods all take the object as a hidden first argument. If the FunctionDecl is a method we add this as the first argument.
+
+We do not need to do this for FunctionDefn since this is encapsulated in the FunctionDecl that belongs to the FunctionDefn.
+
+For all FExprs that are MethodCalls, qualified or unqualified must add this as the first argument.
+
+### `Var`
+
+We must lookup to see if the Var is a field and if it is we wrap it in a FieldAccess
```
sampsyo: Missing punctuation.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+### `FExpr`
+
+If a function expression f() is not already marked as a method call, then we know there is no parent of this node that is a MethodCall. Then if there is a function with the name f in the class it is being called in, then we must convert it to a MethodCall where the object of the MethodCall is this.
+
+### `FunctionDecl`, `FunctionDefn`, `FExpr`
+
+Methods all take the object as a hidden first argument. If the FunctionDecl is a method we add this as the first argument.
+
+We do not need to do this for FunctionDefn since this is encapsulated in the FunctionDecl that belongs to the FunctionDefn.
```
sampsyo: These classes are not familiar to the reader.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+When we see a decl we must add a new binding in the variable context. This is needed in case there is anything that should shadow the scope of a field. I have not handled this shadowing with function arguments. I wonder if the Var part of the FunctionDecl adds this to the context actually.
+
+# IR Generation
+We introduce three new changes to the IR:
+
+1. Addition of `IRClassDefn`
+2. Add list of `IRClassDefn` to `IRCompUnit`
+3. Addition `IRDataArray`
+
+The `IRClassDefn` class contains its methods in `IRFuncDefn` form and the `IRDataArray` which corresponds to the classes's static dispatch vector. The `IRCompUnit` hoists methods out of each class so assembly can be generated for each method. The intention is that method names all have unique encodings and will never conflict with regular function names. Therefore we can output them as normal functions and all code that is generated will call using the correct encoding. More details on the ABI conventions we came up with to accomplish this below.
```
sampsyo: Is there any way to explain what's going on here in a higher-level way? What are the concepts that the IR contains now—not the *classes* it contains now? Without this explanation, the reader is forced to infer the role of `IRDataArray` and stuff from its name.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+1. `objectMethodEncodings` to store mangled function names. This allows us to store a mapping from method name to the actual ABI encoding for that function.T he encoding is the same as the encoding defined for a function or procedure in the Xi ABI except that the function name is proceeded by the class name and an underscore. See the section on DispatchVectorClassResolver for more information.
+2. `classNameToNumFields` represents a mapping from class name to the number of fields that class has
+3. `classNameToFieldNameToIndex` represents a mapping from class name to the index of the field in the object representation.
+
+This class is also responsible for building the dispatch vector and getting the method symbol out of a dispatch vector.
+
+# Dispatch Vectors
+
+Dispatch Vectors were the biggest focus during this project. According to Professor Myers, there are [3 goals](https://www.cs.cornell.edu/courses/cs4120/2022sp/notes.html?id=objects) we had:
```
sampsyo: I also think they were the most interesting part! Is there any way to organize the blog post so this interesting part gets highlighted as much as possible, even if it means removing/demoting a bunch of the earlier stuff that is less interesting?

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+1. Method calls should be fast, since they are the core operation in the language
+2. The space overhead per object should be low. We are less concerned with space overhead on a per-class basis, since the number of objects is typically much larger than the number of classes.
+3. We would like to share inherited code across classes, since for a large class hierarchy, we might otherwise end up with many copies of the essentially the same code, leading to the dreaded “code bloat”.
+
+A straightforward approach that has been used in dynamically typed OO languages like Smalltalk is to simply search for the right method definition. Each object has a hidden field that points to its class object; a method call `o.m()` is implemented by walking up the class hierarchy, starting from o's class, and stopping once an implementation for method m is found. This implementation allows sharing inherited code, so it avoids code bloat, but it is much slower than a simple function call.
```
sampsyo: Put backpacks around "o" in "o's" as well as "m"

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+For faster method dispatch, we need to precompute a data structure—a dispatch table. The key idea is to map different method names onto small integer indices, and to use those indices to look up the appropriate method code pointer in an array of code pointers. These arrays of code pointers go by various names. In C++, this array is called a vtbl, pronounced like “v-table”, standing for “virtual table". In other literature it is known as a selector table or dispatch vector.
```
sampsyo: You could consider linking to [the Wikipedia page for vtables](https://en.wikipedia.org/wiki/Virtual_method_table) or something like that.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+So consider the following classes:
+
+```
+class A           { f(); g(); }
+class B extends A { f(); h(); }
+```
+
+`A` and `B` should share the same call to `f`, `B` should inherit `g` from `A`, and `B` should also define `h`.
+
+We can put `f` and `g` in indicies 0 and 1 respectivley for both dispatch vectors and `h` is put in index 2 for the dispatch vector of `B`. Then, if there is a call to `f`, `g`, or `h`, we can get the pointer to the correct function in `O(1)` time by generating IR that resembles the following
```
sampsyo: Consider adding a `:` to the end of this sentence.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+```
+MOV dv_ptr, o // Get the pointer to the dispatch vector since it is at index 0 of the object o
+MOV dv, [dv_ptr] // Get the dispatch vector
+CALL [dv + index] // Call a function based on its index in the dispatch vector
+```
+
+We build two classes to assist with building dispatch vectors:
+
+First we need to figure out when given some class `A`, and some function `f`, which classes `f` should we call. For example, if `f` is redefined by a subclass then we want to call the `f` of that subclass. But if `f` is defined by a super class we want to call `f` of that super class. This is all handled by the `DispatchVectorClassResolver`
```
sampsyo: classes -> class's

sampsyo: Missing a period.

sampsyo: This is the first we've heard of `DispatchVectorClassResolver`. Can you explain what it is? Is it an IR->IR pass?

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+The next thing we need is given all methods for a given class, what index do they go in in the dispatch vector. For example if `A` defines `f` and subclass `B` also defines `f` they must be at the same index. the `DispatchVectorIndexResolver` handles this.
```
sampsyo: As above. Is `DispatchVectorIndexResolver` just a global resource of the compiler, or is it a pass of some kind? What is it?

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+Both of these classes have been well tested.
```
sampsyo: What is "well tested"? Tell us exactly what you did. Unit tests? Full-program tests? How did you decide you crossed the threshold from "adequately" to "well"?

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+In Xi, static data is represented by an `IRData` class. Out of the box, `IRData` is not a traditional `IRNode` (it does not extend `IRNode`. This means its children nodes are not `IRNode`s. Instead, it has a field `long[] data`. We tried to encode the dispatch vector as the concatenation of function names seperated by the null terminator. Again, we ran into the problem that when we do `call classADVLabel` it only reads the first word of data which corresponds to the first character.
+Even worse, the interpreter tried to jump to the instruction at the address corresponding to that character instead of looking up that character in the function defn to address table and jumping there. I wrote a version of the interpreter that read until the next null terminator and then did the correct lookup. This got things to worked but it felt hacky because we could no longer write IR that jumped to an address -- it required we jumped to a function that was named as the string starting at that memory address until the next null terminator byte in memory.
```
sampsyo: I think it would be good to pop up a level here and describe what went wrong with both of these approaches. Namely, there was a core conceptual confusion: you were putting strings where you want to put (and where "real" vtables put, as described above) addresses. Making this confusion crystal-clear will undoubtedly help other people down the line understand the issue (and perhaps vtables in general) better.

sampsyo: Consider replacing ` -- ` with `—` (an [em dash](https://capra.cs.cornell.edu/styleguide/#emdash)) for consistency with your usage above.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+We explored trying to put the memory address of the function in the IRData instead of the name. That way, we would only need to read one word and the interpeter would jump there as it wanted to do. The problem with this approach is that when we convert to the IR, we have no idea about the memory layout. The memory addresses are defined by the interpeter and would be different than the memory scheme if we went all the way to assembley. Additionally, there was no way to know about the memory layout of the simulator when converting to IR because it does not initialize its memory until after the IR is created.
```
sampsyo: It might be handy to clarify earlier that you are not making a native-code compiler in this project (the reader could be forgiven for assuming that).

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+We had two options: 
+
+1. Modify `IRData` to allow us to have `IRName` or `IRConst` or `IRMem` nodes at each element in the data array instead of `long`s. 
+2. Initialize a non-static dispatch vector for each object on construction so we could use `IRName`, `IRConst` and `IRMem` to represent the dispatch vector and offsets into it.
+
+Both of these approaches would mean there were no changes to the `interpet` function of the interpeter, would allow the generated IR to work for both the `IRSimulator` and for assembley generation. The first approach however requried new classes to be created and modification of the `initialize` function of the `IRSimulator` to work with the new data class.
```
sampsyo: > no changes to the `interpet` function of the interpeter

This is another perfect example of the "wrong level of abstraction" problem. Just say that it was your goal to avoid hacking the already-implemented IR interpreter! Your reader has no idea what the `interpret` function does and why it's important not to change it.

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+We opted for approach 2 for simplicity, although approach 2 is definetly the better approach. We discuss this more in future work section.
```
sampsyo: "definetly" (please spell check)

sampsyo: Maybe explain a little more about why one is simpler and the other is better?

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+undefined memory access.
+
+In order to remedy this problem, we must allocate fields belonging to super classes at runtime since fields are not part of class declarations and are only part of class definitions by the Xi++ specification. I spoke with Professer Myers about this. C++ solves this problem by requiring public and private fields to be declared in the header files (the class declaration) so this information is known at compile time. Java on the otherhand does this at runtime by having a bytecode instruction that refers to the variables by name. In Xi, the runtime would probably need to do something similiar to the dispatch vector indexing scheme where all variables of a subtype are placed at a certain index in the object layout at runtime.
```
sampsyo: I don't quite understand this conclusion: why is it necessary to allocate space at run time? It seems like the "superclass chain" for every class is known statically, so you could just walk that and gather up all the fields for all the classes. You would need to lay out the data fields carefully so that a, if `A` inherits from `B`, then an `A` object can be safely accessed through a `B` reference… but that's just a matter of appending subclass fields at the end.

I also don't quite understand the distinction between definitions and declarations in this context:

> fields are not part of class declarations and are only part of class definitions by the Xi++ specification

…but maybe that's because I don't know Xi well enough.

sampsyo: It is true that Java bytecode seems to contain field names, but it is not true that the compiled code actually dynamically uses that (string) name…

content/blog/2022-05-17-object-oriented-features/index.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Static Dispatch Vectors
+
+As discussed above, much of the work that went into this project was with respect to the representation of dispatch vectors in the IR. I met with Professer Myers and he agreed that his intention is for these dispatch vectors to be static data but there is no way to represent them using the current version of `IRData` since `IRData` requires the physical address of functions but these are not determined until runtime. The solution is to create a different `IRData` class which can represent static data that is of the form `IRNode` instead of physical addresses. This would allow us to say the data is an array of `IRName`s which would be converted to the address of the correct function at runtime.
```
sampsyo: Care to comment on how this approach compares to what you actually did, in terms of effectiveness and effort?

sampsyo (changes requested):
Sounds like a mostly successful project!

I have one big theme in my comments on your writeup, however. While the vtable discussion was interesting and clearly described both the problem and the solution, the stuff that came before that is written at a very low "level of abstraction." That is, it is about the nitty-gritty internals of the actual Java hacking you did—it may be readable/interesting to someone who has already been working on the codebase a lot, but it is really hard to follow for readers who did not just spend a couple of months working on your Xi compiler.

Is there some way to reorganize things to highlight the interesting parts more? Currently, the organization seems pretty linear, i.e., chronological: we hear about every phrase exhaustively, in order, rather than being directed toward the most important ideas first. Is it possible to imagine that you are an average 6120 student who has *not* taken 4120, and certainly has not seen your code already, and think about what they would care about or be able to understand?

To a first approximation, I would remove all references to all class names and methods in your implementation. We don't know what those are. But that is just a microcosm of the overall advice, which is to refocus the description on reusable compilers lessons rather than codebase-specific logistics.

michaelmaitland:
I made changes based on comments. I do not have permission to request review again from @sampsyo since Orko is the PR owner. I did my best to remove specific references to class names where it was not needed. I kept some reference to class names, primarily in the IR section. I explained what each IR node was so the reader did not have to infer. The IR is more a specification that exists outside of this project and I was modifying that specification, which is why I found it relevant to keep class names included here.

sampsyo (approved):
This looks like a lot of extra detail—thank you for adding all this! If either of you want to improve this after the fact, I'm happy to merge further PRs.

PR #337, opened 2022-05-20
Project Report: Bril Bytecode Interpreter

### LINKED ISSUE #304: Project Proposal: Bril Bytecode Interpreter
**What will you do?**
We will implement a bytecode interpreter for bril using C. If we have time, we will extend this to emit assembly, and target x86 and ARM.

**How will you do it?**
First we will create a bytecode for bril. This will limit the size of the program in terms of the number of variables, labels, and instructions for functions, but as of yet there does not exist a bril program that uses $2^{16}$ variables. It will also be challenging to support the speculation extension of the bril language, but ideally we would like to include this as well. Translating to assembly should be relatively easy due to the structure of the bytecode.

Having a working interpreter is a must. We also aim for the interpreter to be faster than the provided brili interpreter, and for the assembly to be even faster than that. 

**Team members:**
@charles-rs and @susan-garry

charles-rs:
here we go!

closes #304


content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
++++
+title = "Bril bytecode interpreter"
```
sampsyo: Please use Title Case for blog posts.

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+[[extra.authors]]
+name = "Charles Sherk"
+link = "https://github.com/charles-rs/"
+[[extra.authors]]
+name = "Susan Garry"
+link = "https://github.com/susan-garry"
++++
+
+## Motivation
+Currently we have access to `brili`, which is written with an emphasis on
```
sampsyo: Maybe one additional sentence would be useful to say what `brili` even is? You could just say "We are on a quest for a fast interpreter for the Bril intermediate language…"

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+readable code and extensiblity, and `brili-rs`, which is supposed to be faster,
```
sampsyo: I think it's called `brilirs`, with no hyphen.

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+but `brili-rs` isn't actually that fast, and seems to be rather bloated for what
```
sampsyo: "Bloated" seems a little pejorative! I think it's just pretty general and meant as a showcase for `bril-rs`, the underlying library that lets people do their homework in Rust. It's, like, an interpreter written in Rust, which happens to be a fast language—not an interpreter hyper-optimized for speed.

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+it is. There were some points in this course where time was spent on "waiting
+for the Bril interpreter" that could have been avoided. So, we propose and
+implement an interpreter for Bril whose sole focus is speed.
+
+## Design overview
+
+### Language choice
+We chose C for an implementation language as it is known for its speed as well
+as simplicity, which means we won't inadvertently write code that is more
+expensive than necessary.
```
sampsyo: Fine, but not sure how relevant this is. If you imagine yourself as a generally compiler-curious person stumbling across this blog posts, will they care why you chose C? I think they will care more about the virtues of a bytecode format for eliminating the overhead of AST interpretation.

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Big Red Bytecode
+In order to make our interpreter as efficient as possible, we have designed a
+very compressed representation of bril instructions: (almost) all of them are
```
sampsyo: bril -> Bril (throughout)

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+represented in 64 bits, with some taking multiple 64 bit words as needed. See
+the [documentation](https://github.coecis.cornell.edu/cs897/fast-bril/blob/master/doc/brb.pdf).
+
+In order to appease the C typechecker and not rely on undefined behavior, we use
+a union type to internally represent instructions, but most have the first 16
+bits as the opcode, so we can pull this out and then determine how to decode the
+rest of the instruction. In principle, we could also use `uint64_t` or
+`uint16_t[4]`, but using unions is the most user friendly, and lets the C
+compiler do more work to access the fields of structs, in case there are any
+cool tricks for that behind the scenes.
```
sampsyo: The use of unions in the interpreter is subordinate to the design of the encoding itself. This paragraph gives a *hint* as to what the encoding looks like, but it doesn't quite spell it out. Can you show us an example or two of what the encoding actually looks like? Maybe even just an `add` instruction? Doing so would force you to describe how variables work in your bytecode, which would be a huge expansion in terms of specificity.

Again, imagine the audience as a generally compilers-curious person. They will want to know a teensy bit more about what this whole "bytecode representation" business really means, and examples could help a lot with that.

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Features
+We support a superset of the behavior provided by `brili`, so options like `-p`
+work exactly the same. We also support the following:
+ - `-b` will read in bytecode instead of the Bril json
```
sampsyo: json -> JSON

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+ - `-bo <file>` will output the bytecode to `<file>`
+ - `-pr` will print the program to standard out (probably more useful with the
+         `-b` option)
+ - `-ni` will NOT run the interpreter.
+ - `-e <file>` will emit assembly to `<file>`.
```
sampsyo: I know this is too low-level for a blog post about the virtues of bytecode interpretation, but I am still confused after reading this list about the combination of flags necessary to accomplish some stuff, like this:

- Translate JSON to bytecode. Probably `fastbrili -bo < foo.json`? Maybe I have to add `-ni`? Maybe I need an output file?
- Translate JSON to JSON. Probably `fastbrili -pr < foo.json`? Also maybe needs `-ni`? Or maybe it *also* needs `pr`? Not sure.
- Interpret bytecode. Maybe `fastbrili -b < foo.bc`??

In other words, I think these options could be made a bit more orthogonal…

"Emit assembly" is also a perplexing to see here. So far we've been talking about interpretation. It also emits assembly? If so, you should say this sooner, or maybe later in the compiler section, or something!

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+
+### The Assembly situation
```
sampsyo: Maybe this section should be titled "fastbrilirs, despite the name, is also a compiler to x86 assembly"?

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+We translate temporaries to consecutive natural numbers (the first seen is 0,
+then 1, etc.), and we couldn't help but notice that these numbers could easily
```
sampsyo: This thing about translating "temporaries" (do you mean Bril variables?) into numbers deserves a mention when you describe the bytecode format earlier.

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+undefined behavior. Most of the type information is erased and then inferred
+(e.g. if you use `add` on two floats, it will not check, assume they are ints,
+and run integer addition on the values), but we do deal with all correct Bril
+programs correctly. We believe that this isn't much of an issue, as `brili`
+isn't actually that helpful with these errors, though we do acknowledge that
+`fastbrili` is worse.
+
+### Extensibility
+One of the really cool things about Bril is that it is relatively easy to
+extend. Our bytecode is designed in a way that it supports $2^15$ opcodes, which
```
sampsyo: To use math, you'll need to include `latex = true` in the configuration at the top. (Please preview your post using Zola.)

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+         `-b` option)
+ - `-ni` will NOT run the interpreter.
+ - `-e <file>` will emit assembly to `<file>`.
+
+### The Assembly situation
+We translate temporaries to consecutive natural numbers (the first seen is 0,
+then 1, etc.), and we couldn't help but notice that these numbers could easily
+be multiplied by 8, and then just turned into stack locations. Also, there isn't
+currently a true compiler for Bril all the way to machine code, so we may as
+well add one.
```
sampsyo: Well:
- https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/run-bril-on-raspberry-pi-natively/
- https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/llvm-backend-for-bril/
- https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/bril-to-llvm-ocaml/
- https://www.cs.cornell.edu/courses/cs6120/2020fa/blog/brilc/

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+merge into the bril repo).
+
+Our assembly does depend on the C standard library for functions like `malloc`
+(to support the memory extension) as well as a few printing functions as doing
+it all with system calls just didn't seem like a great use of time. We have some
+functions in the `lib/` directory for printing base types, as well as parsing
+them in order for the `main` function to read command line arguments. A future
+improvement would be putting `malloc` and other C functions we call into this
+file so that we could keep a compiled version around and not have to link with
+C.
```
sampsyo: This whole compiler is very impressive in that it works, but this section is pretty low on details about how it actually works. Could you maybe include an example or two of how the instructions are translated, or some other way to give people insights into what was hard/interesting/reusable? The general idea of turning a bytecode interpreter into a compiler is very important to the world more broadly, so elucidating the insight with a bit more crunchy technical detail could be very useful!

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Implementation
+Our interpreter consists of an initial pass to translate into this bytecode,
+which turns temps and labels from strings into numbers, and reduces the size of
+the code overall. The bytecode instructions are stored in an array so that we
+can benefit from cache locality. We then interpret them by iterating through
+this array. One minor inconvenience is that certain instructions, such as print,
```
sampsyo: Consider putting backticks around references to code like `print`

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+
+For the armv8 assembly generation, we initially implemented it as a single pass
+over the bytecode that printed the instructions as it went. This is fine, but it
+leaves little room for extensions and improvements, so in order to get started
+on register allocation we refactored to have an internal representation of arm
+instructions. For this, we went with a tagged union, as it is similar to a
+variant which is very convenient for this, as learned in CS 4120.
+
+## Evaluation
+We used the Bril benchmarks from the repo, and used clock time to evaluate the
```
sampsyo: It's often called "wall-clock" time (if that's what you mean), to distinguish it from the CPU's cycle clock.

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+difference. This isn't exactly optimal as most of these benchmarks are very
+small, so the startup time might be significant, but the startup time should
+hurt our interpreter the most, as it has the translation phase. Our results are
+[here](https://github.coecis.cornell.edu/cs897/fast-bril/blob/master/eval/performance.ods),
```
sampsyo: Especially because the broader world does not have access to Cornell GitHub, and especially especially because I don't have OpenOffice installed, can you include some kind of visualization? (A bar chart or something?)

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+in ods format. On average, `brili-rs` has an 84% speedup over `brili`, and
+`fastbrili` has a 77% speedup over `brili-rs`. `fastbrili` also has a 96%
+speedup over `brili`, so I think we can call this a job well done.
```
sampsyo: I don't think you mean a *speedup* of 77% = 0.77, which would indicate that fastbrili was slower. ["Speedup"](https://en.wikipedia.org/wiki/Speedup) is defined as `<baseline time>/<your time>`, so it is greater than 1 when your thing is faster.

Also, what do you mean by "average"? Remember to use the geometric mean or harmonic mean, never the arithmetic mean, when averaging ratios…

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Drawbacks
+While we would love to say that `fastbrili` is the answer to all of your Bril
+needs, there are some minor drawbacks, mainly that invalid Bril programs cause
+undefined behavior. Most of the type information is erased and then inferred
+(e.g. if you use `add` on two floats, it will not check, assume they are ints,
```
sampsyo: Add a comma after "e.g." https://capra.cs.cornell.edu/styleguide/#egie

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+and run integer addition on the values), but we do deal with all correct Bril
+programs correctly. We believe that this isn't much of an issue, as `brili`
+isn't actually that helpful with these errors, though we do acknowledge that
+`fastbrili` is worse.
```
sampsyo: I think it's important to reference the semantics for the Bril language, which explicitly permit this kind of behavior (i.e., they are only defined for well-formed programs).

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Extensibility
+One of the really cool things about Bril is that it is relatively easy to
+extend. Our bytecode is designed in a way that it supports $2^15$ opcodes, which
+is room for plenty of extensions, but if you wanted to add more types this would
+be considerably more difficult. We rely on being able to represent types in 16
+bits, by using 2 of them for the "base type" and then the other 14 for pointer
+depth. Realistic $2^14$ is deeper than any pointer will be nested, so adding a
```
sampsyo:
```suggestion
depth. Realistically, $2^14$ is deeper than any pointer will be nested, so adding a
```

content/blog/2022-05-19-fast-brili.md
```diff
@@ -0,0 +1,141 @@
...
+few more base types wouldn't be too much of an issue, but adding more complex
+types would be a serious complication. This could be solved by having one bit of
+tag for whether the type is complex, and then using the rest of the bits to
+index into some memoized array of the types that show up, but this would require
+a bit of work.
```
sampsyo: As a counterpoint to the previous section, signaling useful errors for *these* limitations seems important. (In other words, silently doing the wrong thing for programs that exceed these limits would be bad.)

sampsyo (changes requested):
Super cool, y'all; I'm thrilled you got the interpreter working and a compiler built on top of it to boot. One question is: does the compiler actually work? We didn't get performance numbers for it.

A couple of things seem important to add here:

* *Some* amount of detail about what the bytecode format looks like. The general design strategy would suffice; no need for a full reference manual. But at the moment it's pretty unclear how this works.
* A bar chart! You have performance data collected for a lot of benchmarks; visualizaing it in its full glory seems pretty important.

Really nice work!

charles-rs:
Does the compiler work? - yes, it passes all of our tests compiling to armv8-a with no register allocation.

Well, then why aren't we seeing performance numbers?? - Ran the benchmarks on my (much faster) x86 chip, so adding the compiled code performance would be apples to oranges, so i would have to rerun everything. I didn't do this since the arm chip is slow.... I'll do it now though

The bytecode format: this is documented separately [here](https://github.coecis.cornell.edu/cs897/fast-bril/blob/master/doc/brb.pdf), so I didn't duplicate the info. I'll make the link more clear.

The bar chart: yes. good idea.

content/blog/2022-05-19-fast-brili.md
charles-rs: fair enough, it's presented [here](https://capra.cs.cornell.edu/bril/tools/brilirs.html) as "fast interpreter" and i may have read a little much into that lol

content/blog/2022-05-19-fast-brili.md
charles-rs: think this is resolved by the documentation pdf?

content/blog/2022-05-19-fast-brili.md
charles-rs: oops :)

content/blog/2022-05-19-fast-brili.md
charles-rs: ok i was doing something stupid where i did (fast - slow)/slow. no idea where i got this. fastbril runs in .02 of the time of brili, and .23 of the time of brilirs

content/blog/2022-05-19-fast-brili.md
charles-rs: ok will add

charles-rs:
ran the benchmarks on my arm laptop. 10 seconds of timeout wasn't enough....

content/blog/2022-05-19-fast-brili.md
charles-rs: think this is resolved by the documentation pdf?

content/blog/2022-05-19-fast-brili.md
charles-rs: not sure about the current status of x86...

content/blog/2022-05-19-fast-brili.md
charles-rs: `fastbrili -bo < foo.json` will print bytecode to the file `my-output`, and run the interpreter
`fastbrili `-bo` foo.bc < foo.json` will print bytecode to the file `foo.bc`, and run the interpreter

`fastbrili -pr < foo.json` will print foo's program to stdout, and run the interpreter. it's the text representation not the json right now... this was initially a debugging feature, but i really should change it to json

spot on for interpreting bytecode.

perhaps options that cause "something else" to happen should disable the interpreter by default, removing the `-ni` flag altogether? probably not that often someone wants to interpret and compile all in one go

content/blog/2022-05-19-fast-brili.md
charles-rs: again hopefully resolved by documentation

sampsyo:
Woohoo! Nice bar chart!

While linking to the reference for the bytecode format is nice for the complete details, the thing I think would be useful to have here (to make the post self-contained) is just an *overview* of the *strategy*. There are lots of degrees of freedom in how the bytecode might be designed, and some "big questions" about how to even think about an appropriate bytecode (should the types be kept? how do you preserve string names of things, if at all? how do you index functions and labels—the same way as variables?). Instead of giving an exhaustive description of the bytecode here, just giving us enough information to understand the overall design would help a lot. (That's why I mentioned the idea of showing off a single instruction… like, you could demonstrate what a simple `x: int = add x y` looks like once encoded, and that would answer a lot of these questions.)

charles-rs:
ohhhh ok i understand now sorry for the confusion. This makes sense.

content/blog/2022-05-19-fast-brili/index.md
```diff
@@ -39,6 +39,51 @@ very compressed representation of Bril instructions: (almost) all of them are
...
+functions with arguments, the first argument is 0, then 1 et cetera, and the
+variables used in the body follow. This allows the following encoding of "most"
+instructions:
+
+<p align="center">
+	<img src="insn-layout.png"/>
+</p>
+
+By "most" instructions we mean the instructions of the form `t1 = t2 op
+t3`. `arg1`, `arg2`, and `arg3` are the aforementioned numbers of the
```
sampsyo, line 53: I think you might mean `t1: type = op t2 t3` to match Bril syntax, but NBD.

content/blog/2022-05-19-fast-brili/index.md
```diff
@@ -39,6 +39,51 @@ very compressed representation of Bril instructions: (almost) all of them are
...
+very well. We will have the instructions of a function in an array of 64 bit
+bytecode instructions, so we represent labels as indices in this array. We also
+have to keep track of which instructions had a label in the original program so
+that we can interpret the φ instruction efficiently.
+
+We notice that nearly all operations are coincidentally encodings of type
+information. For example, `add` only operates on `int`
+operands/results. However, the `id` instruction is an exception to this, but it
+only has one argument so we encode the type in the space for the second
+argument. See the more detailed docs for the type encoding.
```
sampsyo, line 66: Interesting! It also occurs to me that it might have been possible to just not record the type in the `id` at all because it doesn't affect how the program executes… except that then you couldn't recover the original Bril JSON, at least not without doing some global type inference.

content/blog/2022-05-19-fast-brili/index.md
```diff
@@ -39,6 +39,51 @@ very compressed representation of Bril instructions: (almost) all of them are
...
+
+We also have a translation from operations to numbers. These are provided in
+config files in our repo, as well as our documentation PDF.
+
+Some Bril instructions simply won't fit into a 64 bit (or even constant sized)
+word, so they have variable length encoding. The `call` instruction is a prime
+example of this. Function calls may have any number of arguments, so the first
+word contains the information about how many words following it are part of this
+instruction. Everything is always a multiple of 64 in order to maintain
+alignment.
```
sampsyo, line 76: Yeah, that's a super interesting example. I suppose one option would be to invent a pseudo-op like `push_arg` that repeats for each function argument. But it's not clear this is any better than a variable-length encoding.

sampsyo (approved):
Looks great! Nice work!!! Publishing now. :tada:

content/blog/2022-05-19-fast-brili/index.md
charles-rs: this is true for the `id` instruction, so perhaps it was a bad choice. I originally didn't have this, but then thought i would want it for asm generation (floats need different registers), but then realized that if i was just moving a float from memory to memory, this can go through a general register. This might be useful info to have around for register allocation, but im not sure.

The `call` instruction  on the other hand DOES need this information to be conformant to most calling conventions, as float args are not passed through general purpose registers, so maybe that would have been a better example

If we don't record the type in `id`, then we have no use for those 16 bits, so other than making the translation a teensy bit more expensive, there is nothing lost by recording it

content/blog/2022-05-19-fast-brili/index.md
charles-rs: ah now this is actually a very good idea. Solves the problem w/ induction variables i mentioned somewhere else...

content/blog/2022-05-19-fast-brili/index.md
sampsyo: Huh, interesting. It would definitely be a bigger change, but one option would be to associate a variable->type table with every function (because Bril [disallows changing the type of a variable within a given function](https://capra.cs.cornell.edu/bril/lang/wellformed.html)) and then just let instructions look it up there when they need it.

PR #340, opened 2022-05-23
Project Report: Memory Optimization and Profiling for MLIR-based HeteroCL

### LINKED ISSUE #311: Project Proposal: Memory Optimization and Profiling for MLIR-based HeteroCL with Polyhedral Model
### HeteroCL and MLIR
[HeteroCL](https://github.com/cornell-zhang/heterocl) is a programming infrastructure composed of a Python-based domain-specific language (DSL) and a compilation flow that targets heterogeneous hardware platforms. HeteroCL provides a clean abstraction that decouples algorithm specification from three important types of hardware customization in compute, data types, and memory architectures. [MLIR](https://mlir.llvm.org/) stands for Multi-Level Intermediate Representation, which is a modular compiler infrastructure that enables different optimizations performed at different levels of abstraction, and is part of the [LLVM](https://github.com/llvm/llvm-project) project.

### Roofline Model
The [roofline model](https://en.wikipedia.org/wiki/Roofline_model) is a performance upper-bound model for a given compute kernel or a program running on a platform. It depicts the relationship between the attainable compute performance (GOPs/s) and operation intensity (OPs/Byte). Operation intensity is arithmetic operation per byte of off-chip memory access. It is often used to analyze whether a compute kernel is memory-bound or compute-bound. 

### Polyhedral Model 
In the programming sense, a [polyhedral model](https://en.wikipedia.org/wiki/Polytope_model) represents the iteration space of nested loops as a mathematical object called polyhedra. Each loop iteration is a lattice point in the polyhedra. Polyhedral model is widely used in loop optimizations, program scheduling, and memory access analysis.

### Read and Write Buffer Generation
To ameliorate the cost of off-chip data access, we can build read buffers to store data for future reuse, and write buffers to decouple memory write back and computation. FPGAs enable us to build customized memory hierarchies to suit the need of computation kernels. An example of a read buffer is the line/window buffer in a typical 2D convolution kernel implementation on an FPGA. 

## **What will you do?**
The original HeteroCL implementation for memory primitives (1) only supports *read* patterns, (2) is not systematic for high-dimensional stencils, and (3) also does not provide guidance to the users on how to optimize the program. To make HeteroCL target more general programs and further improve the performance, we plan to leverage the [polyhedral model](https://en.wikipedia.org/wiki/Polytope_model) to analyze memory access patterns, generate on-chip buffers, and profile the performance. Basically, we will do the following three subtasks:
1. Generate read buffers for the [`reuse_at`](https://heterocl.csl.cornell.edu/doc/tutorials/tutorial_06_memory.html) primitive: Lots of application may have memory access pattern that incurs data reuse when traversing the arrays. For example, 1D [stencil](https://en.wikipedia.org/wiki/Stencil_(numerical_analysis)) kernel `A[i-1] + A[i] + A[i+1]` may have two duplicated elements to access in two consecutive iterations. For convolution pattern `A[i+rx][j+ry] * B[rx][ry]`, there will be six elements of array `A` that can be reused from the previous iteration. Thus, generating buffers for those arrays and reusing the data can greatly reduce redundant memory accesses and improve the final performance.
2. Generate write buffers for the [`buffer_at`]() primitive: For the GEMM example, `C[i][j] += A[i][k] * B[k][j]`, if no buffers are specified, we need to load `C` from off-chip memory, compute and add the result, and store back to memory, which has large overheads. Instead, we can create an intermediate buffer for `C` to reduce memory loads. 
3. Memory profiling: We can generate [roofline model](https://en.wikipedia.org/wiki/Roofline_model) based on the machine specifications and the calculated performance, and see how different design points are shown on the figure.

## **How will you do it?**
We have already designed and implemented the HeteroCL MLIR dialect facilities in a private repository, so we only need to write passes for those read/write optimizations. As mentioned above, we will leverage the polyhedral model to analyze read/write instructions and generate corresponding read/write buffers to reduce memory access from off-chip memory. More specifically, we will implement the above optimizations in the following way:
1. Read buffer generation (`reuse_at`): We extract read patterns from the polyhedral model and generate reuse buffer axis by axis. Specifically, suppose we have an input tensor of shape `(a1, a2, …, aN)` and a stencil kernel that can be covered by a rectangular hull of shape `(s1, s2, …, sN)`, and it exists reuse patterns in each of the dimensions, then we will generate a sequence of hybercube buffers with the following shapes. The figure below shows a 2D example of generated line buffers and window buffers. For more information, please refer to the [Memory Customization Tutorial](https://heterocl.csl.cornell.edu/doc/tutorials/tutorial_06_memory.html#combine-window-buffer-and-linebuffer).
```
(s1, a2, ..., aN) // reuse at the 1st axis
(s1, s2, ..., aN) // reuse at the 1st and 2nd axis
...
(s1, s2, ..., sN) // reuse at all the axes
```

2. Write buffer generation (`buffer_at`): We extract write patterns from the polyhedral model. Based on the user-specified axis, we generate write buffers in the corresponding place. For example, for the classical GEMM case, we can generate a write buffer in the 1st axis as shown below.
```cpp
// before transformation
for (i, 0, 2)
  for (j, 0, 2)
    for (k, 0, 2)
      C[i, j] += A[i, k] * B[k, j]

// after transformation
for (i, 0, 2) { // non-reduction
    allocate buf_C[2] // on-chip buffer
    // 1. initalization
    for (j, 0, 2) // non-reduction
        buf_C[j] = 0
    // 2. computation
    for (j, 0, 2)
        for (k, 0, 2) // reduction
            buf_C[k] += A[i, k] * B[k, j]
    // 3. write back
    for (j, 0, 2)
        C[i, j] = buf_C[j]
}
```
3. Profiling: We can obtain the memory access instructions from polyhedral analysis and calculate the number of memory access for one specific kernel. By giving the memory bandwidth and peak performance of the target device, the resulting arithmetic intensity `I` is the ratio of MAC operations to total data movement (MACs/byte). Finally, we can plot the results in a figure.

### Read buffer generation
We measure the success of read buffer generation by generating reuse buffers for stencil programs such as 2D convolution and pipelining the kernel to 1-cycle initial interval (II). The output of our compilation framework is Vivado HLS code. We will validate the kernel with read buffers through simulation and synthesis. 

### Write buffer generation
Similarly, we measure the success of write buffer generation with a case study of 2D convolution with [interleaving accumulation](https://ieeexplore.ieee.org/document/9264692/). The goal is to interleave the matrix multiplication or convolution operations by creating an output line buffer and loop reordering so that the initial interval of the design can also achieve II=1.

### Roofline model generation
The goal of the roofline model generation is two-fold. First, we plot the roofline given the off-chip memory bandwidth and hardware resource. Second, we profile the compute kernel’s memory access and analyze the proportion of arithmetic operations and memory access to estimate the kernel’s operational intensity, which is represented as a data point under the roofline model plot.

## **Team members:**
Hongzheng Chen @chhzh123 
Niansong Zhang @zzzDavid 
Jiajie Li @tonyjie

chhzh123:
Close #311


content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+<center><img src="rw_mem.png" alt="alt_text" title="image_tooltip" style="zoom:30%;"></center>
+<center>Fig. 2: Reuse buffer and write buffer</center>
+
+Fig. 2 is an overview of reuse buffer and write buffer generation with `reuse_at` and `buffer_at` primitives. We will go into details in the following sections.
+
+## Reuse Buffer
+The original HeteroCL paper only discusses how to generate reuse buffers for a simple 2D convolutional kernel, while real-life applications generally have high-dimensional arrays with complex access patterns. In this project, we try to extend the idea of reuse buffer in the MLIR framework to support more applications. In this section, we will first discuss the high-level design of reuse buffer and detail the implementation.
+
+### Design
+The basic idea of reuse buffer is to reuse data between adjacent loop iterations so that the number of off-chip memory accesses can be reduced. If without special notification, we suppose the traversal order is the same as the memory access order in this report, i.e., always access the last dimension first and then access the second last, etc.
```
sampsyo: "If without special notification" -> something like "Unless otherwise noted"

sampsyo: By "second last", do you mean "second-to-last"?

content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+
+The applications that have data reuse are mostly *stencil* kernels, so we follow SODA[^3] to give a similar definition. Consider a point $\mathbf{x}=(x_0,x_1,\ldots,x_m)\in\mathbb{N}^m$, a $n$-point stencil window defines a set of offsets $\\{\mathbf{a}^{(i)}\in\mathbb{N}^m\\}\_{i=1}^{n}$ that describe the distance from $\mathbf{x}$. By adding the offset to a specific point, we can get the actual points $\\{\mathbf{y}=\mathbf{x}+\mathbf{a}^{(i)}\\}_{i=1}^n$ of a stencil in a specific iteration. The value of these points will be reduced (e.g., mean, summation) to one value in the output tensor.
+
+We further define the *span* $s_d$ of a dimension $d$ as the largest distance between two stencil points along that dimension.
+
+$s_d=\max_{i}(\mathbf{a}^{(i)}_d)-\min_i(\mathbf{a}^{(i)}_d)+1$
```
sampsyo: If you use `$$` around your standalone math equations (not inline math), it will use "display mode", which means the math gets centered and perhaps looks a little nicer.

content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+
+The minimal size of the reuse buffer consists of small rectangular strips whose total length is equal to the reuse distance of the stencil [^3]. In the five-point stencil example, this requires generate buffer covering elements from $(0,1)$ to $(2,1)$. The total size of the buffer is $4+6+3=13$. However, this may introduce complicated control logic when implemented on FPGA. Thus, for simplicity, we use the *rectangular hull* of the stencil as the reuse buffer, which is labeled as <font color="red">red frame</font> in Fig. 3(a). We call it *window buffer*. The shape of this buffer is $[s_0,s_1,\ldots,s_m]$. We can always reuse data when the buffer moves along the same row.
```
sampsyo: Footnotes go after the period. https://capra.cs.cornell.edu/styleguide/#footnote

sampsyo: "this requires generate buffer" -> this requires generating a buffer

content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+
+The programming interface is also easy-to-use, users only need to attach the following two lines of code to their schedule. Our compiler will automatically generate the buffer and implement the reuse logic.
```
sampsyo: replace the comma in "easy-to-use, users" with a semicolon, colon, or new sentence

content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+\begin{aligned}
+\text{reuse at axis 0}: &[s_0,w_1,\ldots,w_m]\\\\
+\text{reuse at axis 1}: &[s_0,s_1,\ldots,w_m]\\\\
+\cdots &\\\\
+\text{reuse at axis m}: &[s_0,s_1,\ldots,s_m]\\\\
+\end{aligned}
+$
+
+### Implementation
+The above description seems straightforward, but the actual implementation in MLIR is intractable. There are two main challenges here:
```
sampsyo: By "intractable," perhaps you mean "difficult"? The former usually means that it's not really possible, whereas it seems like you've actually done it, even if it's hard.

content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+          }
+        }
+      } {loop_name = "j"}
+    } {loop_name = "i", stage_name = "B"}
+    return %0 : memref<8x8xi32>
+  }
+}
+```
+
+The above steps can be extended to high-dimensional stencils, and we will give more experimental results in the following sections.
```
sampsyo, line 238: Overall, this section makes me wonder whether—in an ideal world—you might have preferred a different program representation (i.e., not the `affine` dialect)? I'm not entirely clear on the distinction, but there seem to be some challenges here that are fundamental and some that are a consequence of the program representation.

content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+
+<center>
+<img src="roofline_intro.png" alt="alt_text" title="image_tooltip" style="zoom:20%;">
+</center>
+<center>
+Fig. 5: Roofline Model
+</center>
+
+<!-- What we do to access the memory access / ops -->
+We add a profiling tool to evaluate the operational intensity by analyzing the MLIR code to count the number of memory access and arithmetic operations. Then a roofline model is generated to guide the optimization process. Specifically, we visit each operation and record the current loop nest. When an arithmetic operation or load/store operation is visited, the number of trip counts in the current loop is added.
```
sampsyo: I'm a tiny bit confused about the high-level idea here: it's called "profiling," but profilers are usually dynamic (i.e., they run a program and measure what it does at run time). This sounds like a static analysis—is that right? If so, is there any worry about precision? Or are all loops statically bounded anyway, so the number of accesses, etc., are always perfectly statically analyzable? If that's the case, it seems like this is worth a mention.

content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+
+### MLIR-Based HeteroCL Compilation Flow
+MLIR-based HeteroCL supports two backends for now: a CPU backend through LLVM dialect, and an FPGA backend through Vivado HLS. A HeteroCL program first generates HCL dialect IR, with HCL operations to represent hardware customizations. Then, a transformation pass implements hardware customizations and erases HCL operations, and the IR is converted to affine dialect, as shown in Fig. 6.
+
+<center>
+<img src="hcl-flow.png" alt="alt_text" title="image_tooltip" style="zoom:20%;">
+</center>
+<center>
+Fig. 6: MLIR-based HeteroCL end-to-end compilation flow
+</center>
```
sampsyo, line 415: This figure refers to something called "SCoP" and "PoCC", which are not defined anywhere, AFAICT?

content/blog/2022-05-19-hcl-mlir/index.md
```diff
@@ -0,0 +1,514 @@
...
+* `blur`: a simple three-point 2D blur filter with $\\{\mathbf{a}^{(i)}\\}^{3}_{i=1}=\\{(0,0),(0,1),(0,2)\\}$.
+* `blur+LB`: blur filter with line buffer of size $3$.
+* `conv2d-nchw`: baseline conv2d implementation with conventional $(n,c,h,w)$ dimensions, and the kernel size is $3\times 3$.
+* `conv2d-nchw+LB/WB`: create line buffer (LB) and window buffer (WB) for conv2d kernel.
+* `5point`: the five-point stencil kernel as mentioned in the previous section (see Fig. 3).
+* `5point+LB+WB`: create line buffer (LB) and window buffer (WB) for 5-point stencil kernel.
+* `diag3d`: three-point stencil in 3D with $\\{\mathbf{a}^{(i)}\\}^{3}_{i=1}=\\{(0,0,0),(1,1,1),(2,2,2)\\}$.
+* `diag3d+xyz`: create three hyper-cubic reuse buffers for this three-point stencil.
+
+Kernel | Latency | Speedup | II | DSP | BRAM | LUT | FF
```
sampsyo, line 432: Is the latency here from the HLS estimates, or from real FPGA executions? (It might also be nice to report, in your experimental setup paragraph above, which FPGA you are targeting.)

sampsyo (changes requested):
This is all impressively complete—nice work, everybody. Thanks for the clear and detailed explanations of all of these moving pieces. I only have a few specific inline comments—feel free to ask for clarification. But it's quite clear that this work is headed for publication, huh?

Congratulations on finishing up 6120! It has been a pleasure.

content/blog/2022-05-19-hcl-mlir/index.md
chhzh123: Right, I think the reason why affine dialect decouples the affine map from operands is that it is easier to do affine transformation. Some facilities for polyhedral compilation should exist in MLIR, but unfortunately, we have not found a good one to incorporate, so we can only write the analysis part ourselves, which is awkward.

chhzh123: Right, I think the reason why affine dialect decouples the affine map from operands is that it is easier to do affine transformation. Some facilities for polyhedral compilation should exist in MLIR, but unfortunately, we have not found a good one to incorporate, so we can only write the analysis part ourselves, which is awkward.

chhzh123: TVM can represent straight one-line code in their IR like `B[i] = A[i] + A[i+1] + A[i+2]`. This may simplify some analysis, but complicate other transformations. Different representations seem to have tradeoff here.

content/blog/2022-05-19-hcl-mlir/index.md
zzzDavid: Hi Adrian, you are right. The PoCC branch is for a different project going on in HeteroCL, which is to verify the program is still correct after adding customizations. Since this is not relevant to the memory optimization, I have removed that branch and updated the figure.

content/blog/2022-05-19-hcl-mlir/index.md
chhzh123: Yes, there are two parts in the roofline model:
1. The number of memory accesses and the number of arithmetic operations (FLOPs) can be statically calculated from the source MLIR code. Right now we only support constant loop bounds, but if complex control flow is involved, we probably need to do some instrumentation to dynamically count the numbers.
2. The performance in the y-axis needs to run the HLS and obtain the latency, thus cannot be statically determined.

sampsyo: Got it. To be clear, however, I would still call that HLS estimate "static"… in that it doesn't involve _actually running_ the program, which would be necessary if there were input-dependent loop trip counts, for example.

sampsyo (approved):
Looking good, y'all! Publishing now.

PR #344, opened 2022-05-23
Project Report: Vectorization for Bril

### LINKED ISSUE #312: Project Proposal: Vectorization
**What will you do?**

I want to add a vector instruction set to the Bril language, and implement a vectorization pass that finds opportunities to vectorize Bril IR code into a parallel form. The vector width would be fixed in advance. In particular, I plan to write a simple vectorizer pass that detects whether operations inside a program can be vectorized together. 

Depending on how much the scope of this project is, if the scope is too small, I would try implementing the polyhedral model for loop optimization, to try to achieve further vectorization, within loops. If this ends up being the case, I will read over lecture nodes/slides relating to the polyhedral model. I need to catch up on aspects of linear programming and integer programming. I would aim to reproduce a basic optimization like Martin Griebl documented as the space-time transformation in https://www.infosun.fim.uni-passau.de/cl/loopo/doc/loopo_doc/node3.html.
In particular, I would not try to reproduce the full power of using the polyhedral model for optimization, but would limit my scope.

These notes seem promising:
https://polyhedral.info/ - from an older version of CS 6120
https://homes.luddy.indiana.edu/achauhan/Teaching/B629/2010-Fall/StudentPresns/PolyhedralModelOverview.pdf
https://events.csa.iisc.ac.in/summerschool2013/slides/automatic-parallelization-introduction-polyhedral-models.pdf

**How will you do it?**

I plan to extend the Bril interpreter to have a vector instruction, with vector operations like vector addition and vector multiplication, concatenation of vectors, as well as loads and stores into and from vectors. This would involve changing the Bril Json Parser and also involving changing the typescript interpreter. 

I will be working with the Bril memory extension, to allow for interaction with arrays and pointers in the Bril language. In order to check for vectorization opportunities, I will implement alias analysis for Bril, to see if vectorization can be done. I would start by following the alias analysis defined in lecture, and try expanding to other forms like Steensgard's alias analysis algorithm. 

I will try to detect whether a loop can be unrolled, and unroll the loop as much as possible to enable further vectorization inside the loop. If not possible to unroll the whole loop, I will unroll the loop by some number of fixed iterations, say 4 iterations at a time.

If I move on to working on the polytope method, I would also have to add utilities to identify for loops and their format, and also represent the polyhedron and allow for transformations on the polyhedron. 

I will either create my own benchmarks or choose some common benchmarks, and translate these benchmarks into bril. I will look for benchmarks that focus on loops, with arrays inside the loop.

Since I am working at the Bril level, I am planning to count the number of vector instructions generated in each pass. The more vector instructions emitted, the more successful the pass.

I might also come up with a cost model that evaluates the cost of each vector instruction and regular instruction. This way, the optimized and unoptimized programs can be compared based on the result of the cost model. 

**Team members:**
Myself, although I am open to having other team members as well.

JonathanDLTran:
Closes [#312 ](https://github.com/sampsyo/cs6120/issues/312)


content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Vector Language
+
+The design of the vector language was based off resources from several students who had previously taken CS 6120, as well as the LLVM vector language. The students' languages are located at [vril](https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/vril-vector-bril/), and [vector-instruction-support](https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/interpreter-vector-support/). The LLVM vector language can be found be found at [llvm-vector](https://llvm.org/docs/LangRef.html#vector-operations).
```
sampsyo: FWIW, this paragraph might be a tiny bit more readable if you used the names of the authors rather than the URL slugs for the blog posts?

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+To interact between vector registers and normal registers. one can load into and store from vectors. For instance, the `vecload` instruction places one integer of data into a vector, at the specified index in the vector. Likewise, `vecstore` extracts one integer of data from a vector into a Bril pseudo-register. 
+
+The vector language is simulated in the Bril interpreter. I implemented the vector registers as variables storing array of width 4. Each vector operation manipulates the array appropriately; for instance, `veczero` instantiates all the cells of the array to have value 0, `vecload v i d` loads data `d` into the array representing vector `v` at index `i` in the array, and `vecadd v1 v2` does element-wise addition on the arrays representing vectors `v1` and `v2`. The implementation of this interpreter is located in the files `bril-vc` and `brili-vc`. I also augment the `print` instruction to print out all elements in a vector, for easy debugging.
```
sampsyo: Very nice! I like this design for a vector extension quite a bit. Would you be interested in contributing some aspects of it to the Bril monorepo so other projects could conceivably build on it in the future?

The one design aspect that I could imagine changing is adding "whole-vector" loads and stores that interact with the memory extension. That is, `v: vec = vecld p` would get 4 integers from memory starting at pointer `p` and put it into vector register `v`. (One could imagine renaming the existing `vecload` and `vecstore` to evoke "insert" and "extract".) Having vectorized loads/stores would bring us even closer to modeling the performance advantage of real vector ISAs.

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Vectorization
+
+Vectorization is broken into several stages. The first stage is to preprocess the code, which enables the vectorization algorithms to have greater opportunities to create vectors. The preprocessing stage is comprised of several passes that attempt to create larger basic blocks with more consecutive instructions that can be vectorized, which I call "vectorizable" instructions. 
+
+The second stage is to create runs of vectorizable instructions. A run of instructions is a sequence of Bril instructions that are all have the same opcode and are independent. A vectorizable instruction is independent if its arguments to do not depend on any other destination variables from the same run of instructions. The definitions of independence and same opcode come from Larsen and Amarasinghe's work on superword level parallelism. [^1] For example, the sequence of Bril instructions
```
sampsyo: I recommend deleting the space before the footnote mark ` [^1]`; it will look a little nicer in the rendered output.

sampsyo: Also, probably add a `:` at the end of this paragraph, before listing some example code.

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+z: int = add d d;
+```
+
+would not be a valid run of vectorizable instructions because `w` is defined and then used in the same basic block.
+
+Finally, the third stage is to run the vectorization algorithms on the runs of vectorizable instructions, and create vectors, which then replace code in the basic block. Vectors are only created when each run has two or more instructions in it.
+
+### Preprocessing
+
+Preprocessing is applied to the Bril code to allow for further optimization potential. Assuming no memory operations are present, the first preprocessing optimizations applied are simple dead code elimination (DCE) and loop invariant code motion (LICM). Removing obviously dead instructions will help simplify the code. As a side note, the condition for non-present memory conditions is because my implementations do not handle memory operations for DCE or LICM. Also, local value numbering (LVN) would have been used, but my version of LVN had a bug, causing me to exclude using LVN.
```
sampsyo: Maybe simplify "non-present memory conditions" to "the absence of memory accesses"?

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+Next, loops in the code are fully unrolled, if possible. Unrolling only applies to loops that have one iteration variable that is increasing by 1 or decreasing by 1. Further, the loop header must have exactly one branch condition, depending on the iteration variable. The loop also must have only one exit, from the branch condition in the header. By unrolling the code, loops get expanded into long sequences of straight-line code without branches or jumps. Having straight line code allows for long sequences of instructions that can be identified as vectorizable, and the possibility that a vector may be used several times.
+
+After unrolling, stores are then moved as late in the basic block as possible. Store movement depends on alias analysis, because a store can only move past another store or another load, assuming no aliasing. Moving stores later in the basic block allows for a greater opportunity of finding vectors, because vectors have to end when a store is identified. In the same vein, `const` instructions and `id` instructions are moved as early as possible, while `print` instructions are moved to be as far down in a basic block as possible.
```
sampsyo: Since it will inevitably come up as a question, maybe you want to head off the reader asking "but what alias analysis did you implement?" by giving a quick preview or saying that you'll address this later?

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+i: int = const 0;
+w: int = vecstore sum i;
+i: int = add i one;
+x: int = vecstore sum i;
+i: int = add i one;
+y: int = vecstore sum i;
+i: int = add i one;
+z: int = vecstore sum i;
+```
```
sampsyo, line 110: Very cool! Looking at this code makes me wonder whether *downstream* analyses (e.g., "opportunistic vectorization") wouldn't have been a little easier if there were a single instruction pack/unpack entire vectors from/to scalar registers…

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+n: int = vecstore sum2 i17;
+i18: int = add i17 one;
+o: int = vecstore sum2 i18;
+i19: int = add i18 one;
+p: int = vecstore sum2 i19;
+```
+
+where only three new vectors are created, rather than four, because vector `v2` can be reused in its entirety. Note that constants like `1` are not regenerated.
+
+To deal with partial matches in opportunistic vectorization, I use the `vecmove` instruction. Consider the vector `v` associated with the quadruple (`a`, `b`, `c`, `a`). If we are required to build the vector (`a`, `b`, `d`, `a`), which differs at one location from vector `v`'s quadruple, we can reuse the vector `v` by copying it to a new vector register `v1`, and then loading `d` into the second index of `v1`. Partial matches are only considered when there are at least two matches, to lessen the number of `vecload` instructions required. Overall, using `vecmove` allows for greater reuse of vectors.
```
sampsyo, line 194: This is a super cool tweak on the basic idea!

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Evaluation
+
+I evaluated the vectorization algorithms for correctness on my own tests, as well as the Bril Benchmarks. I restricted the benchmarks to only benchmarks using the Bril core language, as well as the Bril memory extension. The Bril speculative and Bril floating point extensions are not supported in my vectorization scheme. 
+
+Using Brench, I verified that my passes operated correctly on all the passes, compared to when my passes were not run. I also evaluated the vectorization algorithms to see how many vectors were created at run time. To do so, I instrumented the `brili-vc` interpreter to count the number of `veczero` calls that were created. The results are shown in the graph below:
```
sampsyo: "on all the passes" -> "on all the benchmarks?"

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+I also track the number of vectors created statically. These results are perhaps more revealing of the ability of the vectorization pass to find opportunities to create vector instructions. 
+
+<p align="center">
+<img src="static_vector_creation.png">
+title="Vectors Statically Created" style="zoom:25%;"
+</p>
+
+Most of the benchmarks have no vector instructions created statically, but for those that do, between 1 and 4 vector instructions are created.
```
sampsyo: It would be interesting to hear any thoughts you have about which parts of your implementation were the most important for finding these vectorizable instructions. Was inlining important, for example?

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+Next, the preprocessing passes are not as strong as I intended them to be. For instance, the unrolling pass is very simple in how it detects whether a loop can be unrolled. In fact, the unrolling pass never actually runs, because the only benchmark it could have run on, `mat-mul`, did not hardcode the dimensions of the matrix, instead passing it in as a parameter. As a result, I would need to improve the unrolling pass to make it detect a wider range of fully-unrollable loops. In addition, the alias analysis pass was designed to be very basic, generally only being able to disambiguate between two different arrays. A major weakness is that the alias analysis pass only assigns one address to an entire array, failing to disambiguate between different locations within an array. This means that some instances of store movements cannot occur successfully, because the alias analysis will report two addresses in an array can alias, when in fact the two addresses are at different memory locations. Hence, it would probably be a good idea to augment the alias analysis to model each cell in the array.
```
sampsyo: > As a result, I would need to improve the unrolling pass to make it detect a wider range of fully-unrollable loops.

Another option would be to implement partial unrolling…

sampsyo: I think this is emblematic of a broader lesson about auto-vectorization in compilers in general. It depends so heavily on the success of "standard" compiler analyses—*especially* alias analysis! Really neat to see this observation play out in practice here.

content/blog/2022-05-23-vectorization/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+Additionally, the algorithms that I implemented in this project were very simple. There are more complex ways to allow for further vector reuse across basic blocks, and across entire functions. For instance, Mendis and Amarasinghe's goSLP project uses integer linear programming (ILP) to encode conditions that each vector pack must satisfy in a valid vectorization scheme. [^2] After using an ILP solver, an optimal solution can be found for creating vector packs across an entire function. goSLP also minimizes the number of vector shuffles required, by using a dynamic programming approach. While I never tried any ideas from the goSLP paper, I did try brute force techniques to enumerate every possible permutation of instructions in a vector pack, considering packs of size two, three and four. This attempt failed to scale to five vector packs, and even on four vector packs, produced too many possible vector packs that could be valid vectorization outcomes. This made me realize that I had to be more clever with my brute force technique to create more tractable solutions. I also tried enumerating combinations, rather than permutations, as well, and the results were fairly poor.
```
sampsyo: Another good broad lesson about vectorization! Dealing with the exponential space of data-movement strategies _between_ vectors is both important and hard…

sampsyo (changes requested):
Really, really nice work, Jonathan—this is impressively complete and the blog post describes the project well. Even if the opportunistic approach only finds vectorization opportunities in a few benchmarks, that's still a demonstration of it working at all, which is big!

If you're interested, I would love to merge at least the basic support for vectors into the Bril monorepo so others can build on this extension, which seems quite clean. Let me know if you'd ever be interested in discussing this further!

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: I fixed this by using the authors' names.

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: I would be interested in contributing to the Bril monorepo! I agree that the "whole-vector" loads would be nice, and that the renaming would be more consistent.

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: Fixed!

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: I've added a paragraph describing the general mechanism for alias analysis.

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: Fixed!

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: I added more commentary at this point to describe inlining and loop unrolling, and their impacts.

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: A single unpack would have been very useful in this circumstance, at least from the view of writing to code to generate all the `vecstore` operations. I stuck with multiple unpack operations because I was under the impression, from the LLVM vector instructions, that multiple extracts were needed, and I assumed that a single unpack would be generated at the instruction selection stage later in the compilation pipeline.

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: Fixed!

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: I changed this to also mention partial unrolling. I had actually started with a basic form of partial unrolling, but I kept in a lot of branches, and because of the branches, I figured it would not be too useful.

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: This was a good lesson for me to learn from, especially from seeing how the enumeration space became too large to search even with relatively few instructions.

content/blog/2022-05-23-vectorization/index.md
JonathanDLTran: I've added in the colon at the end of the paragraph.

JonathanDLTran:
Thank you for the feedback! I incorporated the changes into a series of new commits.

content/blog/2022-05-23-vectorization/index.md
sampsyo: Yeah, that definitely makes sense for this version. Bril doesn't really have a way to write to multiple registers at once anyway, so making this work would be pretty hard.

sampsyo (approved):
Wahoo; looking good! Again, extremely nice work on this project, Jonathan.

PR #362, opened 2023-08-22
Add Zak as leader for Compcert Discussion

zachary-kent:


sampsyo:
Yay!

stephenverderame:
Hey, Ali and I were also interested in doing this paper as well. Would it be alright if we lead it with you?

PR #364, opened 2023-08-22
Update content.toml

jdroob:
Claimed MLIR paper


sampsyo:
The last slot!! Wahoo!

PR #365, opened 2023-08-22
add name to Alive paper

ak-2485:


sampsyo:
Hello! Looks like @bcarlet & @surendraGh got here first. Can y'all please check in about whether you want to team up?

PR #373, opened 2023-08-22
claiming the synthesis-aided compiler paper

yxd97:
This is Yixiao Du; I'd like to lead the discussion of _Chlorophyll: Synthesis-Aided Compiler for Low-Power Spatial Architectures_.


sampsyo:
Wonderful; thanks!!

PR #383, opened 2023-08-28
Added Alice to the Dynamic Languages Paper

AliceSzzze:
Checked with Benny, still waiting to hear from Collin but I hope he is fine with this.


sampsyo:
It's a plan!

PR #423, opened 2023-12-11
Jiahan CS 6120 Final Project Blog

### LINKED ISSUE #410: Project Proposal: Support LLVM GlobalISel for RISC-V Vector Extension
**What will you do?**

Support LLVM [Global Instruction Selection](https://llvm.org/docs/GlobalISel/index.html) for a subset of [RISC-V Vector Extension](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc), namely [vector integer add and sub](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#11-vector-integer-arithmetic-instructions) and [basic vector load/store](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#7-vector-loads-and-stores).

**How will you do it?**
1. For scalars, Apple has been working on it to support GlobalISel for AArch64 for years: [post](https://discourse.llvm.org/t/enabling-globalisel-for-apple-aarch64-platforms/63953) and [source code](https://github.com/llvm/llvm-project/blob/3e32b809001cf3a9d6f9cea4030fb9fe66f083f2/llvm/lib/Target/AArch64/GISel/AArch64GlobalISelUtils.h#L4).
2. For vectors, although GlobalISel does not fully support them, [SelectionDAG](https://llvm.org/doxygen/SelectionDAG_8cpp_source.html) does.

Therefore, I will leverage these two, combined with the semantic meaning of [LLVM Generic Opcodes](https://llvm.org/docs/GlobalISel/GenericOpcode.html) and [RISC-V Vector Extension Spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#risc-v-v-vector-extension) to support the aforementioned basic RISC-V Vector Extension instructions.

For the actual implementation, I will support the full pipeline:
- [IR Translation](https://llvm.org/docs/GlobalISel/IRTranslator.html);
- [Legalization](https://llvm.org/docs/GlobalISel/Legalizer.html);
- [Register Bank Selection](https://llvm.org/docs/GlobalISel/RegBankSelect.html); and
- [Instruction Selection](https://llvm.org/docs/GlobalISel/InstructionSelect.html).

The beginning of the pipeline should be straightfoward and interesting things will happen downstream.

I will use LLVM IR as inputs and inspect the RISCV-V Vector Extension assembly outputs.

First, I need to make sure the assembly is doing the correct thing as expected.

Then, I will compare my GlobalISel outputs with the outputs from SelectionDAG, which will be used as the ground truth. If these two are equal, that means I have an "optimal" GlobalISel implementation.

**Team members:**

I will do the implementation job individually, but will ask @michaelmaitland for guidance if necessary.

jiahanxie353:
Closes #410 

In this final project, I worked with @michaelmaitland on supporting LLVM GlobalISel for the RISC-V vector extension on part of the vectorized ALU operations.


content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+[[extra.authors]]
+name = "Michael Maitland"
++++
+
+# Introduction
+
+The open [RISC-V instruction set architecture (ISA)](https://riscv.org/technical/specifications/) has an interesting extension, [the RISC-V "V" Vector Extension](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc). The unique part of RISC-V Vector Extension is that its vector registers have flexible widths, VLEN, which makes programming in RISC-V Vector Extension agnostic to the vector register sizes. This feature really distinguishes the RISC-V Vector Extension from the traditional SIMD extensions, such as [x86 SSE](https://en.wikipedia.org/wiki/Streaming_SIMD_Extensions) (with a fixed size 128-bit vector length)/[AVX](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions) (256-bit), and [Arm NEON](https://developer.arm.com/Architectures/Neon) (128-bit). The increasing vector lengths can pose a challenge to the traditional SIMD extensions as they have to address compatability and support all existing fixed size vector lengths in their ISAs. On the contrary, with the vector lengths agnostic principles, binary code generated by RISC-V assembly is automatically portable between different CPUs.
```
michaelmaitland: RISC-V folks more commonly refer to VL as the unique part.

> The vl register holds an unsigned integer specifying the number of elements to be updated with results from a vector instruction

This is in contrast to many SIMD approaches where the instruction pneumonic and different sizes of register impose a fixed number of elements processed. For example, the [ARM Neon VADD](https://developer.arm.com/documentation/den0018/a/NEON-Intrinsics-Reference/Arithmetic/VADD).

On a given hardware implementation, the widths of registers in RISCV are flexible due to LMUL. This is also unique to RISCV. 

The VLEN however, is fixed for that hardware implementation.

michaelmaitland: If a processor implements RISC-V V extension, the vector code is portable to other CPU that support the V extension. However, if a processor implements say the ARM Neon extension, then the NEON code is portable to other ARM processors that support the Neon extension.

The real power of vector length agnostic is when it comes to loops. Take the following example:

```
for (i = 0; i < N; i++)
  A[i] = B[i] + C[i]
```

On RISC-V, we can generate the following (pseudo) code
```
ph:
  i = 0
body:
  vsetvli (N-i) // set the VL as N-i, or VLMAX if N-i is too large for the hardware to process
  vadd.vv a_ptr, b_ptr, c_ptr
  i += VL // Increment I by the number of elements processed in this iteration
  branch_if_done exit, body // decide whether to exit loop
exit:
  ret
```

On a SIMD architecture, it is more complicated. You need to have two versions of the loop: the vector loop, and a scalar remainder loop. If N-i is smaller than the number of elements processed in the vector loop, then the scalar remainder loop needs to be executed. What is the best size for the number of elements to be processed in the vector loop? It depends on whether N will often be large or small. You may not know this ahead of time.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+The most important parameters are undoubtly `SEW`, `VLEN`, `VL`, and `LMUL`; and one of the most interesting and powerful instructions is `vset{i}vl{i}`.
+
+Let's begin with the crucial parameters:
+- `SEW`: Selected Element Width (in bits), set dynamically by the programmers. It sets the width/length of a single element in a vector element. Each vector element can compose multiple single element.
+- `VLEN`: Vector register LENgth (in bits). The number of bits in a single vector register. It is hardware dependent. 
+- `VL`: Vector element Length (in bits) that the programmers actually deal with, which can be treated as the vector operation building blocks. It defines how many elements the vector operations will execute.
+- `LMUL`: The vector Length MULtiplier. It is used for grouping vector registers. It is a power of 2 and it ranges from 1/8 to 8. For instance, when `LMUL=8`, only `v0`, `v8`, `v16`, and `v24` indices are allowed to used, as for example, group `v8` encodes 8 vector elments `v8v9`...`v15`. Note it can also be fraction numbers because sometimes we want to use only parts of the vector registers.
```
michaelmaitland: Technically the ABI imposes the fact that `v0`, `v8`, `v16` can be used, not the RISC-V Vector spec. You could have used `v1, v9, v17`. The reason the ABI does it the way it does is so that v17-v31 can be used in the caller/callee save scheme. The important part here is that on LMUL 8, the register `vN` acts as a grouping of `vN, vN+1, ..., vN+7` registers. That forces us to partition what registers can be passed and preserve the expected semantics of the instruction.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+  
+After introducing these basic and the most important parameters, there are still two paramters we will be dealing with, `AVL` and `VLMAX`:
+- `AVL`: Application Vector Length. The application specifies the total number of elements to be processed as a candidate for `VL`.
+- `VLMAX`: MAX Vector Length. Although programmers can dynamically select vector register lengths, they cannot be set to arbitrary numbers because they might be larger than the hardware constraints, which is imposed by `VLMAX`. Specifically, `VLMAX=(LMUL × VLEN) / SEW`. For example, if the length of a single basic elment is 32 bit, and this specific hardware supports `VLEN`=256 bit, there are 8 basic blocks in a vector element. And we want to group every `LMUL`=2 vector elements, `VLMAX`=4.  As a result, the actual value `VL` is the smaller value of the desired application vector length `AVL` and the vector length value constrained by the hardware `VLMAX`: `VL = min(AVL, VLMAX)`.
```
michaelmaitland: `8 basic blocks` might be confusing terminology since basic blocks usually refers to the sequence of instructions that have continuous control flow. Maybe it would be better to say `8 elements can fit in vector register`

michaelmaitland: You say VLMAX=4 here but I think VLMAX=M2 * 256 / 32 = 16 in this case.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+Implementation-wise, the [`select` function](https://llvm.org/doxygen/classllvm_1_1InstructionSelector.html#a50058a922d4f75ed765c34742c5066c6) is invoked, which in turns call [the corresponding RISC-V `selectImpl` function](https://github.com/llvm/llvm-project/blob/d96f46dd20157be9c11e16d8bdd3ebf900df41fc/llvm/lib/Target/RISCV/GISel/RISCVInstructionSelector.cpp#L56). To achieve this final phase, there are essentially four steps to take. First is to identify the vectorized opcode/gMIR; then we create the lowered version of that gMIR using the vector instruction pseudos; and we need to erase the old instruction once the lowered version has been picked; finally we choose a register from the register bank. For this final phase in GlobalISel, LLVM TableGen might generate `selectImpl` and we can use it out-of-the-box; otherwise, we need to implement extra logics to customize the selection pass outlined above. Luckily, TableGen does pick up and we only need implement some helper functions to mesh everything together.
+
+[This patch](https://github.com/llvm/llvm-project/pull/74114) implements both Register Bank Select and Instruction Select passes. 
+
```
michaelmaitland, line 144: Maybe you can show what code gets generated by running all the way from IR to RISC-V ASM too? I'm sure there will be readers who want to know what code got generated without having to understand the Pseudo instruction internals of LLVM.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+define <vscale x 1 x i8> @vadd_vv_nxv1i8(<vscale x 1 x i8> %va, <vscale x 1 i8> %b) {
+  %vc = add <vscale 1 x i8> %va, %vb
+  ret <vscale x 1 x i8> %vc
+}
+```
+Here, we define a vector-vector add function that takes in two vector arguments, and returns the sum. For the background knowledge of LLVM [vector types](https://llvm.org/docs/LangRef.html#vector-type), it has both fix-length vectors and scalable vectors. We will be focusing on scalable vectors not only because it's more complex and flexible, it also fits into our abovementioned discussion regarding flexible-widths RISC-V vector registers. The form of scalable vectors in LLVM is as follows:
+```
+<vscale x VF x primitive data type>
+```
+where `VF` is the vectorization/grouping factor, it specifies how many vector elements are there in this group; a primitive data type; and `vscale` is a constant multiple of the specified number of vector elements, and `vscale` is unknown at compile time. For example, `<vscale x 4 x i32>` represents a scalable vector, which is a multiple of 4 32-bit integer values.
```
michaelmaitland: VF which stands for VectorizationFactor in the LLVM world has type `ElementCount`, which makes up `vscale x N` for scalable vectors. It would probably be more clear to use `<vscale x M x scalar data type>`. `M` is the minimum nuber of elements in the vector.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+1. `lowerReturn` enables functions to return scalable vector types. To be specific, scalable vector types in the input LLVM IR programs can be lowered into physical vector registers MIR based on the RISC-V vector calling conventions for return instructions. To implement this, we just write a function that translate to a return MIR if the return value is a scalable vector. It used to fall back to SelectionDAG before we implement this, now we can directly enable it in GlobalISel. [This patch](https://github.com/llvm/llvm-project/pull/71587) implements this functionality.
+2. `lowerFormalArguments` supports passing scalable vectors as arguments. We do it by implementing a function that takes in type information and returns true/false based on the type and other relevant information, such as if the target hardware supports vector instructions, and so forth. The existing argument-lowering function can take this as a helper function and enables passing scalable vectors as function arguments. [This patch](https://github.com/llvm/llvm-project/pull/70882) supports lowering formal arguments.
+3. `lowerCall` is used to lower the given call instruction according to the ABI.
```
michaelmaitland: This is not yet implemented for vectors on RISCV since it was not a blocker to do CodeGen for vector add. Maybe you clarify?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Legalizer
+After the IRTranslator translates the input program written in LLVM IR into:
+`%vc:_(<vscale x 1 x i8>) = G_ADD %va:_(<vscale x 1 x i8>), %vb:_(<vscale x 1 x i8>)`, we want to legalize all the types and operations. 
+
+Before explaining our work, let's define what does "legal" mean in LLVM GlobalISel. An instruction is legal if the target (RISC-V in our case) will later be able to select it to a target-specific MIR instruction.
+
+Prior to our support, the `G_ADD` operation is legal but the scalable vectors as Low Level Types are illegal in GlobalISel, so we focused on legalizing the LLTs. Just as a note, [Low Level Type (LLT)](https://llvm.org/doxygen/classllvm_1_1LLT.html) is intended to replace the usage of Extended Vector Type (EVT) in SelectionDAG. LLTs are specified in terms of whether they are scalar or vector, and their widths; but they don't differentiate between integer and floating values.  For example, `LLT s16` is simply defined as `LLT::scalar(16)`. For scalable vector types, if we want to define `nx4s32` (`<vscale x 4 x s32>`), it's nothing but `LLT::scalable_vector(4, s32)`.
```
michaelmaitland: `is legal but` -> `is legal for scalar types`

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+/           MF8    MF4     MF2     M1      M2      M4       M8
+SEW       
+i64         N/A    N/A     N/A     nxv1i64 nxv2i64 nxv4i64  nxv8i64
+i32         N/A    N/A     nxv1i32 nxv2i32 nxv4i32 nxv8i32  nxv16i32
+i16         N/A    nxv1i16 nxv2i16 nxv4i16 nxv8i16 nxv16i16 nxv32i16
+i8          nxv1i8 nxv2i8  nxv4i8  nxv8i8  nxv16i8 nxv32i8  nxv64i8
+```
+The complete chart can be found in [this `RISCV/RISCVRegisterInfo.td` file](https://github.com/llvm/llvm-project/blob/75d6795e420274346b14aca8b6bd49bfe6030eeb/llvm/lib/Target/RISCV/RISCVRegisterInfo.td). And note that `MF` stands for fractional `LMUL` and `M`s are integer `LMUL`s.
+
+Some values are `None` because currently RISC-V vectors assume `VLEN=64`. Take the combination (`MF8`, `i16`) as an example. If we were to write it in terms of LLVM scalable vectors, it would be `nx1/2i16` ((64 x 1/8) / 16 = 1/2), which is illegal. Now consider a legal (`LMUL`, `SEW`) combination: (`i32`, `M4`). Since `VLEN` = 64 and `SEW` = 32, there are 2 basic block elements in a single vector element. And since the grouping factor is 4, there are 2*4 = 8 multiples of elements, hence `nxv8i32 == <vscale x 8 x i32>`.
```
michaelmaitland: ditto about basic block being confusing here.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
...dd instruction to replace the generic add `G_ADD` instruction. Please note that I said "the corresponding MIR" because we will not be generating the actual RISC-V `vadd` or `vsetvli` in the current pass. It's because the process of instruction selection involves transforming code into target-specific MIR/machine instructions. Later down the pipeline, the `RISCVInsertVSETVLI` function, for example, will executed. Additionally, the `RISCVAsmPrinter` will translate MIR into MCInst at later stage, representing the final assembly language form. With that being said, what we actually want to get out of instruction selection pass is in this form: `%vc:vr = PseudoVADD_VV_MF8 %va, %vb, -1, 3 /* e8 */, 3 /* ta, ma */`, where `PseudoVADD_VV_MF8` is vector instruction pseudos for vector-vector add with `LMUL` = 1/8, the position where -1 stands is for the `VL` operand and -1 means `VLMAX`, the first 3 stands for `SEW` as log2(8) = 3, and the second 3 is the encoding for the policy tail agnostic and mask agnostic. RISC-V vector instruction pseudos in LLVM are essentially used for efficiently handling the complex, `vtype`-dependent behavior of vector instructions, such as in register allocation.
```
michaelmaitland: Important to say PseudoVADD_VV_MF8 is RISCV specific.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+Implementation-wise, the [`select` function](https://llvm.org/doxygen/classllvm_1_1InstructionSelector.html#a50058a922d4f75ed765c34742c5066c6) is invoked, which in turns call [the corresponding RISC-V `selectImpl` function](https://github.com/llvm/llvm-project/blob/d96f46dd20157be9c11e16d8bdd3ebf900df41fc/llvm/lib/Target/RISCV/GISel/RISCVInstructionSelector.cpp#L56). To achieve this final phase, there are essentially four steps to take. First is to identify the vectorized opcode/gMIR; then we create the lowered version of that gMIR using the vector instruction pseudos; and we need to erase the old instruction once the lowered version has been picked; finally we choose a register from the register bank. For this final phase in GlobalISel, LLVM TableGen might generate `selectImpl` and we can use it out-of-the-box; otherwise, we need to implement extra logics to customize the selection pass outlined above. Luckily, TableGen does pick up and we only need implement some helper functions to mesh everything together.
```
michaelmaitland: We don't actually choose a register from the register bank. We already have the register. In the case above, its the virtual register %vc. What we're really doing is constraining the type of this virtual register so that register allocation has enough information to convert the virtual register into a physical register.

michaelmaitland: LLVM TableGen always generates `selectImpl`. The "might" part has to do with whether there are TableGen patterns that exist to help us go from MachineInstr with GlobalIsel GenericOpcode -> SelectionDAG SDNode with SelectionDAG ISD opcode -> MachineInstr with RISC-V opcode. 

SelectionDAG implements the ISD -> MachineInstr with RISC-V opcode transformation. In this case G_ADD maps pretty well onto ISD::ADD. So GlobalISel defines an equivalence.

As we spoke before, it would be better if we didn't go through SelectionDAG at all, but this approach makes it easy for architectures to onboard onto GISel, and hopefully in the future we can remove going through SelectionDAG entirely.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+Learning the RISC-V vector extension was also a headache at the beginning because I had to figure out the difference between RISC-V vector extension with standard SIMD vector instructions. Learning the semantic meaning of `vsetvli`, differentiating the concepts of `ELEN`, `VLEN`, and how `SEW`, `LMUL`, `VLMAX` come into play was also confusing.
+
+# Were You Successful?
+
+This project is a success and we have become one of the first developers to support GlobalISel for the RISC-V vector extension.
```
michaelmaitland: And the first developers to support scalable vectors in GISEL for any target.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+name = "Michael Maitland"
++++
+
+# Introduction
+
+The open [RISC-V instruction set architecture (ISA)](https://riscv.org/technical/specifications/) has an interesting extension, [the RISC-V "V" Vector Extension](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc). The unique part of RISC-V Vector Extension is that its vector registers have flexible widths, VLEN, which makes programming in RISC-V Vector Extension agnostic to the vector register sizes. This feature really distinguishes the RISC-V Vector Extension from the traditional SIMD extensions, such as [x86 SSE](https://en.wikipedia.org/wiki/Streaming_SIMD_Extensions) (with a fixed size 128-bit vector length)/[AVX](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions) (256-bit), and [Arm NEON](https://developer.arm.com/Architectures/Neon) (128-bit). The increasing vector lengths can pose a challenge to the traditional SIMD extensions as they have to address compatability and support all existing fixed size vector lengths in their ISAs. On the contrary, with the vector lengths agnostic principles, binary code generated by RISC-V assembly is automatically portable between different CPUs.
+
+# What Was the Goal?
```
sampsyo: You don't need to include the bullet points [from the syllabus](https://www.cs.cornell.edu/courses/cs6120/2023fa/syllabus/#experience-report) as section headers in your blog post. This would flow a bit more naturally if you just transitioned directly into a recap of the goal as part of the intro. "In this project, our goal was to…"

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+The goal was to support LLVM Global Instruction Selection for RISC-V Vector Extensions on some ALU operations, such as `vadd`, `vsub`, `vand`, `vor`, and `vxor`.
```
sampsyo: Maybe it would be helpful to include a few words (not even a full sentence) here on what global isel is? I know you get into detail below, but a tiny overview here might help people understand the post's context.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+
+# A Primer on the RISC-V Vector Extension
+There are a lot of instructions and cool features in the [RISC-V "V" Vector Extension Spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc), I will only pick the most relevant and important parts to explain. 
+
+The most important parameters are undoubtly `SEW`, `VLEN`, `VL`, and `LMUL`; and one of the most interesting and powerful instructions is `vset{i}vl{i}`.
+
+Let's begin with the crucial parameters:
+- `SEW`: Selected Element Width (in bits), set dynamically by the programmers. It sets the width/length of a single element in a vector element. Each vector element can compose multiple single element.
```
sampsyo: I'm not sure what this means: "Each vector element can compose multiple single element." These sound like the same thing?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+- `VLEN`: Vector register LENgth (in bits). The number of bits in a single vector register. It is hardware dependent.
```
sampsyo: You might want to clarify "it is hardware dependent" here… earlier, you have said that this is flexible ("its vector registers have flexible widths, VLEN"), but here it seems to imply this is a hardware parameter. Maybe clarify which one is right?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+# A Primer on the RISC-V Vector Extension
+There are a lot of instructions and cool features in the [RISC-V "V" Vector Extension Spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc), I will only pick the most relevant and important parts to explain. 
+
+The most important parameters are undoubtly `SEW`, `VLEN`, `VL`, and `LMUL`; and one of the most interesting and powerful instructions is `vset{i}vl{i}`.
+
+Let's begin with the crucial parameters:
+- `SEW`: Selected Element Width (in bits), set dynamically by the programmers. It sets the width/length of a single element in a vector element. Each vector element can compose multiple single element.
+- `VLEN`: Vector register LENgth (in bits). The number of bits in a single vector register. It is hardware dependent. 
+- `VL`: Vector element Length (in bits) that the programmers actually deal with, which can be treated as the vector operation building blocks. It defines how many elements the vector operations will execute.
```
sampsyo: Can you clarify whether this must be less than or equal to VLEN?

Also, this bullet point seems to say that it is both in bits and in elements ("how many elements"). Can you clarify which is the actual quantity and which is implied?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
sampsyo: @michaelmaitland, just to clarify: one important constraint here (aside from the ABI) is that, when LMUL=8, you are not allowed to use `v1`, `v2`, `v9` at all, right? That is, the only "allowed" registers are at indices $\mathit{LMUL} \cdot k$ for natural numbers $k$? (Otherwise accessing these unaligned registers is an error?)

content/blog/2023-12-11-rvv-llvm-gisel/index.md
sampsyo: I'd write out "maximum" instead of "MAX", which looks like an acronym.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+- `AVL`: Application Vector Length. The application specifies the total number of elements to be processed as a candidate for `VL`.
+- `VLMAX`: MAX Vector Length. Although programmers can dynamically select vector register lengths, they cannot be set to arbitrary numbers because they might be larger than the hardware constraints, which is imposed by `VLMAX`. Specifically, `VLMAX=(LMUL × VLEN) / SEW`. For example, if the length of a single basic elment is 32 bit, and this specific hardware supports `VLEN`=256 bit, there are 8 basic blocks in a vector element. And we want to group every `LMUL`=2 vector elements, `VLMAX`=4.  As a result, the actual value `VL` is the smaller value of the desired application vector length `AVL` and the vector length value constrained by the hardware `VLMAX`: `VL = min(AVL, VLMAX)`.
+
+The visualization of vector registers is shown below: ![](vec-regs.png)
+
+One of the most interesting instruction is `vset{i}vl{i}`, which allow rapid configuration. The `vset{i}vl{i}` instructions set the `vtype` and `VL` based on arguments, and write the new value of `VL` into `rd`. Take `vsetvli` as an example, the spec of `vsetvli` is:
```
sampsyo: Delete "rapid"; I don't think it means anything here. Maybe just say "which configure the vector length"?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+2. [Legalize](https://llvm.org/docs/GlobalISel/Legalizer.html): This pass replaces unsupported operations with supported ones. So there will be no illegal instructions after this step.
+3. [Register Bank Select](https://llvm.org/docs/GlobalISel/RegBankSelect.html): This pass assigns the corresponding [Register Banks](https://llvm.org/docs/GlobalISel/GMIR.html#gmir-regbank) to the [generic virtual registers](https://llvm.org/docs/GlobalISel/GMIR.html#gmir-gvregs) in the gMIR. In essence, Register Bank is a framework for grouping and managing different sets of registers in a target architecture. 
+4. [Instruction Select](https://llvm.org/docs/GlobalISel/InstructionSelect.html): This is the final pass, it selects target MIR instructions based on the gMIR instructions. As mentioned above, the gMIR has been constrained enough that it becomes MIR at this point.
+
+A visualization of the pipeline is shown below: ![](pipeline.png)
+
+Please note that Combiner is not in the scope of this blog.
```
sampsyo: "Combiner" is not defined… do you want to link to it, or briefly explain what it is and why it is excluded?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+# What Did I Do?
```
sampsyo: Again, there is no need to use these bullet points as the section headings. Can you think of a section heading that would make more sense to a reader who stumbles across this blog post and wants to learn about instruction selection for RVV?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
sampsyo: This part is critical to understanding the LLVM IR input, and it could be made clearer. Namely, this is the question you need to answer for a reader: what's the difference between the types `<vscale x 4 x i32>` and plain old `<4 x i32>`? (The answer is that the latter is 4 i32s; the former is _any multiple of 4_ i32s, e.g., 0, 4, 8, or 12 i32s.)

content/blog/2023-12-11-rvv-llvm-gisel/index.md
sampsyo: Hmm… this paragraph says that scalable vectors as low-level types are illegal, but it does not yet say _why_ they are illegal. Is this a fundamental constraint? Is it a good idea? Could it be changed in the future? Or is this something that wouldn't make sense?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+Since we just brought up LLT, I'd like to introduce how to concretely associate the `VLEN`, `LMUL` things we discussed before with LLVM scalable vector types. Please note that as the date of this blog post, the LLVM community only has comprehensive support when `vscale == VLEN/64`, and that we only have partial support for `VLEN=32` using intrinsics when `vscale` is not used and there is no spill/reload of vectors. Supporting `VLEN=32` would require twice as many instruction selection patterns in `RISCVGenDAGISel.inc` which is already a very large table so the community hasn't done it. With that being said, I can present the chart of how RISC-V vector types are mapped to LLVM types for integers:
```
sampsyo: What is "vscale" as a variable? I am only familiar with `vscale` as a keyword in the type `<vscale x N x T>`. Maybe you mean `N` in such types?

I'm also just not sure what it means to have the constraint that `vscale == VLEN/64`. Where does `VLEN` come from in this equation? The hardware target?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+```
+LMUL
+/           MF8    MF4     MF2     M1      M2      M4       M8
+SEW       
+i64         N/A    N/A     N/A     nxv1i64 nxv2i64 nxv4i64  nxv8i64
+i32         N/A    N/A     nxv1i32 nxv2i32 nxv4i32 nxv8i32  nxv16i32
+i16         N/A    nxv1i16 nxv2i16 nxv4i16 nxv8i16 nxv16i16 nxv32i16
+i8          nxv1i8 nxv2i8  nxv4i8  nxv8i8  nxv16i8 nxv32i8  nxv64i8
+```
+The complete chart can be found in [this `RISCV/RISCVRegisterInfo.td` file](https://github.com/llvm/llvm-project/blob/75d6795e420274346b14aca8b6bd49bfe6030eeb/llvm/lib/Target/RISCV/RISCVRegisterInfo.td). And note that `MF` stands for fractional `LMUL` and `M`s are integer `LMUL`s.
```
sampsyo, line 114: Maybe you can clarify this table by just listing the LMUL itself? Like, instead of "MF8", just write "1/8". Instead of "M8", just write "8".

content/blog/2023-12-11-rvv-llvm-gisel/index.md
sampsyo: "currently RISC-V vectors assume `VLEN=64`": Who exactly assumes that? Surely not the ISA itself… maybe all of LLVM? Or just GlobalISel?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+The legal scalable vector types in the chart should all be legalized in the `legalize` pass. To implement this, we use the `getActionDefinitionsBuilder` function in [the `LegalizerInfo` class](https://llvm.org/doxygen/classllvm_1_1LegalizerInfo.html). In addition, we utilize [the `LegalityQuery` structure](https://llvm.org/doxygen/structllvm_1_1LegalityQuery.html) to query and filter the input LLTs. 
+
+It is noteworthy that some target hardwares only support `ELEN` = 32; on the other hand, `nxv1i64`, `nxv2i64`, `nxv4i64`, and `nxv8i64` require `hasVInstructionsI64` as we want to support max 64-bit vector elements in these cases. For example, in [Zve, Vector Extensions for Embedded Processors](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#182-zve-vector-extensions-for-embedded-processors), `Zvex32` does not provide support for `SEW` = 64. To this end, we need to have extra separate files when the target hardwares are different, which will be discussed further in the evaluation part.
```
sampsyo: What is ELEN? I don't think it's defined in this blog post.

sampsyo: Also, what is `hasVInstructionsI64`?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
...by using the [`ValueMapping` structure](https://llvm.org/doxygen/structllvm_1_1RegisterBankInfo_1_1ValueMapping.html). And we will create the mapping based on the size of the incoming operands. The rule here is we have 4 different sizes in the mapping for the 4 different integer `LMUL` register size cases. We pick based on the known minimum size of the register type, 64, 128, 256, and 512. We map any size less than 64 to 64 as well. And the reason being is `MF8`, `MF4`, and `MF2` have similar behaviors as `M1`; but `M2`, `M4`, and `M8` do not. Consider we have two instructions that use `vsetvli` to set `LMUL` to 1/2 (`MF2`) and 1 (`M1`) respectively. Assume we want to read a vector register in the first instruction. We can only read/extract the first half of that vector register, say `v1`, out of it since we have fractional `LMUL` `MF2`. In that sense, regardless of using `MF2` or `M1`, we are essentially using the whole registers so they all take one cycle to process even in the case of fractional `LMUL`s. However, if we set `LMUL` to 2, if we were reading `v1`, we end up reading a group of 2 registers: `{v1, v2}`. And this behavior is fundamentally different from the former case.
```
sampsyo: > Remarkably, we need a function to map operands values to register banks…

Why is this remarkable?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
sampsyo: "will executed" -> "will execute"

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+[This patch](https://github.com/llvm/llvm-project/pull/74114) implements both Register Bank Select and Instruction Select passes. 
+
+
+
+
+# What Were the Hardest Parts?
```
sampsyo: One more chance to retitle your sections to be more meaningful to the outside world.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+Definitely learning the whole LLVM and its GlobalISel infrastructure, and it was also hard to understand the vector length agnostic features/instructions in the RISC-V vector extension.
```
sampsyo: Can you make this a complete sentence? Again, imagine that the reader is someone who cares about the topic but does not care that this was a 6120 course project.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+LLVM is a huge project and it was hard to navigate at the beginning. They have the TableGen file, all the CodeGen GlobalISel source code, and the target-specific (RISC-V in our case) implementations. They also have all the regression tests infrastructure that I had to learn at the beggin. Besides, the GlobalISel infrastructure is also rather new and under active development as well. And since we are one of the earliest people start supporting RISC-V vector extension in LLVM GlobalISel, there had been some extensive research and discussion going on. And we didn't have that many existing work to refer to, so we had to start pretty much from scratch. My initial strategy was to look at AArch64 GlobalISel for their vectors because AArch64 is one of the pioneers in GlobalISel and their work is pretty mature.
```
sampsyo: "beggin"

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+LLVM is a huge project and it was hard to navigate at the beginning. They have the TableGen file, all the CodeGen GlobalISel source code, and the target-specific (RISC-V in our case) implementations. They also have all the regression tests infrastructure that I had to learn at the beggin. Besides, the GlobalISel infrastructure is also rather new and under active development as well. And since we are one of the earliest people start supporting RISC-V vector extension in LLVM GlobalISel, there had been some extensive research and discussion going on. And we didn't have that many existing work to refer to, so we had to start pretty much from scratch. My initial strategy was to look at AArch64 GlobalISel for their vectors because AArch64 is one of the pioneers in GlobalISel and their work is pretty mature. 
+
+Learning the RISC-V vector extension was also a headache at the beginning because I had to figure out the difference between RISC-V vector extension with standard SIMD vector instructions. Learning the semantic meaning of `vsetvli`, differentiating the concepts of `ELEN`, `VLEN`, and how `SEW`, `LMUL`, `VLMAX` come into play was also confusing.
+
+# Were You Successful?
```
sampsyo: Maybe this doesn't need a separate section heading from the above discussion?

sampsyo (changes requested):
Thanks for the very detailed and technical writeup! This sounds like a tremendous amount of work, and I'm glad you were able to make some progress. I have marked a few places where the post could be a tiny bit clearer, especially for outsiders who do not know that much about RVV.

One high-level ingredient I'd be interested in, if you feel like adding it (not a requirement): what *fundamentally* made this project interesting/difficult? What I mean is that you have identified several challenges that come from two sources:

1. RVV stuff is hard to think about
2. GlobalISel is somewhat complicated to extend

…but is there anything specific to _the intersection between those two things_? Like, is there anything about RVV in particular that makes global instruction selection harder? Or is this project approximately equal to the sum of its parts?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
michaelmaitland: That is correct. According to the spec:

```
When LMUL=2, the vector register group contains vector register v n and vector register v n+1, providing twice the vector length in bits. Instructions specifying an LMUL=2 vector register group with an odd-numbered vector register are reserved.

When LMUL=4, the vector register group contains four vector registers, and instructions specifying an LMUL=4 vector register group using vector register numbers that are not multiples of four are reserved.

When LMUL=8, the vector register group contains eight vector registers, and instructions specifying an LMUL=8 vector register group using register numbers that are not multiples of eight are reserved.
```

I think my statement above that `You could have used v1, v9, v17` is incorrect by this part of the spec.

sampsyo: Got it; thanks!

jiahanxie353:
Thanks for all the detailed and constructive feedback @sampsyo @michaelmaitland !

I tried my best to answer the questions. And please let me know if there's anything needed to be further supported/clarified!

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,173 @@
...
...IMD extensions with fixed vector lengths face challenges when dealing with the changing data sizes. They must maintain compatibility and support all existing fixed-size vector lengths in their instruction set architectures. This often leads to inefficiencies, especially in loop operations where the data size/loop stride may not align perfectly with the fixed vector size, necessitating additional scalar processing for the remaining elements. And the most suitable size for the number of elements to be processed in the vector loop is hard to decide ahead of time. In contrast, the RISC-V Vector Extension eliminate this concern with its vector length agnostic principle. Particularly in loop scenarios, the RISC-V's ability to adaptively handle varying data sizes stands out. For instance, in a simple loop adding two arrays, the RISC-V can dynamically adjust the vector length for each iteration by dynamically setting the vector length. This means it can process as many elements as possible in each pass, depending on the hardware capabilities and the remaining data. This adaptive approach really simplifies the code by eliminating the need for separate scalar loops for the leftover elements.
```
michaelmaitland: nit: `often` -> `may`

michaelmaitland: `the RISC-V` -> `the RISC-V program`

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+In this project, our goal was to support LLVM Global Instruction Selection (GlobalISel), a framework that operates on whole function for instruction selection, for the RISC-V Vector Extension on some ALU operations, such as `vadd`, `vsub`, `vand`, `vor`, and `vxor`. Apart from adding support for RISC-V vector types and operations for GlobalISel by going down GlobalISel's pipeline, it's a challenge to bridge the LLVM world (concepts like scalable vector) and the RISC-V world (concepts like vector length and vector register grouping factor) together. And we will showcase how we address the challenge in the following sections.
```
michaelmaitland: `function` -> `functions`

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+
+# A Primer on the RISC-V Vector Extension
+There are a lot of instructions and cool features in the [RISC-V "V" Vector Extension Spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc), I will only pick the most relevant and important parts to explain. 
+
+The most important parameters are undoubtly `SEW`, `VLEN`, `VL`, and `LMUL`; and one of the most interesting and powerful instructions is `vset{i}vl{i}`.
+
+Let's begin with the crucial parameters:
+- `SEW`: Selected Element Width (in bits), set dynamically by the programmers. It sets the width/length of a single element in a vector element/register. Each vector element can compose `VLEN`/`SEW` single elements.
```
michaelmaitland: ``Each vector element can compose `VLEN`/`SEW` single elements.`` -> ``Each vector can contain `VLEN`/`SEW elements.``

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,173 @@
...
+- `ELEN`: The maximum size in bits of a vector element that any operation can produce or consume.
```
michaelmaitland, line 29: `ELEN` is missing from list on line 25 above

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,173 @@
...
+- `VL`: Vector element Length (in bits) that the programmers actually deal with, which can be treated as the vector operation building blocks. It implies how many vector registers/elements the vector operations will execute. As will be discussed below, `VL = min(AVL, VLMAX)`, which means there's no fixed relationship between `VL` and `VLEN`. `VL` might be greater than, less than, or equal to `VLEN` depending on the value of `AVL`, `LMUL`, and `SEW`. 
+- `LMUL`: The vector Length MULtiplier. It is used for grouping vector registers. It is a power of 2 and it ranges from 1/8 to 8. For instance, when `LMUL=8`, the ABI imposes that only `v0`, `v8`, `v16`, and `v24` indices are allowed to used, as for example, group `v8` encodes 8 vector elments `v8v9`...`v15`. Note it can also be fraction numbers because sometimes we want to use only parts of the vector registers.
+  
+After introducing these basic and the most important parameters, there are still two paramters we will be dealing with, `AVL` and `VLMAX`:
+- `AVL`: Application Vector Length. The application specifies the total number of elements to be processed as a candidate for `VL`.
```
michaelmaitland, line 35: You introduce AVL above when you talk about VL. Maybe add a note above to `(see below about AVL)` so the reader knows you will explain AVL.

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+# A Primer on LLVM Global Instruction Selection
+LLVM Global Instruction Selection ([GlobalISel](https://llvm.org/docs/GlobalISel/index.html)) is a framework that provides a set of reusable passes and utilities for instruction selection — translation from LLVM IR to target-specific Machine IR (MIR). It is "global" in the sense that it operates on the whole function rather than a single basic block. 
+
+GlobalISel is intended to be a replacement for [SelectionDAG](https://llvm.org/docs/CodeGenerator.html#introduction-to-selectiondags) and [FastISel](https://llvm.org/doxygen/classllvm_1_1FastISel.html), to solve performance, granularity, and modularity problems. GlobalISel does not need to introduce a new dedicated IR as in SelectionDAG so GlobalISel can provide faster code generation; GlobalISel operates on a function, whereas SelectionDAG only considers a basic block, losing some global optimization opportunities; in addition, GlobalISel enables more code reuse for instruction selection for different targets.
```
michaelmaitland: `GlobalISel does not need to introduce a new dedicated IR as` -> `GlobalISel does not introduce a new dedicated IR since it works on the already existing and documented MIR, compared to SelectionDAG which has its own ISD representation.`

Also, are you sure that not having a separate IR is a reason for the faster code generation?

michaelmaitland: `in addition, GlobalISel enables more code reuse for instruction selection for different targets.`

Can you cite your source here?

content/blog/2023-12-11-rvv-llvm-gisel/index.md
```diff
@@ -0,0 +1,173 @@
...
+SEW       
+i64         N/A    N/A     N/A     nxv1i64 nxv2i64 nxv4i64  nxv8i64
+i32         N/A    N/A     nxv1i32 nxv2i32 nxv4i32 nxv8i32  nxv16i32
+i16         N/A    nxv1i16 nxv2i16 nxv4i16 nxv8i16 nxv16i16 nxv32i16
+i8          nxv1i8 nxv2i8  nxv4i8  nxv8i8  nxv16i8 nxv32i8  nxv64i8
+```
+The complete chart can be found in [this `RISCV/RISCVRegisterInfo.td` file](https://github.com/llvm/llvm-project/blob/75d6795e420274346b14aca8b6bd49bfe6030eeb/llvm/lib/Target/RISCV/RISCVRegisterInfo.td). And note that `MF` stands for fractional `LMUL` and `M`s are integer `LMUL`s.
+
+Some values are `None` because currently the LLVM community assumes the RISC-V vectors to have `VLEN=64`. Take the combination (`1/8`, `i16`) as an example. If we were to write it in terms of LLVM scalable vectors, it would be `nx1/2i16` ((64 x 1/8) / 16 = 1/2), which is illegal. Now consider a legal (`LMUL`, `SEW`) combination: (`i32`, `4`). Since `VLEN` = 64 and `SEW` = 32, there are 64/32 = 2 elements that can fit in a single vector element. And since the grouping factor is 4, there are 2*4 = 8 multiples of elements, hence `nxv8i32 == <vscale x 8 x i32>`.
```
michaelmaitland: I think some values are `N/A`, not `None`

michaelmaitland (commented):
A small round with minor verbiage changes to be more specific. I think this is the last round of changes I have before approval.

sampsyo:
This PR has "timed out"—please let me know if you'd like to reopen it.

michaelmaitland:
Happy to let this be merged if Jiahan wants. My last round of review was basically all nits.

sampsyo:
Any thoughts, @jiahanxie353?

jiahanxie353:
> Any thoughts, @jiahanxie353?

Could we reopen and merge it? I believe have addressed all reviewing questions:)

sampsyo:
Sure, I'll reopen it! It looks like some of @michaelmaitland's suggestions are still outstanding though? Let me know whenever you think we're ready. :)

jiahanxie353:
> Sure, I'll reopen it! It looks like some of @michaelmaitland's suggestions are still outstanding though? Let me know whenever you think we're ready. :)

Thanks, I just resolved the latest suggestions, thank you so much again for the review! @michaelmaitland 
And I rebased the latest `2023fa` commits, I think this PR is ready to be merged!

sampsyo:
Oh awesome; nice work! I hope all is well, @jiahanxie353!

jiahanxie353:
> Oh awesome; nice work! I hope all is well, @jiahanxie353!

Yeah absolutely! I had a great road trip around UT-AZ, settled down in Cali, and transitioned smoothly to my internship at Cerebras :)

Issue #394, opened 2023-10-19
Proposal: RISC-V Binary Translation to WebAssembly (WASM)


evanmwilliams:
**What will you do?**

We will design and implement a binary translator that converts RISC-V binaries into WebAssembly (WASM) binaries. This will enable RISC-V software to be executed in any environment with a WASM runtime, including web browsers, servers, and certain embedded systems.

**How will you do it?**

First, we'll need to familiarize ourselves with the RISC-V and WASM specifications. The three of us are quite familiar with RISC-V already, but we are less familiar with WASM. We'll pick a suitable subset of RISC-V instructions that we can then translate into WASM. The initial subset will just ensure we can get out a proof of concept. We'll extend the subset as time allows to be more ambitious. 

We'll need to develop a parser for RISC-V binaries. We've written interpreters for RISC-V before in other courses, so this should not be too involved. It might take a bit more work to design good data structures to store RISC-V instructions and their operands. Then, we'll create a mapping schema to transform each RISC-V instruction (or a series of instructions) into its corresponding Wasm representation. Some mappings may be direct, while others might require innovative solutions due to architectural differences. 

As a more lofty objective, we'll also try to resolve some of the differences between the memory models of the two ISAs. We'll need to emulate RISC-V's memory model within Wasm's linear memory. To deal with system calls, we'll need to implement shims or stubs to emulate or handle RISC-V system calls within the Wasm environment. Our mapped instructions can then be used to produce WASM binaries. 

If time permits, we will explore optimizations to improve the performance or reduce the size of the translated WASM binaries. Some of these optimizations can be more classical compilers techniques like dead code elimination or loop unrolling. Other more tailored optimizations might include profiling for hotspot optimization and tailoring optimizations for specific WASM runtimes. 

As a sidenote, we're planning on doing all of this in Rust. Two of us are pretty experienced with Rust already, and the third is eager to learn. Seems fitting, given some of the conversations we've had in class recently! 

**How will you empirically measure success?**

[...]

PR #424, opened 2023-12-11
[Blog] AsyncBril

evanmwilliams:
Final project for CS 6120 with @emwangs and @he-andy


evanmwilliams:
Here's the link to our actual code: https://github.com/sampsyo/bril/pull/305

https://github.com/he-andy/bril

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+bio="""
+Andy He is a junior studying CS and Math. He is particularly interested in systems and compilers development. 
+
+Emily Wang is an M.Eng student studying CS. She is very interested in systems, architecture, and compilers. 
+
+Evan Williams is an M.Eng student studying CS. He is broadly interested in distributed systems and compilers. 
+"""
++++
+# AsyncBril: Enhancing Bril with Asynchronous Programming and Thread-Style Features
+By Andy He, Emily Wang, and Evan Williams
```
sampsyo, line 19: Your names display at the top of the post when rendered, so no need to duplicate them here.

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Overview and Motivation 
+
+In the final lecture of CS 6120, we talked about concurrency and parallelism and 
+the challenges they raise for compiler developers in the modern era of 
+computing. In particular, we examined how the problem of semantics is 
+inherently embedded in shared-memory multithreading and came across a key idea 
+by Boehm: threads cannot be implemented as a library.
```
sampsyo, line 27: Maybe it would be nice to make this a link to the paper?

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+easily extensible nature makes it ideal for experimentation. We thought that 
+future CS 6120 students would benefit from being able to use our library to 
+implement their own synchronization primitives and other asynchronous 
+programming features using our extension. 
+
+
+## Design and Implementation 
+
+In this projet, we added two main features to Bril: promises and atomic 
+primitive types (i.e. `atomicint`). Promises are a fundamental
```
sampsyo, line 44: I'd recommend always using a comma after "e.g." and "i.e.". https://capra.cs.cornell.edu/styleguide/#egie

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+The heap is thread-unsafe, allowing multiple threads to concurrently read/write 
+to the heap without checking. This is done using Rust `unsafe` blocks. 
+Alternatively, we could've used a `Rw<T>` to lock/unlock the heap when threads 
+were modifying it, but we ultimately decided that it would be more performant 
+to not implement it this way and provide support for synchronization primitives,
+allowing for more flexibility. 
+
+Although the interpreter's heap is not thread safe, calls to `alloc` are still 
+made atomically, so no thread can accidentally be assigned the same index 
+in `state.heap.memory` when `alloc` is called simultaneously between threads.
```
sampsyo, line 70: I think it makes sense to assume an audience that knows about Bril but not about the Rust implementation… so your reader probably doesn't know what `state.heap.memory` is. Maybe use a direct description of what this is (e.g., the interpreter data structures that implement the heap)?

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+allowing for more flexibility. 
+
+Although the interpreter's heap is not thread safe, calls to `alloc` are still 
+made atomically, so no thread can accidentally be assigned the same index 
+in `state.heap.memory` when `alloc` is called simultaneously between threads. 
+
+### Promises
+
+We have introduced two critical syntax features to Bril: `Promise<T>` and 
+`Resolve`, significantly enhancing its capabilities in threaded execution and
```
sampsyo, line 75:
```suggestion
`resolve`, significantly enhancing its capabilities in threaded execution and
```

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+other's state. However, they also access a global heap, which is crucial for 
+modifying pointers shared across threads.
+
+Here's an example of how our notion of promises is implemented in Bril: 
+```
+@main {
+  size: int = const 4;
+  size2: bool = const false;
+  prom: promise<int> = call @mod size;
+  prom2: promise<bool> = call @mod2 size2;
```
sampsyo, line 91: There's no need to do anything particular about this now, but here is a broader comment on the language design here. Unlike many async/await frameworks, you've elected to avoid any changes to function definitions and the `call` instruction; instead, the `call` instruction is "overloaded" to do two things: (1) call a function normally, and (2) spawn a new thread.

While this minimizes the surface area of the changes, it does create a bit of awkwardness in the final language. The big one I can see is that it's not possible to create or call "normal" functions that happen to manipulate promises. For example, the identity function on integer promises cannot be used:

```
@id(p: promise<int>): promise<int> {
  ret p;
}
```

In other words, because the way you define async functions is with their return type, it's not possible to write non-async functions that happen to use the `promise<T>` type in call cases.

---

Thinking about it a little bit, one possible alternative here would be to put all the async-ification into the call instruction. That is, in this alternative version, functions need not mark themselves as returning `promise<T>` in order to be called from a spawn. Instead, we introduce a special `spawn` instruction that works like `call`. Something like this:

```
@something(x: int): int {  // just a normal function!!
  y: int = add x x;
  ret y;
}

@main() {
  five: int = const 5;
  i: int = call @something five;  // normal call
  p: promise<int> = spawn @something five;  // async call
}
```

That is, the same function `@something` could be used in either kind of call (synchronous or asynchronous); we just distinguish based on the invocation rather than the definition.

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+to not implement it this way and provide support for synchronization primitives,
+allowing for more flexibility. 
+
+Although the interpreter's heap is not thread safe, calls to `alloc` are still 
+made atomically, so no thread can accidentally be assigned the same index 
+in `state.heap.memory` when `alloc` is called simultaneously between threads. 
+
+### Promises
+
+We have introduced two critical syntax features to Bril: `Promise<T>` and
```
sampsyo, line 74:
```suggestion
We have introduced two critical syntax features to Bril: `promise<T>` and 
```

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+}
+
+@mod2(r: bool): promise<bool> {
+  ret r;
+}
+```
+
+This Bril program exemplifies the newly implemented asynchronous features in 
+the language, particularly demonstrating the use of `Promise<T>` and the 
+`Resolve` mechanism. In the `@main` function, two promises are created: `prom`
```
sampsyo, line 109: As above, I think you want to globally make `Promise` and `Resolve` lower-case to match the syntax…

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+The `Resolve` feature, synchronized with the promise mechanism, is instrumental 
+in managing the life cycle of asynchronous operations. It allows threads to 
+signal the completion of tasks and communicate results back to the main 
+execution flow. 
+
+Unresolved promises are automatically cleaned up when they go out of scope. This 
+is implemented through an atomic flag which indicates to all sub-threads to clean 
+up all resources used and abort. However, as this may cause undefined behavior
+when interacting with the global heap, a warning is printed to indicate that a 
+promise was left unresolved. In general, all promises should be resolved.
```
sampsyo, line 139: Maybe it would be good to take a stance here and say "it is an error to leave a promise unresolved"? As it stands, I can't tell whether it's actually disallowed or just discouraged.

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+
+The introduction of these asynchronous capabilities necessitated a medium-sized 
+refactor of the `brilirs` interpreter, including parameterizing the types of 
+State to accommodate the new features. A significant enhancement is the use of
+atomics in the Rust interpreter. The atomics ensure a consistent and thread-safe 
+mechanism to increment the dynamic count across all threads, which is crucial 
+for tracking and managing active threads. Overall, this implementation not only 
+enriches Bril with robust asynchronous programming and safe concurrency 
+management but also significantly extends its utility as a tool for learning 
+and experimenting with compiler design and parallel computing concepts.
```
sampsyo, line 149: Maybe you could point out the exact parts of the rust stdlib that you used for these atomics? Do you literally mean the elementary types from `std::sync::atomic`? Maybe you could discuss why you chose this over `std::sync::Mutex`?

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+In a typical mutex implementation using atomics, an atomic variable is used as 
+a lock flag. Before a thread enters a critical section, it uses atomic 
+operations to check and set this flag. If the flag indicates that the mutex 
+is already locked, the thread will wait (or "spin") until the flag changes. 
+This is often implemented using a compare-and-swap atomic operation, where the 
+flag is only set if it is currently in the expected unlocked state. Once the 
+thread has finished executing the critical section, it uses another atomic 
+operation to set the lock flag back to its unlocked state.
+
+Here's an example of how atomics work in Bril:
```
sampsyo, line 180: Do you think you could include a direct listing of all the new operations you added? Just a simple bulleted list could help complement the in-context use in the example.

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+matches the expected value, ensuring that the update occurs only if `size` has 
+not been altered by another thread. We print the result of this operation to 
+reflect its success or failure.
+
+Furthermore, we employ the `loadatomic` operation to safely read the current 
+value of `size` into `res1`, ensuring a consistent and uncorrupted value. 
+Following this, we use the `swapatomic` operation, which atomically swaps 
+`size`'s value with `three` (3). The previous value of `size` is captured in 
+`res2`. This swap operation is particularly useful for updating a shared 
+variable while simultaneously retrieving its old value. Another `loadatomic`
```
sampsyo, line 221: I'm not sure what you mean by "particularly useful" here… this seems like the _only_ thing it is useful for?

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+The `@acquire_lock` function tries to acquire a lock on a shared resource 
+represented by `lock`, an atomic integer. The function first loads the current 
+value of `lock` atomically, and then loops to keep retrying to acquire the lock 
+if it's already being used by another thread. To attempt to acquire the lock, a
+Compare-and-Swap (CAS) operation is used. 
+
+The `@release_lock` function releases the lock. It sets the lock to `0` using 
+an atomic swap operation, indicating that the lock is free. The use of atomic 
+operations ensures that these lock acquire and release operations are 
+thread-safe and prevent race conditions in concurrent environments.
```
sampsyo, line 264: Really cool! I like your suite of atomic operations, and I like this demo of how to use them to build up to a more usable synchronization construct.

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Evaluation 
+
+To evaluate our extension, we primarily focused on two key attributes: 
+correctness and runtime. To evaluate the runtime gains from asynchronous 
+programming, consider the following simple benchmark: we use a loop that just 
+adds 1 to a variable. On the first thread it performs 2000000 iterations, on 
+the second thread it performs 1000000, and so on and so forth. One good example
+of when a benchmark like this would be used is in multiplying two large matrices
+together. Below, we show the execution time in miliseconds against the number
```
sampsyo, line 286: I think I would delete the sentence about the matrix multiply… I don't think your microbenchmark/stress-test has much in common with a GEMM. :smiley: It's fine for it just to be itself, i.e., a contrived microbenchmark does nothing more than seeing how fast you can issue atomics.

content/blog/2023-12-11-AsyncBril.md
```diff
@@ -0,0 +1,328 @@
...
+of threads used in the task: 
+
+![Benchmark](eval.jpg)
+
+We observe that from 1 thread to 5 threads, we see substantial performance 
+improvements. This makes sense, as we are taking direct advantage of
```
sampsyo, line 292: One thing I don't quite understand about this benchmark: how does it change as you scale up the number of threads? Is the amount of work (total increments) held constant, so when you have N threads you are incrementing the counter, like, 2000000/N times?

sampsyo (changes requested):
This is a really interesting idea! I like how it resulted in a design with a separate heap for explicitly allocated atomics. I think it would be even more awesome if you had gotten some actual parallel computations going for the benchmarking effort, but the implementation itself is pretty cool.

I have one major IL design suggestion inline. I'd be interested in your thoughts on this alternative.

I also note that you changed your project entirely from the original proposal in #394. Can you say anything about why and how you decided to scrap your original idea?

content/blog/2023-12-11-AsyncBril.md
he-andy: I think originally what I had in mind was that we could use a this as a drop in replacement for functions already implemented, so if we had a function f: int -> int, we could wrap it in a function f_promise: int -> promise<int>. But I think your idea where we could have a dedicated async call function would've definitely been a cleaner approach to satisfy what I had in mind (and simpler to implement too, I think). Thank you for the suggestion!

evanmwilliams:
@sampsyo For sure, there were two main reasons we pivoted from the original idea.

The first is that we weren't confident in our ability to implement re-looper with the time we had for the project. We read the paper and it was pretty gnarly to understand, and without re-looper the translations become pretty trivial so maybe the scope of the project wasn't well suited for us.

The other reason is that we were much more interested in concurrency (especially after the last lecture). We wanted to do something related to performance and lost interest in our original idea. We thought the Async idea was a cool idea way to explore this. We probably should've let you know before we pivoted - in all honesty we switched up pretty late and were heads down on development so it slipped our minds - sorry about that!

sampsyo:
Cool; thanks for the extra detail!

sampsyo:
Hi, Evan et al.! This PR has "timed out"—please let me know if you'd like to reopen it.

PR #425, opened 2023-12-11
[Blog post] Final project: A DSL for Studying Metastability

### LINKED ISSUE #395: Proposal: Designing a cloud DSL for analyzing metastable failures
Researchers have recently characterized a new class of computer system failures: [metastable failures](https://www.usenix.org/system/files/osdi22-huang-lexiang.pdf).
A general blueprint for these failures is as follows:
1. The system is operating within a given environment.
2. A trigger event happens. The trigger can come in different flavors: server crash, network partition, load surge, etc.
3. The trigger pushes the system into a self-sustaining loop, in which the system is also "unavailable".
4. Even after the triggering effect is remedied, the system remains in the self-sustaining loop.

I have been trying to rigorously define metastability as a property of a suitably general system abstraction. We have a working definition that seems to capture the gist of it. However, since we are interested in actual computer systems that might experience metastability, it would be great if we could:
1. express a generic "computer system", especially one that resembles the cloud in that it is made of clusters of homogenous machines,
2. and analyze whether a given system - expressed in the previous step - "is metastable or not".

To this end, I am planning to design a generic domain-specific language (DSL) for the cloud context. There are various challenges to this from a research perspective: getting the syntax and the semantics, checking whether it is accurate with respect to real-world systems, and so on. Even more, it would be super cool if the DSL itself aids us in studying the metastability behavior of a given system via some sort of inference engine baked into it ([NetKAT](https://dl.acm.org/doi/pdf/10.1145/2578855.2535862) is a great example of this if you are interested).

My goal for this project is far more humble: to design a primitive DSL that captures most of the essentials of a cloudesque system, and to interpret it into a trace containing concrete and actual results from the execution of the system. This would serve as a basis for all future inquiries concerning metastability. Of course, the interpreter would have to output information that would be useful from a metastability analysis perspective.

For a realistic implementation, the real test for such a DSL-interpreter mash-up is accuracy with respect to some notion of a realistic modern computer system. Mimicking Kubernetes is an example that comes to mind. However, to keep things in the scope of a course project, I will only test to see whether my implementation is "correct". To expose the flavor of this correctness, here is an example. Say I have modeled an entity in the system as an agent with one input queue and one output queue, with some primitive logic to take stuff from the input, process them, and then give them to the output. A correctness criterion for this would be: "if the input queue is empty, you should not be able to get anything from it".

alifarahbakhsh:
[Closes #395]


content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+2. A load surge happens, pushing the system into the overloaded state. This causes the system to drop some requests, and some requests timeout.
+3. The requests that are being timed out start retrying, and the retries keep growing as long as the overload state is sustained.
+4. After a certain threshold, even if the load surge is mitigated and the load returns to a level that *should* be safe according to the spec, the retries keep the system in the overload state.
+
+Based on this, my first attempt to capture this type of behavior was to imitate a method that researchers have used for other systems that demonstrate such *dynamic* behaviors: I modeled a computer system as a group of agents with finite capacity that interact according to predetermined rules.
+This finitude was captured through the following abstraction for an agent:
+
+1. An agent has a finite input queue and a finite output queue.
+2. An agent has finite receiving and sending rates on its input and output queues, respectively.
+3. An agent has a finite processing rate, where "processing" means taking something from the input queue and putting it into the output queue (note that there is no content to processing).
```
sampsyo, line 40: By "no content to processing", do you mean that the only thing an agent can do with a message it receives is to put it, unchanged, into its output queue? It cannot, for example, produce N output messages for each input message?

EDIT: Never mind! I can see now that you have generalized this model in the paragraphs below. Probably just ignore this comment, but I'm leaving it here because I think it points out an interesting constraint on this notional baseline model.

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+Actions: $\alpha, \beta ::= \texttt{nop}\ | \ a = v \ | \ a = r \ | \ a = \texttt{get}(q) \ | \ \texttt{put}(q, a) \ | \ a\ +=\ 1 \ | \ x.\alpha + (\neg x).\beta \ | \ \alpha.\beta \ | \ \alpha^*$
+
+Some notes are in order (I will not include inference rules or stepping semantics in this post, so if one day you find yourself in dire need for a DSL to study metastability, this is all you have got):
+1. Every "operation" in this DSL is an action. The meaning of the term "operation" will be clearer in the next section.
+2. The "+" operator for an action designates nondeterministic choice in general, and that is the intended long-term purpose here. For now, I have limited it to deterministic branching based on the value of a predicate.
+3. The "*" operator is our favorite Kleene star. The intended long-term behavior is repeated application of an action. In practice, just like branching, it will be guarded by a predicate for now (I did not want to clutter the star - it's so beautiful!).
+4. $\texttt{put}$ only accepts state variables with a request for their value as it's second argument. I could not find a neat way of expressing this in the grammar (yes, I'm not good with grammars).
```
sampsyo:
```suggestion
4. $\texttt{put}$ only accepts state variables with a request for their value as its second argument. I could not find a neat way of expressing this in the grammar (yes, I'm not good with grammars).
```

sampsyo: :) Typically, this constraint would not be expressed in the grammar—it's the sort of thing you'd express with something like a type system or some other static rule.

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+
+Note that there are entire languages and logics just to study this networking behavior.
+This is, of course, bread and butter for distributed systems, but it is still subject to on-going research.
+I have not tried to express these networking behaviors via existing frameworks, and I am planning to capture them with adequate infrastructure and API in the execution engine for the whole system.
+
+There is also an entity called the environment, which is responsible for *triggers*.
+The environment is not expressed by the DSL, and is assumed to operate based on pure choice: each "action" for the environment is just an arbitrary state transformation subject to the limitations imposed on the environment.
+These limitations are expressed by specifying the variables that the environment can touch.
+For instance, the code for some agent can be wrapped with a giant branching action with a guard $\texttt{not crashed}$, and a $\texttt{nop}$ for the else clause.
+The environment can then choose to crash and uncrash the agent however it likes.
```
sampsyo, line 124: This conception of the environment makes sense to me, but I don't quite understand what a "trigger" is… does that terminology matter?

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+1. The environment takes an action.
+2. The system as a whole takes an action, i.e., each agent takes a big step.
+3. The network handles the messages.
+
+## Example
+Time for an example.
+This example is from an old paper that was trying to study a real-world phenomenon in routing networks.
+We have identified the phenomenon as being metastable, which is good news for the DSL's ability to express systems in which metastability is observed.
+There are other examples as well (the retry storm and a membership scenario from an industry setup), but for the sake of brevity I have not included them here.
+Note that I have heavily used sugaring, as you can imaging it is not pleasant to write actual code in the specified grammar.
```
sampsyo:
```suggestion
Note that I have heavily used sugaring, as you can imagine it is not pleasant to write actual code in the specified grammar.
```

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+    }
+}
+
+### Send
+while not outq.is_empty:
+    m = get(outq);
+    give(m);
+```
+First, some notes:
+1. $\texttt{take}$ and $\texttt{give}$ are keywords to interact with the network. I did not write them in the grammar as I am still not sure how to treat them. Informally, $\texttt{take}$ receives the next request to be delivered by the network, and $\texttt{give}$ gives a request from the output queue to the network.
```
sampsyo, line 204: This probably doesn't matter at all, but FWIW it is possible to typeset code in Markdown with backticks. So `` `take` `` would be a suitable alternative to `$\texttt{take}$` that avoids the need for the LaTeX renderer.

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+3. $r.f$ is sugar to access the field $f$ of request $r$, and $q.$is_full is sugar for checking the size of the queue (similarly for is_empty).
+
+The example here is from a [paper by Floyd and Van Jacobson](https://ee.lbl.gov/papers/sync_94.pdf).
+The paper studies synchronization of periodic routing messages.
+In a nutshell, this is what happens:
+1. Each agent, upon the expiration of its $T_t$ timer enters a listening state, during which it listens for routing messages from other nodes and processes them. Moreover, it sends them a routing message once it enters the listening state.
+2. For each message that it receives during the listening state, it spends an extra step in this state, captured by the increment to the timer for this state, $T_p$.
+3. Once the $T_p$ timer expires, the agent enters the idle state in which it is waiting for the expiration of the $T_t$ timer, captured by incrementing $t_t$. If it receives messages during this period, it just ignores them.
+4. There is a special message of the type "update", that can be injected anytime by the environment. This message abruptly takes the agent to the listening state, no matter the value of the timers.
```
sampsyo, line 214: Here is a broader question that I have been wondering about all along, but which is especially salient given the example: where does the "initial input" for these programs come from? For example, it's clear how the program in the DSL reacts to messages and events produced by the environment. But where do the messages come from, originally?

I guess I think this is important because, of course, whether the program enters metastability depends on the way it is used. I understand that your long-term goal is to do some model-checking, i.e., to check whether there is _any_ condition under which the system enters metastability. But I think you're not there yet, and you're starting with just executing a _single_ run of the system to see if it exhibits the relevant conditions. So where does the initial stimulus for that single run come from?

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+
+Metastability in this example happens when a triggering update message is injected into one of the agents, possibly by a network operator playing the role of the environment.
+As a result, all of the agents are eventually synchronized, meaning that they send their routing messages at axactly the same time.
```
sampsyo: "axactly"

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+This is undesired behavior in a realistic network, as it introduces bursts of packets periodically.
+As an analogy to the retry storm example, the self-sustaining loop here is the synchronous firing of routing messages.
+
+This instance of metastability is particularly robust, as it does not depend on concurrency at all: no matter what the ordering of messages is, the system converges to a bad self-sustaining state after the trigger.
+
+# Implementation
+I decided to implement this DSL in Rust.
+The main reason was that I wanted to learn Rust, and after spending countless hours wrestling with the borrow-checker, I can confirm that one should not toy around with Rust.
+I am, by the way, still wrestling as we speak.
```
sampsyo, line 226: It can be a serious challenge! But people often find that it can be worth the struggle in the end…

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+
+The bulk of the interpretation part of the system is ready.
+If you remember the [Typescript Bril interpreter](https://github.com/sampsyo/bril/blob/main/brili.ts), this is essentially similar to that: an infrastructure to define and hold a program, and a long list of "what to do" for each action.
```
sampsyo:
```suggestion
If you remember the [TypeScript Bril interpreter](https://github.com/sampsyo/bril/blob/main/brili.ts), this is essentially similar to that: an infrastructure to define and hold a program, and a long list of "what to do" for each action.
```

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+The pipeline first receives a piece of code written in the DSL, parses it according to a parser that I have written using [Lark](https://lark-parser.readthedocs.io/en/stable/) in python, and then uses the interpreter written in Rust to return a result.
```
sampsyo:
```suggestion
The pipeline first receives a piece of code written in the DSL, parses it according to a parser that I have written using [Lark](https://lark-parser.readthedocs.io/en/stable/) in Python, and then uses the interpreter written in Rust to return a result.
```

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+3. On the long run, one would hope to have a performant interpreter, i.e., one that can interpret large systems in a short span of time. The crux of the difficulty lies in parallelizing the execution in the right way. While there are straightforward ways of attempting this, I have not yet found a good way of doing it.
```
sampsyo, line 243: Indeed; writing parallel interpreters can be really hard. One thing you might consider, in the long term, is to sacrifice parallelism entirely for the straightforward interpreter mode. Instead, you can exploit parallelism at the level of your model checker, i.e., the exploration of a large space of inputs. That exploration can be much more naturally amenable to parallelism.

content/blog/2023-12-11-A DSL for Metastability.md
```diff
@@ -0,0 +1,237 @@
...
+4. The previous point is exacerbated by the fact that one would also like to have a model-checker that would use the interpreter. Ideally, the model-checker would receive a given system and produce a trace "ending" in a bad self-sustaining loop caused by a trigger. Following the maxim "immature optimization is the root of all evil", I have not spent much time thinking about this yet, but I am expecting some serious issues down the road.
```
sampsyo: One big thing that occurs to me would be required for this: a way to _detect_ metastability. This would require operationalizing a definition of metastability that you can check by looking at the current state, or perhaps the trace of recent states.

sampsyo (changes requested):
This is really cool! Thanks for the extremely clear writeup; I think you did a great job introducing the background, and the example really shows off where you would like to go with the DSL. It would, of course, have been even more impressive if you had somehow gotten the whole interpreter working far enough to actually execute the programs you have evaluated on paper, but I think this is an impressive start regardless.

I just have a few minor suggestions within!

content/blog/2023-12-11-A DSL for Metastability.md
alifarahbakhsh: That's a good point! To explain this I would have to unpack our definition of metastability, which is based on temporal logic, and it would take a lot of space. That is why I decided to leave it out. In summary, there exists a predicate P over environment actions, and a trigger is an initial period of the environment going crazy, until it "eventually" satisfies "always" P.

content/blog/2023-12-11-A DSL for Metastability.md
alifarahbakhsh: For now we have been crafting the initial conditions by hand, in order to capture some metastability examples. I think the initial conditions have to be defined via a specification of the system's desired behavior.

As for the inputs, they are usually packed into environment actions. For instance, the input load for a consumer service is captured by the environment just arbitrarily putting requests into the network, to be sent to some agent in the system.

content/blog/2023-12-11-A DSL for Metastability.md
alifarahbakhsh: Absolutely. We have realized that, to make things even harder, metastabilty might actually be a hyperproperty and not a property.

content/blog/2023-12-11-A DSL for Metastability.md
alifarahbakhsh: Fixed it.

content/blog/2023-12-11-A DSL for Metastability.md
sampsyo: Sounds great. Maybe it be worth just briefly mentioning that you are also interested in defining metastability formally so you can build an automatic detector, but not going into detail about what that definition actually is.

content/blog/2023-12-11-A DSL for Metastability.md
sampsyo: That makes sense! Could be worth briefly mentioning how that works in the post.

alifarahbakhsh: Done.

content/blog/2023-12-11-A DSL for Metastability.md
alifarahbakhsh: Done.

sampsyo (approved):
Looks great! Nice work again.

PR #430, opened 2023-12-12
[Blog] Caiman Frontend Project

### LINKED ISSUE #402: Project Proposal: Caiman Frontend
**What will you do?**

Caiman is an IR for CPU/GPU computing that makes explicit how, when, and where computations occur. It seeks to strike a balance between single-source solutions like CUDA and SYCL which make optimization more difficult by hiding details pertinent to performance and direct API usage. Caiman consists of value, spatial, and timeline specification languages and an implementation language.

I will be implementing a frontend to allow users to write Caiman programs in an easy-to-use, high-level syntax. Specifically, for this project, I will implement the lowering from the front-end AST to the Caiman assembly. Caiman assembly is in continuation passing style, so one of the challenges will be converting the front-end control flow into CPS. Furthermore, the Caiman assembly lacks many of the instructions one might find in a typical IR such as arithmetic operations. The front end, however, has such operators that must be lowered to invocations of external functions on the host device.

If I have time, I'll also need to implement type deduction with Hindley–Milner.

**How will you do it?**

First, I'll need to flatten the nested AST of the front-end which should be straightforward.
I'll then generate external functions in Rust for basic operators that are built into the front end and convert other higher-level constructs into a series of Caiman assembly instructions. For example, slices could be lowered as a buffer (pointer) and a start and end index. 

For converting to CPS, I can turn each basic block into separate functions and convert conditions to selections of the next continuation to run.

I plan on starting by writing some simple-ish motivating examples and focusing on the work required to fully lower these frontend examples into caiman assembly and running them end-to-end through the caiman compiler. Success will be the successful lowering of these examples/test cases and getting the correct result when executing the compiled programs.

stephenverderame:
Closes #402


content/blog/2023-12-11-hlc/index.md
```diff
@@ -0,0 +1,538 @@
...
+the scheduling language (implementations) for this post.
+
+# Design
+
+The way I achieved this was by introducing a new high-level IR (HIR) that would
+be the target of the front-end AST.
+
+![Architecture](HlcArch.svg)
+
+The HIR I designed is somewhat similar to BRIL, however, its canonical
```
sampsyo, line 172: BRIL -> Bril

sampsyo, line 172: I'd replace the first `,` with a `;`.
https://capra.cs.cornell.edu/styleguide/#however

content/blog/2023-12-11-hlc/index.md
```diff
@@ -0,0 +1,538 @@
...
+and divided up into basic blocks, but unlike Caiman Assembly 
+it has traditional (non-CPS) control flow, and the inputs and outputs of
+funclets aren't explicit. One thing to note is that HIR function calls are considered
+basic block terminators, since in CPS a function call must be the last
+statement in a function. Furthermore, CFGs are canonicalized so that they have 
+a single entry and a single exit block.
+
+Once the AST is in CFG form, we perform a few local transformations as "preprocessing".
+Namely, we convert all operators to their typed forms 
+(so `add` becomes `_add_i64_i64` for adding two `i64`s) and convert all
```
sampsyo, line 185: This is usually called "name mangling."

content/blog/2023-12-11-hlc/index.md
```diff
@@ -0,0 +1,538 @@
...
+mutations of variables to loads and stores of a reference.
+
+One could imagine employing the progressive lowering ideas of CompCert and MLIR
+to separate the HIR into two HIRs, one before these transformations and
+a second representation afterward. I didn't go this route because it would
+involve decent code duplication for the definitions of the two IRs.
```
sampsyo, line 191: Good point! I'd use your strategy in this situation, fwiw.

content/blog/2023-12-11-hlc/index.md
```diff
@@ -0,0 +1,538 @@
...
+        v = one;
+    }
+    @in { v: [node(main.z)-usable, none(space)-save] };
+    return v;
+}
+```
+
+It's still not pretty on account of the current need to match pieces of 
+the implementation back to the spec almost everywhere (denoted with the `@`) 
+and keep expressions unnested. However, the schedule in this example has the following
```
sampsyo, line 244: Do you think it would be possible to include a brief explanation of what the types mean? That is, in `bool @ node(main.b)`, how should we read each part? How about in `[node(main.r)-usable, none(space)-save]`? Including a few direct descriptions like these might make it easier for readers to parse the example code.

content/blog/2023-12-11-hlc/index.md
```diff
@@ -0,0 +1,538 @@
...
+    style="max-height: 50vh; display: block; margin-left: auto; margin-right: auto" />
+
+
+### CFG Generation
+
+During CFG generation, one piece of information we need to keep around is what
+the continuation of each basic block is. In the above example, the
+continuation of `foo` is `foo2` and the continuation of `foo2` is `foo3`.
+`foo4`, `foo5`, `foo6`, and `foo7` are at the maximum depth in this example, so
+when they're done, control is going to pass back to their parent's continuation.
```
sampsyo, line 257: Based on the examples, am I correct in concluding that the continuation for a given block $b$ is the "join point" following $b$? Or, more formally, that it's $b$'s immediate post-dominator?

content/blog/2023-12-11-hlc/index.md
```diff
@@ -0,0 +1,538 @@
...
+```
+
+Here we can see that the actual return value of `%block1` is just
+`%v`. More generally, a block's return values are the inputs to
+its continuation. For the last block in a function, then the
+outputs will be the outputs of the function instead.
+
+Luckily, we already know what the continuation of every block is because
+we determined this information when we generated the basic blocks from the
+source program.
```
sampsyo, line 376: This is very tricky! Interesting observation here.

sampsyo (changes requested):
Wow! Really nice work on this writeup, @stephenverderame. You have done an absolutely fantastic job explaining some tricky concepts in a very complicated language. The frontend itself is looking good, and the writeup here is excellent. Thanks for the very clear exposition of the problems and their solution. :clap:

I have included a few low-level suggestions for the text!

sampsyo:
Hi, Stephen! This PR has become somewhat stale over the last couple of years, and I'm cleaning up old PRs. Please let me know if you'd like to finish this off at some point. :)

PR #431, opened 2023-12-12
[Blog - Final Project] Array IR for Mixed Cryptography

### LINKED ISSUE #401: Project Proposal: Array IR for Mixed Cryptography
**What will you do?**

We will work on [AIRduct](https://squera.github.io/fcs-pdf/dingFCS2023.pdf), the array-based IR for the [Viaduct](https://www.cs.cornell.edu/andru/papers/viaduct/viaduct-tr.pdf) compiler.
This is based on a research project that Vivian has been working on for the last couple of years.

Viaduct is a compiler which synthesizes secure, distributed programs employing an extensible suite of cryptography.
It uses security specifications in the form of information flow labels to select cryptographic protocols that make the
compiled program secure. In this project, we'll assume this choice has been made correctly, and focus on the IR.

AIRduct is unique in that it is an IR for a cryptographic compiler which is capable of representing programs which
cannot be implemented as a single crytographic circuit. This enables support for interactive programs, and programs
mixing multiple cryptographic back ends and local computation, which is useful for performance.
AIRduct represents code employing multiple cryptographic mechanisms as structured control flow over calls to circuit
functions. This allows for a separation of impure behavior (I/O) and control flow from pure cryptographic computation,
which does not allow complex control flow.
We use an array-based IR since this preserves useful structures upon which we can optimize computations.

Cryptographic back-ends (extensible portions of the compiler which implement crypto mechanisms) can (probably) optimize most aggressively when the circuit functions are larger. So, one aspect of implementing the IR we are particularly interested in is how to produce maximally large circuit blocks.

We are focusing on implementing the following:
- Supporting control flow (if, while)
- A grouping procedure for creating circuit blocks (in the IR) from programs in which each statement is annotated with
a protocol. This will include reordering of code.
- Optimizations by back-ends, to show the fruits of our work which enables them

**How will you do it?**

- Control flow should hopefully be straightforward.
- Grouping requires analysis regarding which reorderings are safe, with respect to both data dependencies and security.
- Optimizations will be non-trivial but inspired by prior work.
    - One optimization is simply calling into SIMD instructions which are natively supported by many cryptographic
    libraries.
    - A more interesting optimization is the combination of bulk instructions. One example is merging two array map 
    expressions into one, when the second map operates on the output of the first one (similar to loop fusion).

- We will measure success of implementing control flow checking correctness on existing test cases.
- We are not confident how to measure success of grouping yet. A good metric for the quality of a grouping might be the resulting number of blocks, but it's not obvious whether one can compute the minimal number of blocks to compare our implementation to. There doesn't exist a unique maximal grouping, but there definitely exists some set of maximal groupings for any given program.
- The input code will be manually written test cases. We will try to write programs which use each implemented
feature, such as taking different paths through the CFG and taking advantage of vectorization. We will also include
benchmark programs, which were used by the original Viaduct paper and are also often used to evaluate other
cryptographic compilers.
- We will test for correctness by ensuring that output is the same with and without calling into cryptographic
protocols, that is, a computation using cryptography should yield the same result as in cleartext; the only difference
being the added security properties.
- We will think very hard about security. This is something that needs to be reasoned about abstractly rather than 
relying on testing.
- We can measure the benefit of optimizations by examining run time.

**Team members:**
@vivianyyd
@willwng

willwng:
Closes #401

Final Project with @vivianyyd


content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
+ +++
```
sampsyo:
```suggestion
+++
```

content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
...
+We did not complete the proposed task of implementing optimizations for various protocol backends. 
+
+# Control flow
+We added three types of statements: `if`, infinite `loop`, and `break`. Implementation was quite straightforward, as anticipated. 
+
+We decided that control flow instructions should be statements rather than expressions. This is because all expressions are associated with a protocol, but control flow must be executed outside of any cryptographic protocol. 
+
+Implementing control flow required a slight change to how “blocks” (sequences of code delimited by curly braces) are represented in our AST. Previously, blocks were only used to represent function bodies, and therefore required that the last statement was always a `return`. We decided to introduce distinct notions of blocks for function bodies and control flow branches. This is because the invariant that each function contains a single unique `return` statement makes code generation far more straightforward, and doesn’t sacrifice expressivity. 
+
+We evaluated this implementation using handwritten tests to ensure all branches through control flow are properly executed.
```
sampsyo: I presume you also checked that everything still worked for the "old" use of blocks (i.e., on all test cases you have that *don't* use control flow)?

content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
...
+1. All let-bindings are annotated with a protocol. In the circuit IR, only let-bindings involving a “command” (input, output, and function/circuit calls) are annotated with protocols denoting where results are stored. Let-bindings involving computations occur within circuits, which are annotated with the associated protocol for all included statements. 
+2. It includes downgrade nodes. These are explicit violations of secure information flow, such as declassifying secret data or endorsing untrusted values. It is necessary that downgrade instructions are preserved in the precircuit IR, since they enforce a secure ordering on computations.
+
+# Grouping
+The grouping procedure places statements of the same protocol together whenever possible. This will enable larger circuit blocks, reducing the overhead of communication between protocols.
+
+This reordering is subject to data and security dependencies. Node `x` is a dependency of node `y` when:
+- `x` defines data that is used by `y`,
+- `x` is an `input` statement, and `y` an output,
+- `x` is an `endorse` statement, and `y` an declassify, or
```
sampsyo:
```suggestion
- `x` is an `endorse` statement, and `y` a declassify, or
```

content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
...
+- `x` is an `output` statement, and `y` an input from the same host.
+
+The first rule corresponds to data dependencies. The second and third rules hold because moving a statement which reveals information before one that commits to data violates security. The last rule preserves the user interface.
```
sampsyo: Maybe this is silly, but what about an `output` followed by an `output`? Do those have to have their order preserved? (This would be true for memory writes, for example, because you want to preserve "overwrites" of the same memory location. But maybe this is not possible with AIRduct outputs… if so, maybe the blog post needs a little more background on what "inputs" and "outputs" are in this setting?)

content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
...
+	<B>
+	<C>
+}
+```
+Then `B` might be reordered with `C`, and the entire `if` block could be reordered with `A`. However, we choose not to allow reorderings where `B` or `C` may be lifted out of the block. This allows us to essentially ignore control dependencies, making reordering simpler.
+
+The grouping procedure constructs a dependency graph for each block (a sequence of code enclosed with curly braces), where nodes are statements. For instance, given the above code, the procedure constructs a dependency graph containing `A` and the `if` statement, and another for `B` and `C`. Reordering occurs only within these two graphs. Notice that these graphs do not need to maintain control-flow information, so statements can be reordered strictly based on data and security dependencies.
+
+
+We formalize the reordering task as the following problem. Given a directed acyclic graph $G = (V,E)$ where vertices are statements and edges are dependencies, and an assignment of protocols to nodes $p : V \to \mathbb{N}$, we aim to find a sorting $\pi$ of the nodes ($v_{\pi_1}, v_{\pi_2}, \dots, v_{\pi_n}$) such that:
```
sampsyo, line 68: I think you need to add `latex = true` to the front matter to make the LaTeX math render.

content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
...
+- For all $e = (v_1, v_2) \in E$ (the execution of $v_2$ depends on $v_1$), $v_1$ is ordered before $v_2$
+- The number of protocol changes within the sequence is minimized. For each consecutive $v_{\pi_i}, v_{\pi_j}$ in the sort, there is a change if $p(v_{\pi_i}) \neq p(v_{\pi_j})$.
```
sampsyo, line 70: Great! This is a very clear description of the optimization objective.

content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+We are unsure of the hardness of this problem. We (unsuccessfully) tried several reductions from NP-Complete problems, and have not yet proven that our polynomial-time algorithm produces an optimum result. This would be an interesting task to pursue in the future. We think the problem might be in P.
```
sampsyo: I also don't see an immediate NP-hardness reduction. The fact that the scheduling constraints are just expressed as a dag rather than, like, integer deadlines makes it hard to see how to do a reduction from any "classic" scheduling problem. I guess I agree with you that it seems likely that it is in P?

content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+This Python code generates a random DAG and assigns a random protocol to each node. The code finds the optimal solution by brute-forcing every topological ordering of the nodes and finding the minimum number of protocol changes required; the limitation of these tests is that they are not very scalable and we could only realistically test on small DAG’s. We then compare this result to our heuristic and find the following “correctness” distribution:
+<p align="center">
+    <img src="error_histogram.png" alt="error_histogram" width=450 />
+</p>
+
+Here, the error is defined by the number of swaps computed by our algorithm compared to the optimal solution. `p` is defined to be the probability of a DAG edge existing. 
+
+From these results, we see that our heuristic gives an optimal solution nearly 90% (at least for smaller DAG’s). We also estimate that the smallest approximation factor that describes our algorithm is `ɑ=2`, where our algorithm gives twice as many swaps as the optimal solution (number is obtained by a small adversarial example).
```
sampsyo: "nearly 90%": add "…of the time"? Or make it "for nearly 90% of the randomly generated graphs"?

content/blog/2023-12-11-array-ir/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+We also computed how the average percent error is affected by the connectivity of the DAG:
+<p align="center">
+    <img src="error_vs_p.png" alt="error_vs_p" width=450 />
+</p>
+
+
+These results seem to make sense; sparse graphs are rather uncomplicated and most solutions will already be optimal, and highly-connected graphs have a smaller space of possible topological sorts. It would be interesting to perform a more rigorous analysis of the algorithm to explain the empirical results.
+
+We also tested our implementation on handwritten programs in the pre-circuit language, both based on existing benchmarks for cryptographic compilers and programs which mix multiple protocols in suboptimal orders. For each example, maximally large groupings were constructed. One important task for future work is to test on more representative benchmarks, as many of the existing benchmarks are small and do not use more than two cryptographic protocols.
```
sampsyo: While it is also useful to see the general characterization of the heuristic on randomly generated data, it would be really interesting to know it performs on "real" programs. One thing I'm not quite clear on is whether a frontend exists for your pre-circuit language… is this something you can obtain from other benchmarks that have already been written somewhere? Or do you need to manually translate these benchmarks into your system?

sampsyo (changes requested):
Sounds very cool! I liked the detailed description of the optimization problem and your heuristic solution to it. Y'all have done a great job coming up with a creative way to characterize the optimality of this solution empirically, in the absence of theoretical guarantees.

One thing that would have been useful to know is something about the prevalence of these issues in real code. Maybe that would be good to explore in the future!

willwng:
Thanks for the suggestions! I couldn't get the comments to line up with the revised file, but I believe all the comments should hopefully be answered in the latest commit. Most notably, we added the statement about how benchmarks for typical cryptographic compilers support only one protocol, so we had to hand-write our own. Hopefully that will change in the future!

sampsyo (approved):
Got it; looks great! Nice work on this.

PR #433, opened 2023-12-12
[Blog] Polyhedral Compilation

### LINKED ISSUE #405: Project Proposal: Polyhedral Compilation
## What will you do?

The high-level goal of this project is the implementation of loop optimizations via the polyhedral model. There is potentially a lot in scope here, so I intend to start small and then scale up. A baseline proof of concept might be to implement a loop interchange optimization that transforms a nested loop of the form

```
for i in range(10):
  for j in range(10):
    a[j][i] = f(i, j)
```

to one of the form

```
for j in range(10):
  for i in range(10):
    a[j][i] = f(i, j)
```

with the intent of improving locality. However, the goal of the project is more about implementing the polyhedral framework rather than any one particular optimization.

## How will you do it?

Currently, I don't have a perfectly clear picture of all the steps necessary to implement such an optimization via the polyhedral framework. Thus, the first step will be to review the relevant literature. A good start would be [Cédric Bastoul's Thesis](http://icps.u-strasbg.fr/people/bastoul/public_html/research/papers/Bastoul_thesis.pdf), which seems to be one of the more accessible introductions.

A rough idea of the steps required might be:

1. Identify static control parts.
2. Identify dependencies.
3. Identify a potentially profitable transformation.
4. Build constraints representing the bounds, dependencies, and transformation.
5. Solve via an off-the-shelf solver such as [`isl`](https://libisl.sourceforge.io/) to determine the legality of the transformation.

Though again, I should emphasize that these steps are based only on my rudimentary knowledge of the framework and may well be incomplete or incorrect.

For the IR, Bril seems inappropriately low-level. One possible choice is the [FPCore](https://fpbench.org/spec/fpcore-2.0.html) benchmark language, which would allow me to reuse existing infrastructure that I have built for the language.

## How will you empirically measure success?

The [PolyBench](https://web.cse.ohio-state.edu/~pouchet.2/software/polybench/) benchmark suite should be a good source of programs with which to test correctness and performance. If I were to use FPCore as the benchmark language, this would require translating the PolyBench benchmarks from C.

Since the goal of many loop optimizations is to improve cache performance, it will be important to evaluate optimizations by running the code on a real machine rather than an interpreter. The [FPBench exporter](https://fpbench.org/tools.html) can facilitate such an evaluation by compiling the transformed programs to C, which can then be compiled and executed.

## Group Members

Just @bcarlet.

bcarlet:
Closes #405.

### DIFF
--- a/content/blog/2023-12-11-polyhedral/index.md
+++ b/content/blog/2023-12-11-polyhedral/index.md
@@ -1,0 +1,134 @@
++++
+title = "Implementing the Polyhedral Model"
+[extra]
+latex = true
+[[extra.authors]]
+name = "Benjamin Carleton"
++++
+
+## Introduction
+
+The oft-repeated wisdom that programs spend most of their time in loops motivates the need for a wide variety of loop optimizations. A challenge is finding a representation that can efficiently reason about the large sets of operations arising from loop programs, whose number and schedule may also depend on dynamic parameters.
+
+In the polyhedral model, executions of a statement within a loop nest are represented by points within a convex integer polyhedron, with loop bounds and other constraints forming the boundaries of the polyhedron. Formally, the iteration domain is given by
+$$
+\mathcal{D} = \\{ x \in \mathbb{Z}^n \mid Ax + b \geq \mathbf{0} \\},
+$$
+where $n$ is the depth of the loop nest, which admits a compact representation in terms of $A$ and $b$. For example, consider the following program:
+
+```c
+for (int i = 0; i < n; i++)
+    for (int j = 1; j <= i; j++)
+        a[i][j] = a[i - 1][j - 1] + a[i - 1][j];
+```
+
+The iteration domain for the innermost statement can be visualized as a two-dimensional region:
+
+<p style="text-align: center">
+  <img alt="Iteration Domain" src="domain.svg" style="width: 40%">
+</p>
+
+(where we have fixed $n = 5$ for the purpose of visualization), and the region may be described compactly by a system of affine inequalities:
+$$
+\begin{bmatrix} 0 & 1 \\\\ 1 & -1 \\\\ 1 & 0 \\\\ -1 & 0 \end{bmatrix} \begin{bmatrix} i \\\\ j \end{bmatrix} + \begin{bmatrix} -1 \\\\ 0 \\\\ 0 \\\\ n - 1 \end{bmatrix} \geq \mathbf{0}.
+$$
+The execution order is a simple lexicographic ordering of the points. A wide range of loop transformations can be represented by a simple change of basis producing a new lexicographic ordering. A larger class of loop transformations may be represented by affine transformations of the polyhedron.
+
+## Goals and Design Considerations
+
+The goal of this project was to implement a simple optimization such as loop interchange via the polyhedral framework, with emphasis on implementing interesting parts of the model rather than plumbing together existing tools.
+
+There exist varying approaches for many parts of the polyhedral framework. In this work, I closely followed the approach outlined by Cédric Bastoul in his thesis on the polyhedral model.[^1]
+
+### Intermediate Representation
+
+The polyhedral model requires reasoning about loops at a high level of abstraction; a good intermediate representation should make it easy to query high-level information such as loop bounds, steps, and nesting. A low-level IR that represents loops only implicitly through unstructured control flow would thus be inappropriate. Similarly, memory references should expose their subscript expressions explicitly rather than via low-level pointer arithmetic.
+
+The above considerations motivate the choice of [MLIR](https://mlir.llvm.org/) for our intermediate representation. MLIR's multi-level design admits dialects that can represent all the above features explicitly. In particular, the `scf` dialect represents structured control flow such as `if` statements and `for` loops, while the `memref` dialect represents (possibly multidimensional) array accesses.
+
+For example, consider the following program fragment:
+
+```c
+for (int i = 0; i < 10; i++)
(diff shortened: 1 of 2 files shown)


bcarlet:
[Repository.](https://github.com/bcarlet/cs-6120-project)

content/blog/2023-12-11-polyhedral/index.md
```diff
@@ -0,0 +1,134 @@
...
+2. Merge consecutive operations with static control into static control parts.
+
+The result is a list of the SCoPs within a given function. Implementing this analysis in MLIR is relatively straightforward due to the IR's explicitly nesting regions.
+
+### Building Iteration Domains
+
+Having identified regions with polyhedral domain, the next step is to construct those domains explicitly in a format that can be manipulated by integer programming tools. We use [isl](https://libisl.sourceforge.io/) to build sets from affine constraints and discharge queries such as emptiness checking.
+
+For a given operation, we begin by determining the space in which its iteration domain should be constructed. This determination depends on the SCoP in which the operation resides: loop induction variables within the SCoP correspond to dimensions of the space, while other variables correspond to parameters.
```
sampsyo: I think this is just my ignorance about the polyhedral model showing, but I don't quite grok what "space" and "dimensions" mean concretely here. Maybe it would be helpful to add a little context-setting above about what isl defines a "domain" to be? What are all the ingredients required to create an iteration domain?

content/blog/2023-12-11-polyhedral/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+To construct the iteration domain, we traverse the loop nest moving outwards from the operation, accumulating constraints imposed by loop bounds and `if` conditions, until we reach the root of the SCoP. Each constraint we encounter introduces an inequality between affine expressions. We convert these expressions from MLIR `Value`s into isl's internal representation of affine functions, construct integer sets corresponding to the inequalities, and finally compute the intersection across all constraints.
+
+In the implementation, this analysis is wrapped up in an MLIR pass that annotates each operation in the IR with an attribute encoding the operation's iteration domain. The figure from the introduction was generated from the pass' output.
```
sampsyo: Do you think you could include one or two examples of what that annotation looks like, in MLIR syntax? I know it will be messy, but I think it would be helpful to get a concrete picture of what the domain consists of, in a way that is separate from the internal isl data structures.

content/blog/2023-12-11-polyhedral/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Building Dependence Polyhedra
+
+Testing the legality of transformations in the polyhedral model requires encoding dependence relations between statement executions. Dependences arise from accesses to a common memory location, where one access is a store. There are many techniques for identifying such dependences; following Bastoul, we take the approach of dependence polyhedra.
+
+A dependence polyhedron, for a given pair of statements, encodes dependences between executions of the respective statements as integer points. The question of whether the statements are free of dependences then amounts to an emptiness check on the polyhedron.
```
sampsyo: FWIW, it took me a couple tries to understand this paragraph: namely, each dependence is _one point_, so the set of all dependencies is a polyhedron (a set of points).

content/blog/2023-12-11-polyhedral/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+1. An equality constraint between the subscript functions from each access to the shared array.
+2. The constraints describing the iteration domain of each statement, as computed above.
+3. A set of *precedence constraints* which, admittedly, I don't fully understand. Roughly speaking, these encode information about the ordering of the statements' executions.
+
+The implementation reuses much of the infrastructure needed for the previous analyses. One inconvenient technical issue that arose in the implementation was the need to build constraints between affine expressions that live in different spaces. I was able to get by with some trial and error, but likely there were better ways of handling this.
+
+## Evaluation
+
+Unfortunately, the project has not reached a point where it is practical to perform a systematic evaluation across a wide range of benchmarks. To check correctness, I have thus far relied on manual inspection of the internal data structures. I've tested with a modest set of examples from the literature, with mostly encouraging results.
```
sampsyo: That's OK! Can you say a bit more about what that modest set of examples consists of?

content/blog/2023-12-11-polyhedral/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Challenges
+
+There was a considerable amount of background material to learn before and throughout the project, given that I had no prior experience with polyhedral compilation. Learning the requisite infrastructure (MLIR, isl) was also occasionally troublesome. In particular, figuring out how to do relatively simple things in isl sometimes required lots of experimenting with sparsely documented functions.
+
+With respect to the implementation itself, construction of the dependence polyhedra was certainly the trickiest part, and is also the part of the implementation in whose correctness I'm least confident. And as I mentioned above, testing is currently limited to manual inspection of internal data structures (though I did write utilities to visualize those structures where appropriate).
+
+## References
```
sampsyo: I don't think you need this section heading; the footnote will be rendered in a way that makes it clear it's not part of the body text. :smiley:

sampsyo (changes requested):
Thanks for the detailed writeup! This was definitely a hard thing to do, but you made a good case for why you wanted to start from `scf`/`memref` and not `affine`. I think it's pretty cool that you were able to get the domains constructed using isl at all. It of course would have been better to get real code emitted and executing so you can check correctness, but this was a good start.

I just have a few high-level comments on the text.

sampsyo (approved):
Excellent! Nice work on this!!

PR #437, opened 2023-12-12
[Blog - Final Project] CirC Bitwidth Adjustment Optimization

collinzrj:
Final project with @MelindaFang-code

### DIFF
--- a/content/blog/2023-12-11-circ-bitwidth-adjustment
+++ b/content/blog/2023-12-11-circ-bitwidth-adjustment
@@ -1,0 +1,84 @@
++++
+title = "CirC Bitwidth Adjustment Implementation"
+[extra]
+bio = """
+  Collin Zhang and Melinda Fang are students at Cornell
+"""
+[[extra.authors]]
+name = "Collin Zhang"
+link = "https://collinzrj.github.io/"  # Links are optional.
+[[extra.authors]]
+name = "Melinda Fang"
++++
+
+## Link to the project
+We are currently working on creating a PR to include the change to the project. At this point, the best way to see the changes we have made is via this link which compares the latest commit with the commit before our changes https://github.com/collinzrj/circ/compare/a26533baadfd674f7d7d766b995a9a78a02d4ebd..61fb7c0ee2946cf84b92d962236ffe9533cb6c53
+
+We mainly make changes to these files
+```
+## The new optimization pass is implemented in these two files
+src/ir/opt/mod.rs
+src/ir/opt/short_int_adj.rs
+## The logic to handle non-adjust operations in the constraint generation system
+src/target/r1cs/trans.rs
+```
+
+## Background
+[CirC](https://github.com/circify/circ) is a compiler that compiles high level languages {C, ZoKrates, Circom} to (state-free, non-uniform, existentially quantified) circuits {SMT, ILP, R1CS, MPC} for SMT solver, zero-knowledge proof, multiparty computations, and more. To be more specific, it converts a program's source code into a set of mathematical constraints represented by vectors of variables. Optimization of such compilers primarily focus on reducing the size of the constraint system: having fewer constraints reduces the workload for creating proofs, and might also result in shorter proofs to verify. The typical formalism adopted by the majority of proof systems is rank-1 constraint system (R1CS). In an R1CS instance C, there exists a collection of constraints over a finite field F, often a prime p, or in other words, a circuit of additions and multiplications gates mod p. One significant cost of such system is introduced by arithmetic operations. Assume an underlying prime number p and operations such as multiplication done over p′, to obtain a correct result, a remainder operation is always applied on the result to adjust bitwidth and avoid overflow, which leads to a number of constraints that are at least equal to the bitwidth of the result, as it requires at least one split gate. Our project aims to apply a heuristic approach to optimize the translation of arithmetic operations. 
+
+## Design
+Our optimization mainly focuses on addition and multiplication, which would typically increase the bitwidth of resulting element by 1 or double the bitwidth, respectively. We classify elements into two types
+- Elements that could be above the range and required to be returned within range: this includes program output, elements involved in bitwise operations and comparisons, operations like division or remainder, elements involved in memory operations, and so on.
+- Elements that could be above the range but are not required to be within range: This includes intermediate elements that undergoes multiplication and addition operations and does not meet any of the criterions above.
+
+The current system represents basic elements as bitvectors, and is designed to adjust their bitwidth after each arithmetic operation. Based on the classification of the elements above, our projects reduces the number of constraints in the system by greedily deciding when to adjust bitwidth for elements falls under the second category while ensuring no overflows can happen in later operations involving the elements. Or in other words, adjustments are only introduced if the next operation is going to result in an overflow.
+
+Take this short program as an example:
+```
+def main(u32 a, u32 b) -> u32:
+    u32 c = a + b
+    u32 d = a + c
+    return c * d
+```
+
+The old compiler would adjust immediately at `a+b`, `a+c` and `c * d`. By applying our optimization, the system would treat c and d as intermediate elements and only adjust at c*d, since it is producing the final program output. 
+Without our optimization, the cost spent on adjusting the bitwidth will be: 33 constraint to adjust `a + b` and `a + c` respectively, and 64 to adjust `c * d`, which result in 130 constraints in total.
+In contrast, after our optimization, the cost spent on adjusting `a + b` and `a + c` would be 0, and 66 to adjust `c * d`, which result in 66 constraints in total. 
+
+## Implementation
+1. We create a new type of operations: `add_unadjust` and `mul_unadjust` for addition and multiplication to notify the constraint generation system not to adjust the bitwidth.
+2. We implemented two passes to rewrite the operations. First pass is to classify elements in the system. We implemented an initial pass to scan the entire program and store elements requiring adjustments in a set S. Whenever we encounter an operation that definitely requires adjustment, we add the elements used in the operation to S. Then we implemented a rewrite pass to rewrite operations done on bitvectors not included in D to our newly introduced unadjust operation. 
+3. We have also made changes to the constraint generation system to take advantage of the two new operations. In the original implementation, the constraint generation system will always adjust the bitwidth of the result of an operation. We have made changes to the system to not adjust `add_unadjust` and `mul_unadjust` operations. We also proved that such change will not impact the correctness of the system.
+
+## What were the hardest parts to get right
+The hardest part is to make sure the compiler works properly using our newly introduced operation. We encounter many bugs while implementing the optimization. For instance, before since the author always adjust the bitwidth after all types of operations, the bitwidth is fixed. However, now since an operation might be using unadjusted bitwidth the logic of calculating the bitwidth of the current element is more complicated and requires more steps to ensure an operation does not overflow.
+
+## Empirical evaluation
+We uses two types of tests. 
+1. We use the existing benchmarks in the circ repo to test whether the output of the optimized compiler is correct. Our implementation passed all the test cases in CirC.
+2. To measure how well the proposed design improves the system, we use the number of r1cs constraints generated as the metrics. We runs our compiler and the original one over a set of zokrates programs in [CirC](https://github.com/circify/circ) and [Zombie](https://github.com/PepperSieve/Zombie/tree/master) and compares the number of constraints generated. Here is a shorted list of tests we run. The first line are the numbers of constraints with our flexible adjustment implementation and the second line is the result using original compiler.
+```
+../Zombie/circ/zkmb/shaRound.zok
+with optimization:    27137
+without optimization: 29057
+../Zombie/circ/zkmb/DotChaChaAmortized_Isolated.zok
+with optimization:    23366
+without optimization: 41835
+../Zombie/circ/zkmb/DotChaChaAmortized.zok
+with optimization:    111529
+without optimization: 130232
+../Zombie/circ/zkmb/tls_key_schedules/HKDF.zok
+with optimization:    101304
+without optimization: 109187
+../Zombie/circ/zkmb/SHA.zok
+with optimization:    26581
+without optimization: 28505
+../Zombie/circ/zkmb/DotChaChaAmortizedUnpack.zok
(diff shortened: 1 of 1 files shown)


content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+"""
+[[extra.authors]]
+name = "Collin Zhang"
+link = "https://collinzrj.github.io/"  # Links are optional.
+[[extra.authors]]
+name = "Melinda Fang"
++++
+
+## Link to the project
+We are currently working on creating a PR to include the change to the project. At this point, the best way to see the changes we have made is via this link which compares the latest commit with the commit before our changes https://github.com/collinzrj/circ/compare/a26533baadfd674f7d7d766b995a9a78a02d4ebd..61fb7c0ee2946cf84b92d962236ffe9533cb6c53
```
sampsyo, line 15: Instead of listing this right at the top of the post, how about including it more elegantly as a link somewhere in the body text? Like, maybe when you introduce the idea for the first time, you can mention what the current status is and link to your code there.

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+
+We mainly make changes to these files
+```
+## The new optimization pass is implemented in these two files
+src/ir/opt/mod.rs
+src/ir/opt/short_int_adj.rs
+## The logic to handle non-adjust operations in the constraint generation system
+src/target/r1cs/trans.rs
+```
```
sampsyo, line 24: Recall that your audience for this blog post is someone who is interested in compilers and your general topic but is not intimately familiar with the code of the CirC project. These filenames are unlikely to be meaningful to the readers. There is no need to include that level of code-level detail.

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
... zero-knowledge proof, multiparty computations, and more. To be more specific, it converts a program's source code into a set of mathematical constraints represented by vectors of variables. Optimization of such compilers primarily focus on reducing the size of the constraint system: having fewer constraints reduces the workload for creating proofs, and might also result in shorter proofs to verify. The typical formalism adopted by the majority of proof systems is rank-1 constraint system (R1CS). In an R1CS instance C, there exists a collection of constraints over a finite field F, often a prime p, or in other words, a circuit of additions and multiplications gates mod p. One significant cost of such system is introduced by arithmetic operations. Assume an underlying prime number p and operations such as multiplication done over p′, to obtain a correct result, a remainder operation is always applied on the result to adjust bitwidth and avoid overflow, which leads to a number of constraints that are at least equal to the bitwidth of the result, as it requires at least one split gate. Our project aims to apply a heuristic approach to optimize the translation of arithmetic operations.
```
sampsyo, line 27: Any particular reason to use curly braces around {C, ZoKrates, Circom} and {SMT, ILP, R1CS, MPC} instead of parentheses?

sampsyo, line 27: This paragraph might be a little more readable if you were to enable LaTeX mode and use math typesetting for the variables, like $C$, $F$, and $p$.

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+
+## Design
+Our optimization mainly focuses on addition and multiplication, which would typically increase the bitwidth of resulting element by 1 or double the bitwidth, respectively. We classify elements into two types
+- Elements that could be above the range and required to be returned within range: this includes program output, elements involved in bitwise operations and comparisons, operations like division or remainder, elements involved in memory operations, and so on.
+- Elements that could be above the range but are not required to be within range: This includes intermediate elements that undergoes multiplication and addition operations and does not meet any of the criterions above.
```
sampsyo, line 32: Hmm… I have tried quite a bit and I don't understand this. Maybe part of it is that I don't yet know what an "element" is? Do you just mean "variable" or "value" or something? And maybe it's because you haven't defined "the range". What range are we talking about here? I don't understand yet why one category is require to be "within range" and the other is not. Maybe this needs just a little more explanation, or perhaps an example or two?

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+
+The current system represents basic elements as bitvectors, and is designed to adjust their bitwidth after each arithmetic operation. Based on the classification of the elements above, our projects reduces the number of constraints in the system by greedily deciding when to adjust bitwidth for elements falls under the second category while ensuring no overflows can happen in later operations involving the elements. Or in other words, adjustments are only introduced if the next operation is going to result in an overflow.
```
sampsyo, line 34: When you say "if the next operation is going to result in an overflow", do you mean "when the next operation _might_ result in an overflow?" That is, are you detecting _dynamically_ when the specific values are guaranteed to overflow, or are you statically and conservatively avoiding all overflow cases?

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+
+Take this short program as an example:
+```
+def main(u32 a, u32 b) -> u32:
+    u32 c = a + b
+    u32 d = a + c
+    return c * d
+```
+
+The old compiler would adjust immediately at `a+b`, `a+c` and `c * d`. By applying our optimization, the system would treat c and d as intermediate elements and only adjust at c*d, since it is producing the final program output.
```
sampsyo, line 44: I think a major part of my confusion here is that I don't yet see what "adjust" means. Does it mean that we represent `c`, `d`, and the return value of `main` using wider types than `u32`? That seems a little odd because the source code implies that, for example, `c` must fit within 32 bits. The only semantics I am aware of out there for any programming language with fixed-width types like `u32` are:

* overflow wraps around
* overflow is a dynamic error
* overflow is undefined behavior

I'm not really familiar with any frontend language that says "overflow is represented precisely, and we ignore the fact that you declared the variable as a `u32` and represent it with an arbitrary-precision BigInt anyway." Is that the semantics of the language you're working with here?

I think I have missed the point and "adjust" means something more subtle than "use more bits to represent it." Can you explain in more detail what the "adjustment" consists of?

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+```
+def main(u32 a, u32 b) -> u32:
+    u32 c = a + b
+    u32 d = a + c
+    return c * d
+```
+
+The old compiler would adjust immediately at `a+b`, `a+c` and `c * d`. By applying our optimization, the system would treat c and d as intermediate elements and only adjust at c*d, since it is producing the final program output. 
+Without our optimization, the cost spent on adjusting the bitwidth will be: 33 constraint to adjust `a + b` and `a + c` respectively, and 64 to adjust `c * d`, which result in 130 constraints in total.
+In contrast, after our optimization, the cost spent on adjusting `a + b` and `a + c` would be 0, and 66 to adjust `c * d`, which result in 66 constraints in total.
```
sampsyo, line 46: Can you please make it a bit clearer _why_ it is safe to do the adjustment later? I think it's critically important to convince the reader at this point that what you're doing is semantically equivalent to the old strategy.

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+    u32 d = a + c
+    return c * d
+```
+
+The old compiler would adjust immediately at `a+b`, `a+c` and `c * d`. By applying our optimization, the system would treat c and d as intermediate elements and only adjust at c*d, since it is producing the final program output. 
+Without our optimization, the cost spent on adjusting the bitwidth will be: 33 constraint to adjust `a + b` and `a + c` respectively, and 64 to adjust `c * d`, which result in 130 constraints in total.
+In contrast, after our optimization, the cost spent on adjusting `a + b` and `a + c` would be 0, and 66 to adjust `c * d`, which result in 66 constraints in total. 
+
+## Implementation
+1. We create a new type of operations: `add_unadjust` and `mul_unadjust` for addition and multiplication to notify the constraint generation system not to adjust the bitwidth.
```
sampsyo, line 49: What are the semantics of `add_unadjust` and `mul_unadjust`? Do they overflow? What are their output types?

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+2. We implemented two passes to rewrite the operations. First pass is to classify elements in the system. We implemented an initial pass to scan the entire program and store elements requiring adjustments in a set S. Whenever we encounter an operation that definitely requires adjustment, we add the elements used in the operation to S. Then we implemented a rewrite pass to rewrite operations done on bitvectors not included in D to our newly introduced unadjust operation.
```
sampsyo, line 50: But how does it decide whether the operations need adjustments? How does the classification actually work, in terms of a program analysis?

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+
+## Implementation
+1. We create a new type of operations: `add_unadjust` and `mul_unadjust` for addition and multiplication to notify the constraint generation system not to adjust the bitwidth.
+2. We implemented two passes to rewrite the operations. First pass is to classify elements in the system. We implemented an initial pass to scan the entire program and store elements requiring adjustments in a set S. Whenever we encounter an operation that definitely requires adjustment, we add the elements used in the operation to S. Then we implemented a rewrite pass to rewrite operations done on bitvectors not included in D to our newly introduced unadjust operation. 
+3. We have also made changes to the constraint generation system to take advantage of the two new operations. In the original implementation, the constraint generation system will always adjust the bitwidth of the result of an operation. We have made changes to the system to not adjust `add_unadjust` and `mul_unadjust` operations. We also proved that such change will not impact the correctness of the system.
```
sampsyo, line 51: When you say "proved", what do you mean? Did you generate SMT constraints that assert the equivalence and discharge this obligation with a solver? Did you do a pen-and-paper proof (and if so, can you please include it in this blog post)? Did you use a proof assistant?

At the very least, if you have proven something about *all* programs, it is very important to include the formal statement of the theorem. I don't know if we need to see the complete proof, but it seems very important to see the statement that this proof is in service of.

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+
+## What were the hardest parts to get right
```
sampsyo, line 53: Please do not use section headings that recapitulate the bullet points from [the syllabus](https://www.cs.cornell.edu/courses/cs6120/2023fa/syllabus/#experience-report). Can you think of a section heading that summarizes the content of this section in a way that a reader would find useful?

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+The hardest part is to make sure the compiler works properly using our newly introduced operation. We encounter many bugs while implementing the optimization. For instance, before since the author always adjust the bitwidth after all types of operations, the bitwidth is fixed. However, now since an operation might be using unadjusted bitwidth the logic of calculating the bitwidth of the current element is more complicated and requires more steps to ensure an operation does not overflow.
```
sampsyo, line 54: Can you please include at least one example of such a bug? This sounds really interesting, and having some concrete evidence showing exactly what can go wrong would help people understand what was interesting about this project.

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+
+## Empirical evaluation
+We uses two types of tests. 
+1. We use the existing benchmarks in the circ repo to test whether the output of the optimized compiler is correct. Our implementation passed all the test cases in CirC.
```
sampsyo, line 58: circ -> CirC (please be consistent throughout)

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+2. To measure how well the proposed design improves the system, we use the number of r1cs constraints generated as the metrics. We runs our compiler and the original one over a set of zokrates programs in [CirC](https://github.com/circify/circ) and [Zombie](https://github.com/PepperSieve/Zombie/tree/master) and compares the number of constraints generated. Here is a shorted list of tests we run. The first line are the numbers of constraints with our flexible adjustment implementation and the second line is the result using original compiler.
+```
```
sampsyo, line 60: Could you format this as a table or as a bar chart? Simply listing the text output from a tool is pretty hard to digest.

content/blog/2023-12-11-circ-bitwidth-adjustment
```diff
@@ -0,0 +1,115 @@
...
+git clone https://github.com/collinzrj/circ
+git checkout bitwidth_adjustment
+```
+then `cd` into it,
+then run this command to build circ
+`cargo build --release --features r1cs,smt,zok,bellman --example circ`
+For proving and verifying you also need to build zk
+`cargo build --release --features r1cs,smt,zok,bellman --example zk`
+You can find them built in `./target/release/examples/zk` and `./target/release/examples/circ`
```
sampsyo, line 92: I think these operational instructions belong in the README for your repo. They're not that relevant to the notional reader for this blog post, who is someone who just wants to understand what your scientific contribution was.

sampsyo (changes requested):
Hello! Congratulations on finishing your project. I had some trouble understanding your blog post. I think there may be some important missing details around describing the underlying problem, your solution, and its correctness. Can you please clarify some of these things:

* What exactly does "adjustment" mean?
* Why is your different strategy correct?
* How does your program analysis work?
* What theorem did you prove?

I worry that this writeup may be understandable to someone who is already very familiar with CirC, but I hope it's possible to also make it understandable to someone who is only familiar with compilers topics in general.

sampsyo:
This PR has "timed out"—please let me know if you'd like to reopen it.

PR #438, opened 2023-12-13
[Blog] JITNIC

### LINKED ISSUE #393: Proposal: JITNIC
eBPF is an API and language that allows code to be installed dynamically into the kernel and run with various “hooks.” A verifier ensures that this code terminates and does not exhibit any unsafe behavior, such as writes to privileged memory locations. eBPF code installed can communicate with user space code using “maps” and a message-passing model. Historically, eBPF was developed for packet-processing, but due to its versatility can be used for more general applications as well (such as the amazing bpftrace tool). The eBPF language itself resembles a restricted subset of x86 and is executed by a VM that prevents runtime security violations.

With the emergence of software defined networking within the network core, switches have become more programmable, allowing for on the fly changes to switch behavior, without having to fabricate a new chip. This paradigm shift has led to new DSLs for specifying switch behavior and an entire ecosystem of tools and compilers for testing and running programs on switch hardware. A recent advancement in this direction is introducing programmability to the NIC, which is historically a specialized hardware unit for receiving and transmitting packets. Programmability in the NIC allows application and network level processing to be offloaded from the CPU to the NIC, where specialized hardware units can accelerate the computations.

Despite the specialized hardware for packet processing that exists in the NIC, eBPF programs cannot take advantage of it, as they currently only execute on CPU cores. (Insert some overused cliche about the end of Moore's law)
We aim to accelerate packet-processing and other eBPF computations for networked applications by building a tracing JIT that offloads program traces of straight-line code to a pipelined NIC, which contains “read-modify-write” memory stages and other specialized functional units. On receiving a packet, it first traverses the “slow path” through the CPU, where we compute the path conditions–predicates over memory locations and packet data–that determine when a packet will traverse that trace. Doing so “flattens” a branching program into a set of straight line traces. In a sense, we compute the “equivalence class” of packets that will cause any given trace to be executed. Then, we compile this trace to the bytecode executable by the NIC and install it there. Finally, we process this packet by executing the newly installed trace on the NIC; packets received in the future falling into the same equivalence class will also be processed by executing this trace.

All of this sounds super easy, right? 
Unfortunately, we anticipate some especially challenging problems, notably dealing with memory and consistency between the NIC and the CPU. For example, a trace executed on the pipelined NIC may only perform a number of memory operations bounded by the number of memory stages in the pipeline; any more would require falling back to a “slow path” on the CPU. While this packet is being processed on the CPU, we would have to stall the NIC pipeline to avoid potential read-after-write dependencies between the packet on the slow path and those next in the sequence. As a simplifying assumption, we will only deal with traces with a bounded number of memory operations and assume all packets can be processed on the NIC. To help keep this project tractable, we have created a timeline of milestones we intend to hit.

- eBPF program analysis 
- Identify traces
- Calculate path conditions 
- Formalize eBPF semantics: there are various references that vaguely describe the semantics of eBPF instructions, but it would be good to formalize the assumptions we make. Further, if we want to prove any correctness theorems, we will need formalized semantics.
- Finalize and implement the architecture of the pipeline itself. We can either write ourselves a custom Verilog implementation *shudder*, or adapt and modify an implementation of an existing eBPF pipeline developed a few years ago by one of Nate’s former students (hooray!)
- This also necessitates a kernel driver for interacting with the NIC, which thankfully already exists to some extent
- Decide upon a bytecode format (with semantics) the NIC will execute
- Write a compiler from eBPF to the decided upon bytecode format
- Come up with a coprocessor model for sharing memory and processing between the NIC and the CPU 

How will you empirically measure success?
We’re hoping to have something that at least kind of works. 
Jokes aside, there is a wealth of existing eBPF programs in the Linux kernel, which we will use to benchmark the performance of our implementation. As a baseline, we will execute these programs normally; then, we will execute these programs accelerated on the NIC and compare their performance.  We also plan on doing at least one case study of a “realistic” program and show how it gets compiled to the NIC to improve the performance.

**Team members:**
@bennyrubin @zachary-kent

zachary-kent:
With @bennyrubin.

Closes #393.


content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+    [extra]
+    bio = """
+      Zak Kent is a senior interested in programming languages and verification.
+      Benny Rubin is a junior interested in networking and systems.
+    """
+    [[extra.authors]]
+    name = "Zak Kent"
+    [[extra.authors]]
+    name = "Benny Rubin"
+    +++
```
sampsyo: I think you want to "outdent" this front matter? Not sure if Zola cares.

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Summary
+
+For this project, we implemented an eBPF JIT compiler to share eXpress Data Path (XDP) network programs between a host CPU and a smartNIC. The fundamental idea behind this design is to use runtime information to compile and run straight-line traces of eBPF code on a smartNIC. Straight line traces of code enable both compiler level and hardware level optimizations. The key challenge to this approach is to decide if a given trace can correctly run on a packet and to dynamically recompile a new trace if not. We develop the idea of a “path condition” - a predicate over fields of a packet. Each trace has a corresponding unique minimal path condition that is sufficient to determine the set of incoming packets that can be run on it. Finally, we evaluate our approach by looking at worst case and best case trace compilation rates, where fewer recompilations is better for performance. We find in practice that XDP programs require few traces installed on a NIC to cover the majority of incoming packets for realistic workloads.
```
sampsyo: > We develop the idea of a “path condition”

I might slightly rephrase this; it makes it sounds a bit like you invented the idea of "path conditions", which are of course an integral part of symbolic execution.

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+The eBPF hook we explored in our project is the XDP hook. XDP allows users to run programs on packets within the networking driver before all of the expensive transport layer processing, memory allocation, and data copy has occurred. These programs can read/write header fields, route packets out the Tx (transmit) datapath, let packets continue along the kernel networking stack, drop packets, and even store state with the use of eBPF maps, a restricted form of memory. XDP is especially powerful for programs where the common case is dropping or forwarding (i.e. filters, load-balancers), as the majority of packets will not be processed by the kernel networking stack.
```
sampsyo: This is incredibly minor, but I recommend always using a , after "e.g." or "i.e.". https://capra.cs.cornell.edu/styleguide/#egie

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+The use of a smartNIC can further accelerate this processing. A programmable smartNIC is a piece of networking hardware that operates as a NIC, but can be reprogrammed on the fly, rather than just implementing a fixed set of behaviors decided by the hardware vendor. By running an XDP program on a smartNIC, a packet fated to be dropped does not need to be DMA (Direct Memory Access) written to host memory, saving on end-to-end latency, and PCIe and memory bandwidth. We call this the Extra Express eXpress Data path. 
+An important piece of prior work is hXDP, which implements an eBPF interpreter on an fpga smartNIC, allowing it to run arbitrary XDP programs. While this is a great innovation, due to the complex and non-specialized hardware, they were not able to measure any significant throughput increases. This approach also does not allow for a program to be “split” between the CPU and the NIC – instead, all of the XDP program processing is happening on the NIC.
```
sampsyo: fpga -> FPGA

sampsyo: Please include a link to the hXDP paper.

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+- A packet experiences a “miss”.
+  In this case, it must be sent to the compiler on the host CPU to be traced. 
+  The JIT compiler will trace the program, generating the path condition – a set of conditions on the packet header for a packet to take the same path in the program– along with the trace itself. The CPU does not need to send the packet back to the NIC, as it can execute the XDP action itself. Finally, the CPU installs the path condition and trace into the trace cache, ejecting an element, using any preferred cache replacement policy, if necessary.
+
+For this project, we focus on the compiler and implement the “hardware” NIC as an eBPF simulator. For this reason, we do not go into detail about the hardware implementation of certain features. Our system is built in ~1400 of Ocaml code, along with some tools for parsing out the eBPF byte code of an object file and turning PCAP (packet capture) files into byte streams. There are also about ~450 lines of an experimental Rust implementation that is now defunct.
```
sampsyo: Ocaml -> OCaml

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+We are actively working on different memory consistency models – sharing memory between a smartNIC and CPU gets very complicated, but for the sake of the system we implemented for this class, we assume the programs are stateless and there are no memory reads/writes outside of the input packet data. This also removes the problem of having memory reads in path conditions for memory locations that are also written to (you can quickly see how complicated this would get…). 
+
+## Case study: Ipv6 Filter
+
+The following C code is an XDP program that filters all non-ipv6 traffic. This will call the XDP_DROP action on any packets that do not match the ipv6 protocol in their ethernet header. Any ipv6 packets, or malformed packets that do not pass the bounds check will be passed onto the linux kernel networking stack.
```
sampsyo: linux -> Linux

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+	return XDP_PASS;
+
+if (eth->protocol == 0x86dd) /* bytes of eth type ipv6 */
+	return XDP_PASS;
+else 
+	return XDP_DROP;
+}  
+```
+
+For the sake of simplicity, and to avoid introducing eBPF assembly, we will keep this example at the level of C. Note every C instruction has an analogous few instructions in eBPF.
```
sampsyo: By "C instruction", perhaps you mean "C operation"?

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+For the sake of simplicity, and to avoid introducing eBPF assembly, we will keep this example at the level of C. Note every C instruction has an analogous few instructions in eBPF. 
+
+Starting from the beginning, let’s assume a packet arrives that has an ipv4 ethernet header. Taking the execution model from the diagram, this will be a cold miss as the cache is empty. The JIT compiler on the CPU will trace the program with the packet as input and produce the following trace:
+```C
+void *data_end = ctx->data_end; /* pointer to where the packet ends in memory */
+struct ethhdr *eth = ctx->data;
+return XDP_DROP; 
+```
+The path condition will be `eth + sizeof(*eth) <= data_end && eth->protocol != 0x86dd` – notice that when tracing conditions just become part of the path condition. (Also note: in C this is `eth->protocol` but in eBPF this becomes a lookup to an offset within the packet. Similarly sizeof(*eth) is a constant and eth and data_end are given as input). At this point, the CPU will drop the packet.
```
sampsyo: Probably add Markdown code backticks around `sizeof(*eth)`. Ditto for `eth` and `data_end`.

sampsyo: One important thing I couldn't quite understand from this example, which I think you are getting at partially here: this path condition expression, as written, contains variables that are defined in the trace. To deal with this in general, I can think of a couple of options:

* evaluate the path condition after running the trace, but before making a drop/forward decision
* "slice" the computation of all of those variables, so the path condition is a self-contained thing (at the cost of possibly recomputing some values when we do eventually run the trace)
* somehow interleaving the trace and condition evaluation so we can make a decision as soon as all the "ingredients" for the path condition are available

Maybe it would be good to be explicit about exactly what the current system does?

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+This emphasizes the importance of tuning the cache size - which could potentially dynamically change depending on the program. It also shows that the replacement policy can potentially affect the probability of experiencing cache thrashing. In practice, we find XDP programs to be quite short and not have many conditions, meaning a small trace cache can fit most possible traces. 
+
+Despite the limitations in conducting a thorough evaluation of real system performance, we conclude with a key observation: even a simple stateless filtering program can – with few traces compiled onto the NIC – significantly reduce end to end latency (by avoiding DMA and the kernel networking stack), reduce contention on the PCIE and memory buses, and save on CPU cycles.
```
sampsyo: PCIE -> PCIe

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+This emphasizes the importance of tuning the cache size - which could potentially dynamically change depending on the program. It also shows that the replacement policy can potentially affect the probability of experiencing cache thrashing. In practice, we find XDP programs to be quite short and not have many conditions, meaning a small trace cache can fit most possible traces.
```
sampsyo: > In practice, we find XDP programs to be quite short and not have many conditions, meaning a small trace cache can fit most possible traces.

Can you say a little bit more about what evidence you gathered for this assertion? How many programs did you try, and where did you find them?

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Optimizations and Future work
+
+Currently, our system does not very closely resemble a realistic hardware implementation. There are several steps that remain in order to port our software simulator to a real NIC. First and foremost, we must design the instruction set that will be executed by the pipeline; this is where we will have maximum flexibility in optimizing performance. At this point, we are considering how to compile eBPF to a very long instruction width (VLIW) architecture, which exploits the instruction-level parallelism in a program by deciding statically which instructions can be executed in parallel. Many of the memory accesses in eBPF programs–those to packet headers, for example, can be easily parallelized. We will also have to nail down the exact structure of the pipeline, possibly testing different layouts to observe which is most effective. This will require deciding how many memory units are in the pipeline, for example.
```
sampsyo: "programs–those to packet headers, for example, can" -> "programs—those to packet headers, for example—can"

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+The lowering to a VLIW instruction set is also why we decided against focusing on optimizing eBPF itself; the eBPF bytecode produced by clang is already tightly optimized, and additional optimizations at this level would likely not be exceedingly fruitful.
+
+Additionally, we will have to examine the behavior of more eBPF programs to decide upon how to best deal with the difficulties resulting from a realistic distributed memory model, as opposed to the unrealistic shared memory model we currently operate under. Coordination between the host and NIC to ensure memory consistency would be extremely costly, and like outweigh the benefits of JIT’ing code at all.
```
sampsyo: like -> likely?

content/blog/2023-12-11-jitnic/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+## GAI Acknowledgement
+
+We used GitHub Copilot through the project, which often did not produce particularly helpful suggestions. It was useful, however, for producing some of the boilerplate needed to write module signatures both in interface and source files. It was also useful for writing documentation. The autocomplete suggestions were at times humorously incorrect, like the following pyramid:
+
+![A humorously incorrect pyramid-shaped autocomplete suggestion from copilot](pyramid.png)
```
sampsyo, line 109: 😂 Stellar work, Copilot!

sampsyo (changes requested):
Sounds like a super cool direction! It's awesome that you have a working tracer already. I'm also really intrigued by your idea to consider a VLIW pipeline for executing traces; this does indeed seem like a pretty good fit for this kind of straight-line code.

Y'all have identified the right set of problems for future work—the mutable-state issue is definitely a serious one. I admit I am hopeful that it wouldn't be too disastrous to keep all the state on the NIC and require the slow path (host tracing) to do remote accesses to that state. This would require stalling the fast-path NIC pipeline while the host does its tracing, but something like this kind of stalling seems inevitable. By keeping all the state in one place, you "define away" all consistency problems. Of course, this only works for situations where the state is small enough to fit on the NIC! A workload that requires frequent random state accesses seems unlikely to benefit anyway, though. So maybe the only real case that would pose a serious problem for this is one where, like, 99% of state accesses go to a small working set, but you still have to account for the 1% of cases that randomly access a much larger cold footprint.

Anyway, one thing I would love to hear more about—if there is something to report—is just about what kinds of programs you've examined. It sounds like you have done experiments beyond the one ipv6 filter you use as an example and have concluded that traces can be compact—what XDP programs did you observe to draw that conclusion?

Great work overall!

content/blog/2023-12-11-jitnic/index.md
zachary-kent: Added it as a reference at the bottom! I couldn't find an easy way to link the in-text citation to the corresponding reference, though.

content/blog/2023-12-11-jitnic/index.md
bennyrubin: Good point, we should have included this from the start. I've updated the post. In summary, we don't currently support BPF maps, so we couldn't run many programs outside of stateless filtering ones or contrived examples. To reach these conclusions, we manually inspected XDP programs that are included in the Linux kernel, instead of running them in our system. I made this clear and discussed the downsides.

content/blog/2023-12-11-jitnic/index.md
bennyrubin: We have clarified this in the writeup. In summary, we currently produce path conditions that are evaluated from offsets within the packet - in our system we just store memory offset values and then calculate the path condition based on those. If a path condition depends some address that isn't a constant offset, we abort tracing and run the concrete interpreter to execute the program on the packet. We discuss a possible approach for producing traces of this type in the post.

sampsyo: Got it; that makes sense!!

content/blog/2023-12-11-jitnic/index.md
sampsyo: You can do it with a Markdown footnote[^foo] or just link the text to the [ACM DL entry](https://dl.acm.org/doi/abs/10.1145/3543668) for the paper.

[^foo]: like this

content/blog/2023-12-11-jitnic/index.md
zachary-kent: To clarify, we *do* support non-constant offsets; we abort tracing only when a branch depends on the value stored in a memory location that has already been written to.

content/blog/2023-12-11-jitnic/index.md
zachary-kent: Yeah it ended up being easier to just link the text. Thanks!

sampsyo (approved):
Looks great! Nice work again!!

PR #439, opened 2023-12-18
[Blog] SELF post

AliceSzzze:
SO SO SORRY that we are submitting this this late!!! It completely slipped our minds that we had to write a blog after the presentation. Hope late is better than never?

@collinzrj @bennyrubin


sampsyo:
No worries; thanks for putting this together. I'll take a closer look soon.

PR #464, opened 2025-01-23
Add Ernest to GC & Reference Counting discussion

ngernest:


sampsyo:
Looks like @gabizon103 and @parthsarkar17 got there first. Would the three of you be open to collaborating on this? If so, can you please resolve the conflict to include all 3 names?

ngernest:
Happy to collab! Fixed the merge conflict

sampsyo:
Excellent. I'm going to speculate that @gabizon103 and @parthsarkar17 are OK with this too; please post again if not.

PR #461, opened 2025-01-22
Add Mariia to Fancy Memory Management

mariasoroka:

PR #466, opened 2025-01-23
Add Edmund Lam to Interactive Verification

UnsignedByte:


sampsyo:
Hey @UnsignedByte—I think we already have 3 people for this paper (if you include #461). Would you be open to switching to a different presentation that only has 1 or 0 people assigned so far?

UnsignedByte:
@sampsyo Done! I took Interactive Verification, which is empty for now :)

sampsyo:
Awesome. Thanks.

PR #470, opened 2025-01-24
Co lead Dynamic Languages Discussion

smd21:
@adityanathan  Would it be okay if I co-lead this discussion with you?


adityanathan:
Sure!

sampsyo:
Oops, looks like there was a race condition with @ananyagoenka (in #468). I'm going to assume the 3 of you are cool with working together!

PR #472, opened 2025-01-26
Add Sam as superoptimization leader

samuelbreckenridge:


sampsyo:
Sounds good, as long as @mb64 is down to collaborate!

mb64:
Sounds good to me!

sampsyo:
Great!

PR #474, opened 2025-01-27
Add Allen to profiling paper

aw578:


sampsyo:
Sounds good, as long as @scober is down to collaborate!

scober:
I would love to collaborate!

sampsyo:
Excellent!

PR #486, opened 2025-02-07
Swap discussion leads due to conflict

emmanueljs1:
Spoke with the people leading the discussion for "Automatic Verification" and @devv64 was available for a swap!


sampsyo:
Great; thanks for engineering the swap!

PR #489, opened 2025-02-27
Remove Emmanuel from discussion lead list

emmanueljs1:
This class is really fun but my semester got a bit too busy!


sampsyo:
Sorry to hear about your busy semester, but thanks for updating the schedule!

PR #490, opened 2025-02-28
Fix typo in lesson 6 (I think)

scober:
I came across what I think are two typos in the notes for lesson 6. In the spirit of the open source course, I thought I would open a PR!

Let me know if I got the intended text wrong. I'm happy to change it!


sampsyo:
Thanks!! I know it sounds weird, but I actually usually treat "semantics" as a mass noun (not a plural, even though it sounds like one). So I'll keep "their semantics depends."

scober:
That sounds like it is probably the correct semantics for "semantics". I wouldn't say "the commutativity semantic of addition" so I guess "semantics" is not really referring to a group of individual semantics even if it sounds like it.

PR #497, opened 2025-03-18
Update ML Pointers Link in Alias Analysis Lesson

arnavm30:
Current ML pointers link leads to a 404 error page. I couldn't find the exact same content on the OCaml site. 
Some other possible replacements:
https://web.archive.org/web/20210426110256/https://ocaml.org/learn/tutorials/pointers.html
https://caml.inria.fr/resources/doc/guides/pointers.en.html


sampsyo:
Thanks for catching this! Too bad that old site seems to be offline; I liked its discussion about pointers existing everywhere but just being implicit. I found [the contents on GitHub](https://github.com/ocaml/v2.ocaml.org/blob/8414244e502ca04e014ff772a224f9cb31472708/site/learn/tutorials/pointers.md), so maybe I'll link there.

PR #527, opened 2025-05-04
feat(blog): Add tracing paper blog

ethanuppal:
We apologize again for the late due date and appreciate being allowed to submit late.

We tried to omit background and other information we considered "redundant" (e.g. background) to avoid filler; I am not even quite happy with how concise we are, but it will suffice. I include this notice in case you judge the writing to be too short, in which case we will look harder for more relevant information that is simultaneously not precise details of the paper.

### DIFF
--- a/content/blog/2025-05-03-blog.md
+++ b/content/blog/2025-05-03-blog.md
@@ -1,0 +1,51 @@
++++
+title = "TraceMonkey: A Tracing Just-In-Time Compiler for JavaScript"
+[extra]
+bio = """
+  Ethan Uppal Cornell CS '27
+  Tean Lai Cornell CS '26
+"""
+[[extra.authors]]
+name = "Tean Lai"
+
+[[extra.authors]]
+name = "Ethan Uppal"
++++
+
+This blog is about the paper “Trace-based Just-in-Time Type Specialization for Dynamic Languages” by Gal et al.
+To understand the paper, it helps to have some familiarity with JavaScript, Just-In-Time (JIT) compilers, and the differences between tracing-based and method-based JIT compilers. 
+
+Interpreting JavaScript can be quite slow.
+However, efficient JavaScript engines are effectively required for any modern browser given how much JavaScript gets run on even the seemingly simplest websites.
+Consequently, modern-day browsers have to use JIT compilers to get maximal performance gains.
+Since JavaScript is a dynamically-typed language, a traditional compiler must emit code that can handle the different possible types during runtime, producing less efficient code overall; JIT compilers are not bound by this limitation.
+But how can you make an effective JIT for JavaScript?
+
+The authors of the paper expect that programs spend most of their time in hot loops, and that hot loops are generally type-stable — this is their key idea.
+In the paper, they describe the development of a JIT compiler built around this expectation called TraceMonkey.
+
+At a high level, TraceMonkey works by first monitoring for hot loops by using counters.
+Once hot loops are identified, it checks for type stability throughout the loop.
+If it is type stable, then it can emit corresponding code that is closer to efficiency to that of a statically-typed language.
+It can also detect violations of this assumption and resolve them at runtime.
+
+---
+
+TraceMonkey typically outperformed the original SpiderMonkey interpreter on their provided benchmarks, but the improvements were largely inconsistent. Benchmarks focused on bitwise operations had nearly 25x speedups; others showed no improvements. TraceMonkey performed especially badly on programs with a lot of recursion because only loops were tracked.
+
+---
+
+TraceMonkey is only one of several different JIT implementations used in SpiderMonkey, the name of Mozilla’s JavaScript engine used to power Firefox.
+Although TraceMonkey was great at tracing loops and generating efficient code, it (for obvious or aforementioned reasons) was weak in processing programs that did not have as many loops or used a lot of recursion.
+In particular, TraceMonkey had no method-based compiling and could suffer from trace explosion.
+For these and other reasons, TraceMonkey has long been deprecated, but it directly influenced its successors: JagerMonkey, WarpMonkey, and IonMonkey. 
+Each of these compilers use more method-based JIT compiling while still incorporating trace-based compilation in a hybrid approach. Similarly, JavaScript JIT compilers for other popular browsers — Chrome and Safari — also used a method-based approach. Is this to say trace-based JITs are less effective than method-based JITs? Not quite; the ideas presented in TraceMonkey still live on in several ways. In addition to the previously stated fact that the newer Firefox JITs still use tracing, the predominant JIT compilers for Python and Lua (PyPy and LuaJIT respectively) all heavily rely on tracing, with LuaJIT considered being one of the fastest JITs ever. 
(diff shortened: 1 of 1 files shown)


content/blog/2025-05-03-blog.md
```diff
@@ -0,0 +1,51 @@
++++
+title = "TraceMonkey: A Tracing Just-In-Time Compiler for JavaScript"
+[extra]
+bio = """
+  Ethan Uppal Cornell CS '27
+  Tean Lai Cornell CS '26
```
sampsyo, line 6: This is the place where you're supposed to include some brief information about yourself, maybe some links, etc. Like [this](https://www.cs.cornell.edu/courses/cs6120/2023fa/blog/lowering/):
Screenshot 2025-05-05 at 4 43 10 PM

content/blog/2025-05-03-blog.md
```diff
@@ -0,0 +1,51 @@
...
+"""
+[[extra.authors]]
+name = "Tean Lai"
+
+[[extra.authors]]
+name = "Ethan Uppal"
++++
+
+This blog is about the paper “Trace-based Just-in-Time Type Specialization for Dynamic Languages” by Gal et al.
```
sampsyo, line 15: Please make this a hyperlink to the paper.

content/blog/2025-05-03-blog.md
```diff
@@ -0,0 +1,51 @@
...
+Since JavaScript is a dynamically-typed language, a traditional compiler must emit code that can handle the different possible types during runtime, producing less efficient code overall; JIT compilers are not bound by this limitation.
+But how can you make an effective JIT for JavaScript?
+
+The authors of the paper expect that programs spend most of their time in hot loops, and that hot loops are generally type-stable — this is their key idea.
+In the paper, they describe the development of a JIT compiler built around this expectation called TraceMonkey.
+
+At a high level, TraceMonkey works by first monitoring for hot loops by using counters.
+Once hot loops are identified, it checks for type stability throughout the loop.
+If it is type stable, then it can emit corresponding code that is closer to efficiency to that of a statically-typed language.
+It can also detect violations of this assumption and resolve them at runtime.
```
sampsyo, line 30: I don't think this is an accurate description of the technique from the paper. This description makes it sound like the unit of code generation is a loop, i.e., we find the hot loops and compile those _chunks of static code_ by making assumptions about the types. But that's not the spirit of what's going on with tracing: the unit of code generation is a _trace:_ we _start tracing_ when we find a hot loop, and the tracing is what determines the code to generate. So it's not the code of the loop that we start with—it's a _dynamic trace through_ the execution of _a single iteration_ of the loop.

It seems important to capture this central idea if you want to summarize the key contribution of the paper. Concepts like guards and deoptimization flow from this central idea, so it will be hard to discuss further technical points without this foundation.

content/blog/2025-05-03-blog.md
```diff
@@ -0,0 +1,51 @@
...
+In particular, TraceMonkey had no method-based compiling and could suffer from trace explosion.
+For these and other reasons, TraceMonkey has long been deprecated, but it directly influenced its successors: JagerMonkey, WarpMonkey, and IonMonkey. 
+Each of these compilers use more method-based JIT compiling while still incorporating trace-based compilation in a hybrid approach. Similarly, JavaScript JIT compilers for other popular browsers — Chrome and Safari — also used a method-based approach. Is this to say trace-based JITs are less effective than method-based JITs? Not quite; the ideas presented in TraceMonkey still live on in several ways. In addition to the previously stated fact that the newer Firefox JITs still use tracing, the predominant JIT compilers for Python and Lua (PyPy and LuaJIT respectively) all heavily rely on tracing, with LuaJIT considered being one of the fastest JITs ever. 
+It may be helpful to go into more detail about the transition to focusing on method tracing. 
+Trace-based compiling excels when control flow is simpler, but on the web, control flow can be highly irregular for several reasons: dynamic input, user interaction, etc.
```
sampsyo, line 44: > Trace-based compiling excels when control flow is simpler

I'm not sure I agree with this assertion, or at least I think you haven't presented enough evidence for the assertion to be convincing. There is a lot of branching but those branches are "stable," then it seems like tracing is a wonderful way to automatically eliminate a lot of it in the common case.

content/blog/2025-05-03-blog.md
```diff
@@ -0,0 +1,51 @@
...
+It may be helpful to go into more detail about the transition to focusing on method tracing. 
+Trace-based compiling excels when control flow is simpler, but on the web, control flow can be highly irregular for several reasons: dynamic input, user interaction, etc. 
+Method-based compiling is better at handling these irregularities. 
+One point brought up in discussion was that, although TraceMonkey might not compile every type of program well, because there are some types of programs that benefit extremely (like computation-heavy programs with simple loops as shown with the bitop benchmarks results), we enable more types of websites that can be written. 
+
+> Fun fact: TraceMonkey is recognized as the first JavaScript JIT compiler, beating V8 by a couple of months. 
+> In general, we have seen the introduction of JavaScript JIT compilers completely revolutionize the web. 
+> Their introduction meant games can be run in browsers, we could have crazier animations, etc. 
+> It was the start of an era where the web is plagued by insane amounts of JavaScript.
```
sampsyo, line 51: It seems like this fact probably belongs at the very _beginning_ of the blog post, when you're introducing TraceMonkey for the first time?

sampsyo (changes requested):
Hi, Ethan and Tean—thanks for getting this started. I like the points that you have made here (with one exception where I think you've mischaracterized the technical contribution).

However, you're right that this post is very short. It just doesn't say much about the paper yet! I don't think you should add back more background; you probably have enough of that already. But surely you have more thoughts _about_ the paper that you can include—or if you don't, maybe you can steal some from the discussion thread? It seems like there were a lot of opinions, extrapolations, etc. that came up during class, and the idea here is to convey some of that thinking. Can you think of anything else interesting to relate about this paper?

sampsyo:
Hi, @ethanuppal and @tean-lai: do you plan on finishing up this blog post? I'd love to publish it if you do!

ethanuppal:
Yes, I will handle these soon.

sampsyo:
Hi, y'all! Just wanted to check in about wrapping up this post—it would be great to get it merged.

ethanuppal:
Looks like "soon" will have to be redefined to ~3 months now. Not sure how this slipped my mind.

ethanuppal:
So just to confirm - you would like us to add content from the corresponding discussion thread? If memory serves me well, I don't think Tean or I had much thoughts beyond what we wrote on TraceMonkey specifically (although we did have many thoughts on tracing JITs, the evolution of SpiderMonkey, etc., but these weren't relevant to a post specifically about TraceMonkey).

sampsyo:
Any combination of those two things (ideas from the discussion thread, or tenuously-connected thoughts about JITs in general even if they don't have much to do with TraceMonkey specifically) would be good!

ethanuppal:
> Any combination of those two things (ideas from the discussion thread, or tenuously-connected thoughts about JITs in general even if they don't have much to do with TraceMonkey specifically) would be good!

I see, sounds good

PR #530, opened 2025-05-06
add typed based alias analysis blog post

gerardogtn:
Blog post for TBAA with @mt-xing  and @arnavm30.

### DIFF
--- a/content/blog/2025-05-06-typed-based-alias-analysis.md
+++ b/content/blog/2025-05-06-typed-based-alias-analysis.md
@@ -1,0 +1,114 @@
++++
+title = "Typed Based Alias Analysis"
+[extra]
+latex = false
+bio = """
+Michael Xing is a Masters of Science student studying CS at Cornell and a Software Engineer at Microsoft. When not writing code, he enjoys travelling the world, riding roller coasters, and playing the piano. And video games.
+
+Gerardo Teruel is a Masters of Engineering student at Cornell and used to be a Software Engineer at Etsy and Wizeline. When not thinking about software, he enjoys cooking, biking, and hiking with his dog Cardamom.
+
+Arnav Muthiayen is an undergraduate student studying ECE and CS at Cornell. In his free time, he enjoys playing basketball and watching films.
+"""
+[[extra.authors]]
+name = "Michael Xing"
+[[extra.authors]]
+name = "Gerardo Teruel"
+[[extra.authors]]
+name = "Arnav Muthiayen"
++++
+
+One of the largest challenges facing an optimizing compiler is the fact that multiple variables in a program may refer to the same underlying memory, known as aliasing. The possibility of aliasing results in many compiler optimizations becoming unsafe, resulting in the compiler being unable to apply many seemingly reasonable optimizations unless it is able to conclusively determine that two variables are in fact not aliases.
+This week’s paper, “Type-Based Alias Analysis” by Amer Diwan, Kathryn S. McKinley, and J. Eliot B. Moss, proposes a method of using type information and other statically available information like field names in order to better find variable pairs that are guaranteed not to be aliases.
+
+
+## Background
+
+Many reasonable compiler optimizations require that the variables they operate upon do not alias the same piece of memory. For example, the following code:
+```C
+bool myfn(int* a, int* b) {
+	*a = 1;
+	*b = 2;
+	return *a == *b;
+}
+```
+It would be convenient if the compiler could optimize the return value to be the constant false. Especially if a and b were dead after this operation, the entire function could be inlined to simply be false. However, if the pointers a and b both refer to the same memory location, then the result would actually be true, since \*a and \*b would fetch the same value.
+For this reason, compiler optimizations depend upon alias analysis, where the compiler attempts to statically determine which variables could alias which others. Specifically, the compiler is generally looking for cases where two variables must not be aliases, as those are where opportunities for optimization typically arise.
+A naive approach to alias analysis could involve a simple dataflow analysis, checking for what possible memory locations a variable could point to. However, due to the conservative nature of this analysis, it necessarily results in a large number of false positives.
+
+## Summary
+
+Diwan, McKinley, and Moss propose three new, progressively more sophisticated methods of performing alias analysis on statically typed programming languages.
+First, the authors use type information to compare two fields. If the types of two fields are not the same, and they also have no subtypes in common, then by the semantics of most languages, they cannot alias the same memory (at least, not without triggering undefined behavior).
+Second, the authors extend the analysis with field names. Even for two objects of the same type, accessing two different fields on two different objects cannot be aliases of the same memory.
+Third, the authors extend the analysis again by also performing a (flow-insensitive) analysis of assignments, to check if a common subtype is ever assigned to variables of two different parent types. If not, then any variables of those two parent types cannot be aliases either, despite sharing a subtype.
+The authors show that their later analyses are within 2.5% of a theoretically perfect analysis when applied to Redundant Load Elimination. One of the main contributions they emphasize is how they also performed a dynamic and limit analysis, in addition to the static analysis.
+
+## Annotations, Rich Types, and Alias Analysis
+
+
+One of the aspects of alias analysis is that it underestimates the true count of aliasing in a program. In TBAA it might be the case that we write a function that may alias but that in the total usages of the function it never actually aliases. In this scenario, we lose the opportunity to optimize some code, but we accept this tradeoff in order to keep the analysis fast and correct. In particular, TBAA is really powerful as it only relies on language constructs to perform the analysis so it doesn’t require special modifications to the source code to improve the results of the alias analysis. But this same approach opens more doors for optimizations if the source language has a richer type system or if there is a way for programmers to provide hints to the compiler. An example is the restrict keyword in C, where the programmers inform the compiler there will be no aliasing between parameters: 
+```C
+//source code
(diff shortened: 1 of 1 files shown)


content/blog/2025-05-06-typed-based-alias-analysis.md
```diff
@@ -0,0 +1,114 @@
...
+[[extra.authors]]
+name = "Michael Xing"
+[[extra.authors]]
+name = "Gerardo Teruel"
+[[extra.authors]]
+name = "Arnav Muthiayen"
++++
+
+One of the largest challenges facing an optimizing compiler is the fact that multiple variables in a program may refer to the same underlying memory, known as aliasing. The possibility of aliasing results in many compiler optimizations becoming unsafe, resulting in the compiler being unable to apply many seemingly reasonable optimizations unless it is able to conclusively determine that two variables are in fact not aliases.
+This week’s paper, “Type-Based Alias Analysis” by Amer Diwan, Kathryn S. McKinley, and J. Eliot B. Moss, proposes a method of using type information and other statically available information like field names in order to better find variable pairs that are guaranteed not to be aliases.
```
sampsyo, line 21: Can you please include a hyperlink from the title to the paper itself?

content/blog/2025-05-06-typed-based-alias-analysis.md
```diff
@@ -0,0 +1,114 @@
...
+
+Many reasonable compiler optimizations require that the variables they operate upon do not alias the same piece of memory. For example, the following code:
+```C
+bool myfn(int* a, int* b) {
+	*a = 1;
+	*b = 2;
+	return *a == *b;
+}
+```
+It would be convenient if the compiler could optimize the return value to be the constant false. Especially if a and b were dead after this operation, the entire function could be inlined to simply be false. However, if the pointers a and b both refer to the same memory location, then the result would actually be true, since \*a and \*b would fetch the same value.
```
sampsyo, line 34: Markdown tip: instead of escaping the * character with a backslash, an easier (and more readable) way would be to use `a backtick-delimited code span` around all variable names and C expressions.

content/blog/2025-05-06-typed-based-alias-analysis.md
```diff
@@ -0,0 +1,114 @@
...
+For this reason, compiler optimizations depend upon alias analysis, where the compiler attempts to statically determine which variables could alias which others. Specifically, the compiler is generally looking for cases where two variables must not be aliases, as those are where opportunities for optimization typically arise.
+A naive approach to alias analysis could involve a simple dataflow analysis, checking for what possible memory locations a variable could point to. However, due to the conservative nature of this analysis, it necessarily results in a large number of false positives.
```
sampsyo, line 36: It’s also not necessarily the most efficient thing in the world, especially if you have a complicated heap model. (I mention this because one claimed advantage of TBAA is its simplicity/efficiency.)

content/blog/2025-05-06-typed-based-alias-analysis.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Annotations, Rich Types, and Alias Analysis
+
+
+One of the aspects of alias analysis is that it underestimates the true count of aliasing in a program. In TBAA it might be the case that we write a function that may alias but that in the total usages of the function it never actually aliases. In this scenario, we lose the opportunity to optimize some code, but we accept this tradeoff in order to keep the analysis fast and correct. In particular, TBAA is really powerful as it only relies on language constructs to perform the analysis so it doesn’t require special modifications to the source code to improve the results of the alias analysis. But this same approach opens more doors for optimizations if the source language has a richer type system or if there is a way for programmers to provide hints to the compiler. An example is the restrict keyword in C, where the programmers inform the compiler there will be no aliasing between parameters:
```
sampsyo, line 49: > underestimates the true count of aliasing in a program

Could you possibly tweak this wording to make it clear which “direction” you’re claiming? Maybe you mean “underestimates the number of pairs of variables that will alias at run time,” or maybe the opposite, like “underestimates the opportunities for optimization”? The current wording sounds like the former, but the latter seems more correct and aligns with the rest of the paragraph…

sampsyo, line 49: > TBAA is really powerful as it only relies on language constructs to perform the analysis so it doesn’t require special modifications to the source code

Hmm… I would say that this is true of all alias analyses. As in, every alias analysis in the world works with the language as-is and does not requires special aliasing annotations. Stuff like `restrict` isn’t an alias analysis; it’s an “escape hatch” from the analysis.

content/blog/2025-05-06-typed-based-alias-analysis.md
```diff
@@ -0,0 +1,114 @@
...
+	*a = 1;
+	*b = 2;
+	return *a == *b;
+}
+
+// optimized version
+bool myfn(std::unique_ptr<int> a, std::unique_ptr<int> b) {
+	return false;
+}
+```
```
sampsyo, line 80: Just to confirm: is this a hypothetical, or did you actually try this in a real C++ compiler? Hypothetical is totally fine, but if you happen to have tried it out, that would be nice to mention explicitly (along with the compiler name & version you used).

content/blog/2025-05-06-typed-based-alias-analysis.md
```diff
@@ -0,0 +1,114 @@
...
+ 	MyInt b1 = new Record(2)
+	return a1.value = b1.value
+}
+// optimized version
+boolean myfn(MyInt a, MyInt b) {
+	return false;
+}
+```
+
+So although the original paper did not consider annotations or richer type constructs, the idea can easily be extended to provide even better results for alias analysis.
```
sampsyo, line 98: I like this section! It’s great to get somewhat deeper into what language features make TBAA more or less powerful.

content/blog/2025-05-06-typed-based-alias-analysis.md
```diff
@@ -0,0 +1,114 @@
...
+boolean myfn(MyInt a, MyInt b) {
+	return false;
+}
+```
+
+So although the original paper did not consider annotations or richer type constructs, the idea can easily be extended to provide even better results for alias analysis. 
+
+## Measuring an Upper Bound
+
+Despite the large number of compiler optimization techniques, it is often difficult to gauge how impactful an optimization will be in the performance of an arbitrary program; the most common approach is to use a set of benchmarks execute them a number of times in the original source  and in the optimized source do a pairwise comparison (usually in ratios) and aggregate the gains using the arithmetic or harmonic mean. This approach is not perfect, but makes it easy to compare different optimizations and is standard in literature. Diwan, McKinley and Moss used a completely different approach for TBAA, one that relies on establishing an upper bound.
```
sampsyo, line 102: > the most common approach is to use a set of benchmarks execute them a number of times in the original source  and in the optimized source do a pairwise comparison

Seems like there are some missing commas in here?

content/blog/2025-05-06-typed-based-alias-analysis.md
```diff
@@ -0,0 +1,114 @@
...
...ard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) allowed compilers to safely assume pointers of incompatible types wouldn’t alias, treating any violations as undefined behavior. There are some exceptions, though, such as character pointers. Compilers such as GCC and LLVM exploit this rule to enable more aggressive optimization. GCC enables [-fstrict-aliasing](https://gcc.gnu.org/onlinedocs/gcc-7.5.0/gnat_ugn/Optimization-and-Strict-Aliasing.html) by default at higher optimization levels like -O2 and LLVM/Clang enables a similar [TBAA metadata system](https://llvm.org/docs/LangRef.html#tbaa-metadata) by default at all optimization levels. A notable drawback is that legacy code that violated these strict aliasing rules could break unexpectedly, prompting the creation and usage of the [-fno-strict-aliasing](https://gcc.gnu.org/onlinedocs/gcc-7.5.0/gnat_ugn/Optimization-and-Strict-Aliasing.html) flag to preserve the previous behavior; moreover, Clang provides [TypeSanitizer](https://clang.llvm.org/docs/TypeSanitizer.html) (enabled with -fsanitize=type flag), a run-time library that uses TBAA metadata and dynamic instrumentation to detect strict type aliasing violations.
```
sampsyo, line 108: I hadn’t heard of TypeSanitizer; that’s very cool!

sampsyo (changes requested):
Really nice work on this; you’ve done a great job highlighting and contextualizing the main ideas. I just have a few minor suggestions—when you have a chance to address them, please “request a review” in GitHub to let me know to take another look.

sampsyo:
Hi, @gerardogtn @mt-xing @arnavm30! I would love to publish your blog post. Can you please wrap up the revisions discussed above so I can hit the green button?

sampsyo:
Just one last reminder about the above, @gerardogtn @mt-xing @arnavm30—if you don't want to wrap up the small additions above, we can just close this PR.

mt-xing:
Hi Professor. I don't have access to Gerardo's branch, so I've made a separate PR from my own fork that attempts to address the comments. When you have time, would you be able to review the updated PR #550 instead? In particular, [this commit](https://github.com/sampsyo/cs6120/pull/550/commits/23248a1945c88bae06274489e47ff3fd6f036f65) shows the diff of all the changes. Thanks.

sampsyo:
Awesome; can do! Thanks!

sampsyo:
Superseded by #550.

PR #533, opened 2025-05-13
Add mse63 Final Project

### LINKED ISSUE #513: Project Proposal: Bril2C
**What will you do?**
I will try to build a decompiler, to translate Bril to a language like C. The goal is that for any bril program, I would be able to call `bril2c` on it, and create a C program which will always have the same functionality as the bril program. A decompiler could be useful to reverse engineer programs to make them more understandable and human-readable.

**How will you do it?**
I suspect the simplest implementation won't be too difficult. There are only a finite number of types of bril instructions, and each one can be roughly mapped to a C statement. Things like memory management and phi nodes might require a little more tinkering to replicate the functionality, but all-in-all it shouldn't be too much. However, such an implementation would be hardly any better than just reading the bril directly. The goal should be to identify patterns that can be represented as higher-level C constructs like if statements, while loops or for loops, and use those instead of simple goto statements. Additionally, a variable that is only written to once and read from once can and should be removed by putting it in-line, such as `a = b + c; d = a  * e;` could become `d = (b + c) * e;`

Correctness should matter above all else. I'll create a C program from every bril program in the tests in the bril git repo, and compile and run each C program and make sure it produces the same output as the bril program.

Once correctness is verified, I could try to measure progress by minimizing the number of C tokens in the generated C programs (at the risk of falling into Goodhart's Law), with the idea that shorter code is "simpler" and hence more "human-readable".

**Team members:**
@mse63

mse63:
Closes #513

### DIFF
--- a/content/blog/2025-05-bril2c.md
+++ b/content/blog/2025-05-bril2c.md
@@ -1,0 +1,158 @@
++++
+title = "Bril2C"
+[extra]
+bio = """
+  Mahmoud is an MEng student enrolled in CS6120 as of Spring 2025.
+"""
+[[extra.authors]]
+name = "Mahmoud Elsharawy Sampson"
++++
+
+# Bril2C
+This project is intended to translate code from Bril (Big Red Intermediate Language) to C. Bril is an educational intermediate language used in the CS6120 course at Cornell University. As an intermediate language, it is low-level, which would make it difficult to translate to a high-level language. However, C, with features such as manual memory management and goto statements, makes it a natural choice. In my initial proposal, I thought creating a complete and correct translation from Bril to C would be trivial, but indeed quite a bit of hacking was needed, due to limitations of C not placed on Bril.
+
+## Implementation
+Bril2C is written in Rust, using the "bril-rs" package. Most of the actual translation work happens in a single trait, `Crep`, which contains a single function `crep(self) -> String`. Each piece of Bril implements this function with how it translates itself to C, calling `crep` on smaller pieces, effectively inducting on the structure of the program. For example, `crep` for a program will, along with other things, call `crep` for each function in that program, which will call `crep` for each instruction, which will call `crep` for each operation and variable. In this way, adding additional features of Bril into C becomes as easy as implementing a single function for a specific type.
+
+As it turns out, quite some fiddling is required to get even a majority of Bril code working. Almost all of the benchmarks have `main` function which takes in arguments, something not possible in `C`. We get around this by creating our own `main` function in C, which parses arguments from the command line, and passes them into a copy that's supposed to represent the Bril's main function, called `main_f`. Taking an example, `sum-bits.bril`, with a main function that takes in a single integer argument, this is represented as:
+```C
+int main(int argc, char *argv[]) {
+  int64_t input = atoi(argv[1]);
+  main_f(input);
+  return 0;
+}
+```
+This way the resulting program can take in arguments from the command line for the main function in the same way as `brili`.
+
+Additionally, Bril's print statement isn't type-annotated, meaning some form of polymorphism is needed in C. We accomplish this by creating our own generic print macro, allowing us to call `print` as easily as it is called in Bril:
+```C
+#define print(x)                                                               
+  _Generic((x),
+      int64_t: printf("%" PRId64 " ", x),
+      uint8_t: printf("%s ", (x) ? "true" : "false"),
+      double: printf("%.17f ", x))
+```
+For ease of programming, each function declares all of the variables used in it at the beginning of the function After that, it's a "simple" translation from Bril instructions to C statements. Therefore, the following function in Bril:
+```
+@mod(dividend : int, divisor : int) : int {
+  quotient : int = div dividend divisor;
+  two : int = const 2;
+  prod : int = mul two quotient;
+  diff : int = sub dividend prod;
+  ret diff;
+}
+```
+becomes this C function:
+```C
+int64_t mod_f(int64_t dividend_, int64_t divisor_) {
+  int64_t quotient_;
+  int64_t two_;
+  int64_t prod_;
+  int64_t diff_;
+  quotient_ = dividend_ / divisor_;
+  two_ = 2;
+  prod_ = two_ * quotient_;
+  diff_ = dividend_ - prod_;
+  return diff_;
+}
+```
+At this point, you may have noticed the additional characters in each variable name and suffix name. These are inserted to prevent bril registers from being confused with C keywords, which is surprisingly popular in Bril code, especially for words like `if` and `continue`.
+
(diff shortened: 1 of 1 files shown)


content/blog/2025-05-bril2c.md
```diff
@@ -0,0 +1,158 @@
...
+title = "Bril2C"
+[extra]
+bio = """
+  Mahmoud is an MEng student enrolled in CS6120 as of Spring 2025.
+"""
+[[extra.authors]]
+name = "Mahmoud Elsharawy Sampson"
++++
+
+# Bril2C
```
sampsyo: No need to repeat the post title; it's already rendered by default.

content/blog/2025-05-bril2c.md
```diff
@@ -0,0 +1,158 @@
...
+This project is intended to translate code from Bril (Big Red Intermediate Language) to C. Bril is an educational intermediate language used in the CS6120 course at Cornell University. As an intermediate language, it is low-level, which would make it difficult to translate to a high-level language. However, C, with features such as manual memory management and goto statements, makes it a natural choice. In my initial proposal, I thought creating a complete and correct translation from Bril to C would be trivial, but indeed quite a bit of hacking was needed, due to limitations of C not placed on Bril.
```
sampsyo: It might be helpful to add a hyperlink when you introduce Bril.

content/blog/2025-05-bril2c.md
```diff
@@ -0,0 +1,158 @@
...
+
+As it turns out, quite some fiddling is required to get even a majority of Bril code working. Almost all of the benchmarks have `main` function which takes in arguments, something not possible in `C`. We get around this by creating our own `main` function in C, which parses arguments from the command line, and passes them into a copy that's supposed to represent the Bril's main function, called `main_f`. Taking an example, `sum-bits.bril`, with a main function that takes in a single integer argument, this is represented as:
```
sampsyo: No need for backticks around `C`, the name of the programming language.

content/blog/2025-05-bril2c.md
```diff
@@ -0,0 +1,158 @@
...
+This way the resulting program can take in arguments from the command line for the main function in the same way as `brili`.
+
+Additionally, Bril's print statement isn't type-annotated, meaning some form of polymorphism is needed in C. We accomplish this by creating our own generic print macro, allowing us to call `print` as easily as it is called in Bril:
+```C
+#define print(x)                                                               
+  _Generic((x),
+      int64_t: printf("%" PRId64 " ", x),
+      uint8_t: printf("%s ", (x) ? "true" : "false"),
+      double: printf("%.17f ", x))
+```
```
sampsyo, line 34: Interesting approach to supporting `print`! FWIW, for _both_ the issue of `main` arguments and for `print`, there is a way that you could consider simplifying code generation by reusing an existing "runtime library" for one of the other Bril-to-native compilers, like this:
https://github.com/sampsyo/bril/blob/main/brilift/rt.c

That's a library of little C functions that do things like parse each command-line argument or print something of a particular type. You can use the Bril type information to select the appropriate `_bril_print_*` function.

content/blog/2025-05-bril2c.md
```diff
@@ -0,0 +1,158 @@
...
+  int64_t prod_;
+  int64_t diff_;
+  quotient_ = dividend_ / divisor_;
+  two_ = 2;
+  prod_ = two_ * quotient_;
+  diff_ = dividend_ - prod_;
+  return diff_;
+}
+```
+At this point, you may have noticed the additional characters in each variable name and suffix name. These are inserted to prevent bril registers from being confused with C keywords, which is surprisingly popular in Bril code, especially for words like `if` and `continue`.
```
sampsyo: bril -> Bril

content/blog/2025-05-bril2c.md
```diff
@@ -0,0 +1,158 @@
...
+
+Additionally, Bril's print statement isn't type-annotated, meaning some form of polymorphism is needed in C. We accomplish this by creating our own generic print macro, allowing us to call `print` as easily as it is called in Bril:
+```C
+#define print(x)                                                               
+  _Generic((x),
+      int64_t: printf("%" PRId64 " ", x),
+      uint8_t: printf("%s ", (x) ? "true" : "false"),
+      double: printf("%.17f ", x))
+```
+For ease of programming, each function declares all of the variables used in it at the beginning of the function After that, it's a "simple" translation from Bril instructions to C statements. Therefore, the following function in Bril:
```
sampsyo:
```suggestion
For ease of programming, each function declares all of the variables used in it at the beginning of the function. After that, it's a "simple" translation from Bril instructions to C statements. Therefore, the following function in Bril:
```

content/blog/2025-05-bril2c.md
```diff
@@ -0,0 +1,158 @@
...
+}
+int main(int argc, char *argv[]) {
+  int64_t input = atoi(argv[1]);
+  main_f(input);
+  return 0;
+}
+```
+
+## Evaluation
+Creating a program to translate a single Bril program into C is easy. Creating a program to translate _every_ bril program into C is much more difficult. I tested my code with a simple script, `test.sh` which would run every `.bril` file in `benchmarks`, and compare its output with the translated, compiled, and run code from `Bril2C`. Many times, I thought I had completed my project, only to discover a Bril program that broke my seemingly correct implementation, such as by naming a register the same name as a function, or by using the same register with two different types in two different functions. I did manage to resist the temptation of simply declaring that I would only consider a subset of Bril programs with "nice" formatting, however, and powered on, handling edge case after edge case. After enough bug-fixing, however, I was able to successfully produce a correct program for every single Bril program in the benchmark (as shown in `test_results.txt`).
```
sampsyo: Cool! Can you add a link to that results file, `test_results.txt`?

content/blog/2025-05-bril2c.md
```diff
@@ -0,0 +1,158 @@
++++
+title = "Bril2C"
+[extra]
+bio = """
+  Mahmoud is an MEng student enrolled in CS6120 as of Spring 2025.
+"""
+[[extra.authors]]
+name = "Mahmoud Elsharawy Sampson"
```
sampsyo: Do you have a secret last name?

sampsyo (changes requested):
Thanks for the writeup. It sounds like you eventually took my advice from #513 and did not attempt any of the "decompiler" aspects—focusing instead on the straightforward "C backend" aspects. Which makes sense! Even a simple CFG recovery strategy would be complicated to do.

Would you be interested in contributing your bril2c compiler to the Bril monorepo? I can imagine it being useful to other folks down the line!

mse63:
I just pushed a new commit to fix the issues.
I can add bril2c to the monorepo, if you think it'll be useful. What directory should I put it under?

sampsyo:
Looks great; thanks!

Why don't we start with a top-level `bril2c` directory and see how well that works?

PR #534, opened 2025-05-13
Write Bril MLIR project blog post

### LINKED ISSUE #500: Project Proposal: Register Allocation in Bril
**What will you do?**
We will implement and compare two different register allocation approaches in Bril. The first is the linear scan register allocation algorithm, which is a fast algorithm which assigns registers based on live intervals. The second approach is an implementation of SSA-based register allocation that simultaneously converts out of SSA and allocates registers, as described in [this paper](https://compilers.cs.uni-saarland.de/papers/ssara.pdf). If time permits, we may also explore a traditional non-SSA graph coloring approach, or we may consider extensions to the linear scan algorithm such as second-chance bin packing.

**How will you do it?**
We will first need to familiarize ourselves with the two allocation algorithms; linear scan should be straightforward, but implementing the SSA approach will require carefully reading through the above paper to understand how to handle details like coalescing with phi-nodes. To simulate the existence of a limited set of registers, we will use the suggested variable naming scheme (i.e., `r1`, `r2`), with all others being spilled.

To implement the linear scan algorithm, we can utilize our existing dataflow analysis structure for live variable analysis, and we will greedily allocate registers using their live intervals, then insert memory instructions for spilled variables. To implement the SSA-based approach, we will need to construct the SSA interference graph, make spilling decisions, and exit SSA with the described coalescing heuristic.

Once the two allocation strategies are implemented, we will modify the Bril interpreter to support register allocation. Specifically, we will need to track and report key metrics during execution, such as the number of register spills, the total number of dynamic instructions executed, and the number of memory accesses. Since it is easy to varying the number of available registers by modifying the set of variable names which are considered registers, we can also compare the approaches across different register file sizes. We will also test our implementations by constructing a number of small programs which test specific edge case behaviors, such as high register pressure and complex control flow, as well as by running our register allocation over all the existing Bril benchmarks to check functional correctness.

One aspect of correctness is the semantically equivalent transformation of Bril programs via our register allocation implementations. Furthermore, to compare the performance of the different allocation schemes, we will measure the number of spilled registers, the number of dynamic instructions executed, and the number of memory instructions executed across the entire Bril benchmark suite. This will allow us to analyze aspects such as the memory overhead added due to spilling for each approach.

**Team members:**
@bryantpark04 @dhan0779 @lisarli

lisarli:
@bryantpark04 @dhan0779
Closes #500

### DIFF
--- a/content/blog/2025-05-13-bril-mlir.md
+++ b/content/blog/2025-05-13-bril-mlir.md
@@ -15,51 +15,174 @@ This blog post is about Bril2MLIR, a project which translates between Bril JSON 
 ## Overview
 
 Code: 
-[https://github.com/bryantpark04/bril/tree/main/bril-mlir](https://github.com/bryantpark04/bril/tree/main/bril-mlir)
-[https://github.com/bryantpark04/llvm-project/tree/main/mlir/examples/brilc](https://github.com/bryantpark04/llvm-project/tree/main/mlir/examples/brilc)
+[Bril2MLIR](https://github.com/bryantpark04/llvm-project/tree/main/mlir/examples/brilc)
+[Tests and supporting code](https://github.com/bryantpark04/bril/tree/main/bril-mlir)
 
 In our original project proposal, we wanted to implement and compare two different register allocation approaches in Bril, namely the linear scan register allocation algorithm and an SSA-based register allocation described in [this paper](https://compilers.cs.uni-saarland.de/papers/ssara.pdf). Without too much effort, we were able to familiarize ourselves with the linear scan and come up with our [own implementation](https://github.com/lisarli/cs6120-tasks/blob/main/finalproject/register_allocation.cpp), utilizing our existing dataflow analysis and maintaining spilled variables (using `id` instructions before they were operated on). However, it turned out that the SSA-based register allocation mentioned in the paper was significantly harder to implement and ultimately we scrapped this project idea as a whole.
 
-We decided to pivot our project to implementing translation of SSA-based Bril JSON programs into MLIR and vice versa. MLIR is a powerful and extensible compiler infrastructure which supports custom dialects and transformations across different levels of abstraction, making it a popular and flexible system. This pivot allowed us to explore the transformation from the lightweight BRIL IR to the more involved MLIR form. To do this, we created a dialect for Bril consisting of custom operations for instructions such as `id`, `nop`, and `print`, and we built on existing MLIR dialects such as `arith` for common operations like `add` and `sub`. In doing so, we were able to represent all of the core Bril instructions, including all arithmetic and control flow. The first direction of our translation was from Bril JSON to MLIR, which involved parsing the input program into a traversable representation, constructing the corresponding MLIR operations, and emitting a completed `ModuleOp` with correctly assembled functions and basic blocks. 
+We decided to pivot our project to implementing translation of SSA-based Bril JSON programs into MLIR and vice versa. [MLIR](http://mlir.llvm.org) is a powerful and extensible compiler infrastructure which supports custom dialects and transformations across different levels of abstraction, making it a popular and flexible system. This pivot allowed us to explore the transformation from the lightweight Bril IR to the more involved MLIR form. To do this, we created a dialect for Bril consisting of custom operations for instructions such as `id`, `nop`, and `print`, and we built on existing MLIR dialects such as `arith` for common operations like `add` and `sub`. In doing so, we were able to represent all of the core Bril instructions, including all arithmetic and control flow. The first direction of our translation was from Bril JSON to MLIR, which involved parsing the input program into a traversable representation, constructing the corresponding MLIR operations, and emitting a completed `ModuleOp` with correctly assembled functions and basic blocks. 
 
 To organize our project, we created a fork of the llvm-project which was placed inside the `bril` repository in a new directory called `bril-mlir`. This structure allowed us to build our implementation with the tools provided by the LLVM Project in combination with the existing `bril` benchmarks and commands. We created a `Dialect` class to register our custom Bril MLIR operations, with definitions in a `.td` file. We used the MLIR C++ API to develop the MLIR generator and translator.
 
 ## Bril to MLIR
 
-The [MLIR Toy tutorial](https://mlir.llvm.org/docs/Tutorials/Toy/Ch-2/) provides a walkthrough of building a simple language using MLIR with parsing and AST construction. Using this as a starting point, we adapted a similar structure to create a dialect for Bril by defining Bril-specific operations in TableGen, implementing the dialect in C++, and extending the MLIR generation logic to emit those custom operations. For many of the core operations like arithmetic (add, mul, etc…), control flow, and comparisons, we leveraged existing MLIR dialects like `arith` and `cf`. These dialects already provided the correct semantics and functionality for our use cases, allowing us to focus our custom dialect on features unique to Bril. While creating our own MLIR dialect for all Bril operations could have been more “proper”, it didn’t have a significant impact on the MLIR construction or the implementation for translation back to Bril. Here’s a small example of the MLIR conversion for integer arithmetic operations:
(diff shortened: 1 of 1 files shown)


content/blog/2025-05-13-bril-mlir.md
```diff
@@ -0,0 +1,88 @@
...
+name = "Lisa Li"
++++
+
+This blog post is about Bril2MLIR, a project which translates between Bril JSON and a custom Bril MLIR dialect.
+
+## Overview
+
+Code: 
+[https://github.com/bryantpark04/bril/tree/main/bril-mlir](https://github.com/bryantpark04/bril/tree/main/bril-mlir)
+[https://github.com/bryantpark04/llvm-project/tree/main/mlir/examples/brilc](https://github.com/bryantpark04/llvm-project/tree/main/mlir/examples/brilc)
```
sampsyo: May I suggest using a complete sentence for these links instead of just the URLs? Like, you could link the phrases "Bril2MLIR" and "a custom Bril MLIR dialect" above.

content/blog/2025-05-13-bril-mlir.md
```diff
@@ -0,0 +1,88 @@
...
+
+We decided to pivot our project to implementing translation of SSA-based Bril JSON programs into MLIR and vice versa. MLIR is a powerful and extensible compiler infrastructure which supports custom dialects and transformations across different levels of abstraction, making it a popular and flexible system. This pivot allowed us to explore the transformation from the lightweight BRIL IR to the more involved MLIR form. To do this, we created a dialect for Bril consisting of custom operations for instructions such as `id`, `nop`, and `print`, and we built on existing MLIR dialects such as `arith` for common operations like `add` and `sub`. In doing so, we were able to represent all of the core Bril instructions, including all arithmetic and control flow. The first direction of our translation was from Bril JSON to MLIR, which involved parsing the input program into a traversable representation, constructing the corresponding MLIR operations, and emitting a completed `ModuleOp` with correctly assembled functions and basic blocks.
```
sampsyo: Consider adding a link to MLIR when you first introduce it.

sampsyo: BRIL -> Bril

content/blog/2025-05-13-bril-mlir.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Bril to MLIR
+
+The [MLIR Toy tutorial](https://mlir.llvm.org/docs/Tutorials/Toy/Ch-2/) provides a walkthrough of building a simple language using MLIR with parsing and AST construction. Using this as a starting point, we adapted a similar structure to create a dialect for Bril by defining Bril-specific operations in TableGen, implementing the dialect in C++, and extending the MLIR generation logic to emit those custom operations. For many of the core operations like arithmetic (add, mul, etc…), control flow, and comparisons, we leveraged existing MLIR dialects like `arith` and `cf`. These dialects already provided the correct semantics and functionality for our use cases, allowing us to focus our custom dialect on features unique to Bril. While creating our own MLIR dialect for all Bril operations could have been more “proper”, it didn’t have a significant impact on the MLIR construction or the implementation for translation back to Bril. Here’s a small example of the MLIR conversion for integer arithmetic operations:
```
sampsyo: > While creating our own MLIR dialect for all Bril operations could have been more “proper”,

What do you mean by this? Please expand on what you think is missing.

content/blog/2025-05-13-bril-mlir.md
```diff
@@ -0,0 +1,88 @@
...
+    let results = (outs);
+    
+    let assemblyFormat = "attr-dict";
+    }
+
+    llvm::LogicalResult mlirGenInstrNop(...){
+        builder.create<Nop>(loc);
+        return mlir::success();
+    }
+
```
sampsyo, line 188: Instead of listing the source code for the translation (which is quite straightforward), I think it would be more useful to see what the dialect looks like. Can you show a couple of smallish examples to compare the Bril text format with the MLIR text format for the equivalent program? I think that would illustrate more about your design decisions here.

content/blog/2025-05-13-bril-mlir.md
```diff
@@ -0,0 +1,88 @@
...
+The final output is a complete Bril JSON program with labeled blocks, typed instructions, and SSA-correct variable naming. This approach ensures fidelity to the original semantics and enables round-tripping from Bril → MLIR → Bril for debugging, testing, or further analysis.
+
+## Challenges
+One difficult aspect of this project was properly converting Phi nodes in Bril into SSA form in MLIR. In particular, we used the old version of SSA which utilized `phi` instructions as opposed to `get` and `set`. A `phi` instruction consists of the source labels and variables for each predecessor, but in MLIR, this functionality is represented using block arguments. Predecessors provide the value of each of the block arguments before entering the block containing the Phi node. To correctly represent this behavior, we needed to add block arguments to MLIR blocks containing Phi nodes, and we needed to identify predecessor blocks in which to insert arguments for branches.
```
sampsyo, line 334: Can you maybe show at least one example of the conversion? (Comparing the original Bril text form to the MLIR equivalent.)

content/blog/2025-05-13-bril-mlir.md
```diff
@@ -0,0 +1,88 @@
...
+
+Another difficult aspect was correctly implementing control flow translation from Bril JSON into the MLIR block representation. In particular, we needed to devise a scheme to map Bril labels into MLIR blocks, which we then used as the targets for jumps and branches. We also needed to ensure that the targets passed in to MLIR `cf.br` and `cf.cond_br` were correct, as we initially ran into difficulties identifying the correct labels since we did not properly construct the block to label mapping.
```
sampsyo: > we needed to devise a scheme

Please tell us what that scheme was.

content/blog/2025-05-13-bril-mlir.md
```diff
@@ -0,0 +1,88 @@
...
+
+After some investigation, we found that all of the benchmarks we manually tested had segmentation faults during the transformation into MLIR. Due to the difficulty we faced in debugging these segmentation faults, we eventually gave up on trying to fix them, and moved on to transforming MLIR programs back into Bril JSON. Interestingly, we couldn’t find an easy pattern in the segfault-causing benchmarks. We initially hypothesized that function calls may have been the issue, but we found programs without any functions other than main to also segmentation fault.
```
sampsyo: Please explain more. This is maybe the most important part of your evaluation: understand what about your project was successful and what was not. The 43 benchmarks that failed this test should indicate something about the limitations of your current dialect and translation.

Unfortunately, "the crash was because of a segfault" is too generic to understand anything about what went wrong. Even knowing whether the crash came from your code (where?) or the MLIR framework itself (also where?) would say a lot about what is wrong with your translation. I think it's important to do a little bit more work here to know _something_ about what's not working.

sampsyo (changes requested):
Hi, everybody—I'm sorry to hear that you had to abandon your proposal from #500. It sounds like this project taught you something about MLIR, but it ended up with some significant limitations: it relied on the "old SSA" form and therefore inherited the flaws in its implementation, and (even with that restriction) the round-trip fails for many benchmarks for reasons you have not fully explained yet.

To wrap up this project, can you please at least look a little closer into the failures? I think it's important to understand, even at a very high level, what has gone wrong with your current implementation of the translation.

content/blog/2025-05-13-bril-mlir.md
bryantpark04: It was just a map haha - updated the post for clarity

content/blog/2025-05-13-bril-mlir.md
bryantpark04: Found the causes of the errors and fixed them! All benchmarks pass now

content/blog/2025-05-13-bril-mlir.md
bryantpark04: Added an example, let us know if it should be shortened to just one function!

sampsyo:
Cool; it's great that you got things working for the full benchmark suite! I think the long code examples are fine; I hope they help illustrate how the dialect works.

PR #535, opened 2025-05-13
Add flat bril project blog post

### LINKED ISSUE #501: Project Proposal: A Flattened Representation for Bril
**What will you do? How will you do it?**
We will design a flattened representation for Bril!

**Background:** Typically, the way to implement an interpreter is to create an explicit AST data type, but this requires allocating AST nodes on the heap. Instead, as demonstrated in [this blogpost](https://www.cs.cornell.edu/~asampson/blog/flattening.html) by Adrian, we can flatten the AST into an array, i.e. pack all AST nodes into one single contiguous array, and refer to children in the AST using array indices (as opposed to pointers). This has enormous performance benefits!

**Task:** In light of this, we aim to design a flattened representation for Bril. This would have 2 components:
1. Infrastructure to convert existing Bril JSON files to/from our flattened format (+ round-trip tests to make sure this works)
2. An alternate Bril interpreter that operates directly on the flattened data structure (as opposed to the existing one `brili`, which has to parse JSON)

Since all the AST nodes are stored within one contiguous array, the in-memory & disk representations are the same! To serialize, we can just use Rust's `zerocopy` library to convert the array into a blob of bytes that can be written to disk directly (no need for pretty-printing!) To deserialize, we can just use `mmap(2)` it (no need to parse json!) We plan on following the approach outlined in [another blogpost](https://www.cs.cornell.edu/~asampson/blog/flatgfa.html) by Adrian for turning our flattened Bril representation into a file format using `zerocopy`. 

Some foreseen challenges are storing strings for variable names and handling instructions w/ different no. of operands, as in both of these cases we deal with variable-length data. For instance we will probably want to use a separate array to store arguments. Additionally, for simplicity, our initial plan is to flatten the non-SSA version of Bril rather than trying to adapt our data structures to exploit SSA form. 

We plan on implementing our project in Rust for 3 reasons: 1.) to maintain consistency with Adrian’s blogposts, 2.) to take advantage of useful Rust libraries like `zerocopy` and 3.) to make our implementation performant (and comparable with the existing Rust interpreter `brilirs`). Our collective experience with Rust is limited (only one of the 3 of us has experience writing Rust), but we’re happy to learn Rust as part of this project. (If you have ideas on any alternate languages we could use, we’d be happy to hear them! To make the comparison of our interpreter with the existing interpreter fair, it seems like we would either have to use TypeScript or Rust, since the two existing Bril interpreters are written in those languages.)

We will empirically measure success in two components: the correctness of our implementation (i.e. that transformed programs behave the same way as the original programs) and the performance benefit of our implementation. To quantify the performance benefit, we plan on comparing the CPU wall clock time & memory usage of our interpreter vs the standard one (although we’re not sure if these are the best metrics). We also plan on comparing the file size of our file format (that we get from converting the array to bytes) vs the default JSON representation for Bril programs. 

We plan to evaluate our implementation first on a hand-crafted set of microbenchmarks (e.g. random generation of Bril programs à la Adrian’s blogposts), then on the `/core` benchmarks suite in bril. Specifically, for testing the correctness of our implementation, we will perform round-trip tests converting JSON files to/from our flattened format and back, checking that these JSON files remain the same. 

**Team members:**
@ngernest, @samuelbreckenridge, @katherinewu312

samuelbreckenridge:
Closes #501

@ngernest 
@katherinewu312


content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+        op_type: Type,
+    }
+```
+
+`Vec`s and `String`s in Rust are both implemented using pointers, so the representation is clearly not flat even before considering that functions contain pointers to instructions and programs contain pointers to functions. Our goal was to adopt the same approach of flattening data structures and packing them into a single contiguous array for Bril. The resulting representation of Bril programs is then a flat file containing these arrays that can be directly mapped into memory and interpreted, analogous to the approach laid out in [another blogpost](https://www.cs.cornell.edu/~asampson/blog/flatgfa.html) by Adrian. This requires two main components:
+
+1. Infrastructure to convert existing Bril JSON files to/from our flattened format
+2. An alternate Bril interpreter that operates directly on the flattened data structure (as opposed to the existing one brili, which has to parse JSON)
+
+Our complete implementation can be found at https://github.com/ngernest/flat-bril/tree/main
```
sampsyo: Consider integrating this link into the text, like "our complete implantation is [open source on GitHub][…]".

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Design
+
+### Flat data structures
+
+A key design decision was what the flattened data structures would look like. We clearly need similar fields to those of the original representation, but we need a strategy for flattening `Vec`s and `String`s and other non-flat types. Instead of storing variables, labels and function names as Strings and Vec<String>s we aggregate all the names referenced throughout a function into three contiguous arrays of bytes that are stored in our function representation. Then instead of using Strings, our instruction representation stores pairs of indices that can be used to extract the relevant name from the function-level contiguous array. One subtlety here is that if we want to use a Vec, we cannot just index into the array containing all the names, we would not know where the boundaries between each name are! We need to use an intermediary array that stores indices into the top-level name array.
```
sampsyo: You probably want some more code backticks when you refer to Rust types. For example:

> as Strings and Vec<String>s

->

> as `Strings` and `Vec<String>`s

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+
+For instance, in a function with three variable names v1, v2 and v3, the function-level variable array will look like v1v2v3. If an instruction has two arguments v1 and v2, our representation stores a pair of indices into an intermediate array that itself contains pairs of indices. A lookup would proceed as (0,1) → (0,0), (1,1) → v1, v2.
```
sampsyo: > A lookup would proceed as (0,1) → (0,0), (1,1) → v1, v2

Not sure I understand this sentence… are (0, 0) and (1, 1) supposed to be pairs of indices in the name buffer? Assuming half-open intervals, wouldn't that be (0, 2) and (2, 4)? Or, with closed intervals, (0, 1) and (2, 3)?

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+    pub var_store: Vec<u8>,
+    pub args_idxes_store: Vec<(u32, u32)>, // Intermediate array: indexes into var_store
+    pub labels_idxes_store: Vec<(u32, u32)>, // Intermediate array: indexes into labels_store
+    pub labels_store: Vec<u8>,
+    pub funcs_store: Vec<u8>,
+    pub instrs: Vec<Instr>,
+}
+```
+
+Although our function representation does contain Vec s to enable construction, these will be flattened to slices once the entire Function has been created
```
sampsyo:
```suggestion
Although our function representation does contain `Vec`s to enable construction, these will be flattened to slices once the entire Function has been created.
```

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Implementation
+
+### Zerocopy
+
+In order to facilitate conversion between slices of bytes and our flat data structures we used the zerocopy crate. In practice this meant adding the IntoBytes and FromBytes traits to our flat representations, and specifying their byte representations using the repr attribute. It ended up being quite challenging to get this working. Zerocopy was quite finicky about what was allowable in a struct using the zerocopy traits. We ended up needing to create new “extra-flat” versions of many of our data structures to get this to work and experimenting with different repr options. For instance we discovered that zerocopy could not convert pairs of `u32`s, or `Option`s to bytes, so we needed to create a new `I32Pair` struct that itself implemented the zerocopy traits (`I32` as opposed to `u32` because we used -1 to represent the case where the `Option` is None). We probably could have used these extra-flat data structures everywhere, rather than converting between multiple versions, but since we had written most of our flattening logic using the original data structures we decided to stick with using new extra-flat versions.
```
sampsyo: Please use a hyperlink to the zerocopy crate. And probably also to the Rust documentation for `repr`.

sampsyo: By `I32`, do you mean `i32`?

sampsyo: About the need to avoid the use of `Option`: one alternative, at least in the case of index intervals, would be to consider the interval `(x, x)` for any `x` to be empty. (That is, with half-open intervals, when start == end, then the length is zero—so that's a way to represent "no items here.")

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+    pub labels_idxes_store: &'a [I32Pair],
+    pub labels_store: &'a [u8],
+    pub funcs_store: &'a [u8],
+    pub instrs: &'a [FlatInstr],
+}
+```
+
+Once we finally made zerocopy happy about all our data structures we were able to convert seamlessly between the in memory representations and bytes which we could directly mmap to disk, enabling the flat file format we envisioned as one of our goals.
+
+When deserializing from our flat file format back to our flat data structures we ran into a very nasty bug. On some programs we would get alignment errors when attempting to convert bytes to our data structures, but we could fix/break programs just by changing the variable names! We eventually realized that a program would crash if the total bytes used to represent all variable names (or labels, or functions) was not a multiple of 4! This was because our top-level arrays just stored all names contiguously, and so would be unaligned unless the total bytes to represent all names was a multiple of 4. To fix this we just padded all of these top level arrays with null bytes.
```
sampsyo: That does indeed sound nasty!

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Interpreting flat Bril
+
+Once we had finished implementing our infrastructure for flat Bril representations, we implemented an interpreter that operated directly on the flat data structures. We built this interpreter from the ground up rather than trying to adapt the existing Rust interpreter. This was mostly straightforward, and we were able to model our logic after the Typescript interpreter. The main challenges were introduced by needing to keep track of which array each pair of indices in our flat data structures was referring to. We had a few bugs caused by assuming that some indices referred to the top-level name arrays, when in fact we needed to go through an intermediate array. In hindsight we could have done a better job with our naming conventions to avoid this.
```
sampsyo: Typescript -> TypeScript

sampsyo: [idle comment; no actionable change] Honestly, in my experience, this problem where you mix up the association between indices and arrays (and therefore use an index to access some array where that index is meaningless) is one of the bigger frustrations I've had when programming in this style. A little type safety could help a lot here…

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Evaluation
+
+For our evaluation, we decided to test flat bril on 70 core bril benchmarks. To check the correctness of our implementation, we used Turnt to verify that all benchmarks using our flat bril interpreter returned the same result as that of the reference Brili interpreter, for which we were successful. Additionally, to check the correctness of our infrastructure converting JSON files to/from our flattened format, we manually checked that the final json output converted back matched that of the original. To measure the performance impacts, we did the following using Hyperfine:
```
sampsyo: bril -> Bril
json -> JSON

sampsyo: Consider adding hyperlinks to Turnt, the reference interpreter, and Hyperfine.

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+
+1. Measured the CPU wall clock time for the flat bril, brili typescript, and brili rust interpreters, comparing their performance
+2. Measured the CPU wall clock time for json roundtrips (json -> flat -> json). (Although there wasn’t a specific baseline for this)
+
+Below is a table showing the time taken for json roundtrips. We tested this on all the core benchmarks, but due to space constraints, we only list a few here. These are averaged over 10 runs, with a warmup of 3.
+
+| Benchmark | Mean [ms] | Min [ms] | Max [ms] | Relative |
```
sampsyo: What is the "Relative" column here? What's being compared to what?

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+| `bitshift` | 451.5 ± 9.7 | 437.9 | 466.1 | 1.03 ± 0.03 |
+| `call` | 453.9 ± 34.5 | 432.4 | 541.5 | 1.04 ± 0.08 |
+| `const` | 521.2 ± 160.5 | 435.0 | 941.7 | 1.19 ± 0.37 |
+| `euclid` | 472.9 ± 37.2 | 436.8 | 544.3 | 1.08 ± 0.09 |
+| `main-args` | 442.6 ± 10.5 | 430.9 | 464.3 | 1.01 ± 0.03 |
+| `montgomery` | 441.6 ± 5.7 | 434.2 | 454.4 | 1.01 ± 0.02 |
+| `nop` | 521.6 ± 114.0 | 438.3 | 744.3 | 1.19 ± 0.26 |
+| `perfect` | 503.1 ± 37.1 | 462.7 | 578.5 | 1.15 ± 0.09 |
+| `reverse` | 486.1 ± 61.1 | 432.5 | 616.1 | 1.11 ± 0.14 |
+| `rot13` | 438.5 ± 7.8 | 428.4 | 452.8 | 1.00 |
```
sampsyo: Is it really the case that a JSON -> flat -> JSON round-trip takes ~500 ms (half a second)? That seems like a really long time! Because these programs are pretty small, I'd have expected it to be, like, less than a millisecond. Any guesses about what is taking so long?

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
...hows that most of the runtime of the executable spent was in the interpreter-related functions: `mmap`-ing the flat file  format was relatively quick, and so was converting the JSON to a flat file format (omitted from the screenshot). Zooming into the stack chart (second screenshot below) and focusing on the function calls towards the bottom, we see that when interpreting individual individual instructions, our interpreter calls a lot of standard library functions that manipulate `HashMap`s and `Vec`s. This is because the environment datatype in our interpreter is defined as `HashMap<&str, BrilValue>` (mapping variable names to a Bril value), i.e. keys in the hashmaps are (references to) strings. We realized that strings were the canonical way to represent variables in the environment, since different indexes in the arguments field of an instruction may point to the same underlying variable. However, the fast Rust Brili interpreter canonicalises variable names into a numerical representation (as described in their blogpost), allowing for faster look-ups in their environment. We suspect that this is one of the reasons why the Rust Brili interpreter out-performs our flat interpreter.
```
sampsyo: Thanks for doing this investigation! I agree that it is likely that this is the performance bottleneck here. It would be interesting to try to alleviate this bottleneck by requiring that variable name indices be unique…

content/blog/2025-05-13-flat-bril/index.md
```diff
@@ -0,0 +1,170 @@
...
+
+Our interpreter compares favorably to Rust Brili in terms of peak physical memory usage (max resident set size). Notably, our interpreter has an order of magnitude fewer page faults and involuntary context switches compared to the TypeScript interpreter (250 vs 2036 and 210 vs 2480), although it’s unclear whether this is due to our choice of implementation language (Rust vs TypeScript) as opposed to our data structure flattening strategy, since Rust Brili also has similar metrics. 
+
+### What else could be done?
+At the moment, our implementation creates `Vec`-based buffers to store the variables/labels it sees within a Bril program. Since Bril programs are unlikely to contain excessively large no. of variables/labels, perhaps we could use Rust’s `smallvec` or `arrayvec` libraries to create these buffers instead of the standard library’s `Vec` data structure. The aforementioned libraries offer specialized short vectors which minimize heap allocations, which might offer performance benefits over heap-allocated `Vec`s. However, we would need to properly benchmark these supposed benefits to see if swapping to these libraries is worth it.
```
sampsyo: no. -> number (you're not under space limitations here :)

sampsyo (changes requested):
Really nice work here, y'all! I found your description of the flattening process detailed and useful, and it's awesome that you got all the way to implementing and evaluating a "native" interpreter. Fantastic!

I have a few questions within. And one more: would you be interested in contributing your new representation & interpreter back to the Bril monorepo? I imagine that other people might find it useful!

content/blog/2025-05-13-flat-bril/index.md
ngernest: Thanks for the catch! We realized that the original JSON roundtrip benchmark results were faulty for two reasons:
1. The command we passed to Hyperfine for benchmarking was of the form `bril2json < SOME_BRIL_FILE | cargo run -- ...`, and Samply revealed that a lot of the ~500ms was dominated by `bril2json`'s file I/O. I updated our code so that the JSON benchmarks don't require piping the output of `bril2json` to our executable.

2. We forgot to do `cargo build --release` before benchmarking (so some optimizations from the `release` build might have been missed)

I've rerun the benchmarks using the `--release` build and updated the figure in the blogpost!

content/blog/2025-05-13-flat-bril/index.md
ngernest: (By default, the Markdown table that Hyperfine outputs includes this "Relative" column, but there is no baseline here for us to compare against, so the column is inappropriate.) I removed the column -- thanks!

content/blog/2025-05-13-flat-bril/index.md
ngernest: I rewrote this section in the `.md` file, hopefully it's clearer now! 

(In hindsight, our design is a bit confusing in that we use close intervals throughout, and we enforce the invariant that the bytes for the actual variable-name strings must be stored contiguously, and their indexes in the interstitial indexes array must also be stored contiguously.)

content/blog/2025-05-13-flat-bril/index.md
ngernest: Ah good point thanks! We use closed intervals in our implementation, but I'm guessing there's a similar trick we can do.

ngernest:
Thanks Adrian for the comments! I rewrote a few sections of the blogpost and reran some benchmarks based on the comments. I'll put up a PR for the Bril monorepo later!

content/blog/2025-05-13-flat-bril/index.md
sampsyo: Got it; makes sense.

content/blog/2025-05-13-flat-bril/index.md
sampsyo: Sounds good. FWIW, there might also be a significant cost to using `cargo run` instead of just running the executable directly…

sampsyo:
Looks great; thanks!!

PR #537, opened 2025-05-13
Final project blog post: Mark Barbone

### LINKED ISSUE #508: Project Proposal: Polyhedral loop optimizations for Bril
# Project proposal: polyhedral loop optimizations for Bril

In this proposal, I will:

1. Design and implement a polyhedral intermediate representation for Bril code
2. Implement loop optimizations on the polyhedral intermediate representation
3. Evaluate its performance on a benchmark suite of Bril code

## What?

Consider the following code, which sums up the contents of a 100x100 array:

```c
int total = 0;
for (int j = 0; j < 100; j++) {
    for (int i = 0; i < 100; i++) {
        total += arr[i,j];
    }
}
```

If the array is stored in row-major order, this is a slow way to sum it up: since the inner loop strides over 100 elements each iteration, there is no spatial locality! It would be more cache-friendly to loop over `i` in the outer loop and `j` in the inner loop.

A _polyhedral intermediate representation_ is an IR used to do this kind of loop analysis and restructuring in modern compilers: [LLVM] and [GCC] both incorporate polyhedral IRs, as does the relatively new [MLIR] framework on top of LLVM.  In a polyhedral representation, this program will be have add instruction, parameterized by the space `{(i,j) : 0 <= i < 100, 0 <= j < 100}` of points for which it's run.

[LLVM]: https://polly.llvm.org/
[GCC]: https://gcc.gnu.org/wiki/Graphite
[MLIR]: https://mlir.llvm.org/docs/Rationale/RationaleSimplifiedPolyhedralForm/

I will implement a polyhedral intermediate representation for Bril programs, using the Bril memory extension, and use it to implement basic loop optimizations. I have, as a concrete goal, the following:

1. Construct a polyhedral IR representation for a Bril program (in SSA form)
2. Do a dependence analysis on the IR
3. Analyze possible iteration orders for spatial cache locality
4. Turn the polyhedral IR back into concrete loops

This will improve performance by improving cache locality.

### Rationale and scope

There are, of course, many other things you can do with a polyhedral IR.  A [past 6120 project] was also about the polyhedral model, but it was far more ambitious, both (a) using MLIR/LLVM instead of Bril and (b) setting out to implement more non-trivial loop optimizations, and it did not get to a point where it could measure results on benchmarks. In light of this, I am being intentionally unambitious with initial goals, just looking at improving cache locality for Bril programs.

[past 6120 project]: https://www.cs.cornell.edu/courses/cs6120/2023fa/blog/polyhedral/

If all goes well, there is practically unbounded room to make the project more complicated, by implementing new loop optimizations.

## How?

There are a few sources on design considerations for the polyhedral model. I will follow the dragon book, chapter 11, which has a set of algorithms for the polyhedral loop analyses and transformations.

The code transformation parts depend on a lot of integer set transformations and algorithms; for these, I will use the [isl] library.

[isl]: https://libisl.sourceforge.io/

## Evaluation

I will evaluate correctness and performance on multiple benchmark suites:

1. The `core` bril benchmark suite. This does not have arrays in it; therefore, it serves as a baseline to show that it preserves behavior and does not slow down normal code.
2. The `mem` bril benchmark suite. This does have arrays in it, so it would show the performance change for code which _does_ have arrays.
3. Purpose written array-manipulating benchmarks, such as matrix multiplication. These would be directly designed to benefit from my optimizations, and showcase the best-case performance improvement.

In terms of benchmarking methodology, running it via the interpreter is potentially a worse measurement of cache effects than running machine code, so I will benchmark it with the `brilift` Bril-to-native code compiler.

## Team

Just me, @mb64

mb64:
Closes #508 .

### DIFF
--- a/content/blog/2025-05-13-polyhedral.md
+++ b/content/blog/2025-05-13-polyhedral.md
@@ -1,0 +1,190 @@
++++
+title = "A Simplified Polyhedral IR for Bril"
+[[extra.authors]]
+name = "Mark Barbone"
++++
+
+In [this project][github], I set out to build a polyhedral IR for Bril
+programs.  My original goal was to implement polyhedral optimizations on it;
+however, simply constructing and going out of a polyhedral IR ended up being
+complex enough that it filled my budget.
+
+[github]: https://github.com/mb64/cs6120-work/tree/main/poly
+
+## Background
+
+Here's an example program:
+
+```c
+for (i = 0; i < N; i++) {
+    for (j = i; j < N; j++) {
+        foo(i, j);
+    }
+}
+```
+
+This program has a pretty complex control flow graph -- each `for` loop is
+hiding whole host of basic blocks that it compiles to.  Often, this is great
+for a compiler, as breaking up the control flow into small but regular pieces
+means that general-purpose optimizations can handle all kinds of programs
+uniformly.
+
+However, sometimes we want to do more high-level optimizations. For example, it
+may be possible to interchange the two loops:
+
+```c
+for (j = 0; j < N; j++) {
+    for (i = 0; i <= j; i++) {
+        foo(i, j);
+    }
+}
+```
+
+Figuring out how to do this interchange correctly is extremely non-trivial,
+even on the high-level `for`-loop-containing program text! For this, it would
+be very convenient to have some intermediate language that explicitly accounts
+for iteration spaces like `{ [i, j] : 0 <= i and i <= j and j < N }`, so that
+these transformations would be possible. This is precisely the goal of the
+_polyhedral model_.  In my IR, this is represented as:
+
+```c
+for (i ∈ { [i] : 0 <= i < N }) {
+    for (j ∈ { [i, j] : 0 <= i <= j < N }) {
+        foo(i, j);
+    }
+}
+```
+
+Here, the iteration domain appears as a first-class entity in the IR's looping
+constructs. I don't exactly use the classic polyhedral framework, but rather a
+simplified form inspired by [MLIR's simplified form][mlir], and I'll explain
+the details in the next section.
+
+
+## The intermediate representation
+
+A classic polyhedral IR has a bag of ordered statements, each with an
+associated domain of points on which it's run, and an associated schedule of
+when to run it, in the form of an affine function outputting a timestamp.
+
+With this classical representation, turning it all back to code in the end is
+very hard: you need to come up with an optimal set of loops and `if` statements
+that correctly implements the ordering requested by the schedule, all without
+incurring too much blow-up or obscuring hot loops with too many conditionals.
+Here are a few papers I found on how to do this, roughly ordered by how complex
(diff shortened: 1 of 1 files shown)


content/blog/2025-05-13-polyhedral.md
```diff
@@ -0,0 +1,190 @@
...
+   |  instr_1; ...; instr_N; c
+   |  if b then c else c
+   |  jmp .l
+   |  ret v
+```
+
+A jump to label `.l` is only allowed when `.l` is in scope (this is what makes
+it structured!). Jumping to a `block` label is like a `break`, escaping to the
+end of the block, while jumping to a `loop` label is like a `continue`, going
+back to the loop header for the next iteration.
```
sampsyo, line 140: This is a clever way to repurpose Bril labels and `jmp` in a structured representation.

content/blog/2025-05-13-polyhedral.md
```diff
@@ -0,0 +1,190 @@
...
+the moment does zero backend optimizations, which would be quite useful to
+clean up the code. I expect this to go back to close to 1x with simple backend
+clean-up optimizations.
+
+**Applicability.** Not every program has static control flow, so only some
+could potentially benefit from future polyhedral optimizations. In the combined
+mem and core benchmarks, 176 out of 272 functions had fully static control
+flow, or around 65%. Considering just the relooper, interestingly, all 272
+functions in these combined test suites exhibit reducible control flow, and
+successfully round-trip through just the structured control flow IR.
```
sampsyo, line 189: I agree that this is interesting! I would have expected _some_ benchmark in there to do something weird with its control flow.

sampsyo (approved):
Really awesome work, Mark—this is fantastic. Thanks for the clear explanation of your MLIR-inspired IR and the overview of all the steps in the chain required to make the whole round trip work.

This seems like a fun foundation that someone might want to build on in the future. Would you be interested in contributing your machinery back to the Bril monorepo to perhaps help facilitate that?

mb64:
> This seems like a fun foundation that someone might want to build on in the future. Would you be interested in contributing your machinery back to the Bril monorepo to perhaps help facilitate that?

I would be interested in this! But I think it deserves some polishing before contributing it. Namely,

* Either a very clear write-up of the Python interfaces, or a JSON format for Bril with static control flow
* Better internal docs / comments about code layout, how the algorithm works, and what the limitations are
* Addressing whatever hacks are there (either redoing them better or at least saying what they are)
* I would probably want to include a short note about ISL, too. It's not exactly the most intuitive to work with, as the [prior polyhedral post pointed out](https://www.cs.cornell.edu/courses/cs6120/2023fa/blog/polyhedral/#challenges).

Let me know your thoughts -- if this seems reasonable, I'll do it sometime over the summer.

sampsyo:
Great! Sure, all of these sound quite reasonable. FWIW, I wouldn't bother trying to create a JSON representation for the static-control-flow form of Bril… that sounds great to have, but not necessary for an "MVP" that we could merge. Concocting a sensible serialization format for ISL sets seems like an "interesting" design problem that we could leave to future generations.

PR #538, opened 2025-05-13
Add blog post for my final project

### LINKED ISSUE #510: Project Proposal: Tractable Validation of Floating-Point Verification Tools
# Motivation and Problem Statement
There has been a flurry of work on verifying numerical methods and, in particular, bounding round-off error. (See FPTaylor, interval analysis approaches, tools listed on FPBench, etc.) These verification tools are often implemented as underdocumented research code.  It is unclear how to properly validate these tools in practice. There are also likely bugs. This project aspires to develop techniques to 1) test floating-point verification tools, and 2) find bugs in real world verification software.

# Approach
We can't simply compare the output of two different FP verifiers. If the bounds differ, it may be due to differences in precision, rather than soundness issues.

One might want to test FP verification tools by executing programs over their concrete domain (of floats). For example, for a given program $P$ and input vector $\vec{x}$, a verification tool might say that:

```latex
\forall \vec{x}, \llbracket P(\vec{x}) \rrbracket_{real} - \llbracket P(\vec{x} \rrbracket_{float} \leq B
```

for some bound $B$. A naive approach would be to search for a concrete vector $\vec{x}$ that violates the above bound.  This might find bugs.

However, observe that many (but not all) verification tools operate over an abstract domain using overapproximate simplifying assumptions.  For example the "standard model of floating-point error" assumes that each floating point operation `(op x y)` can introduce up to $1*e$ error. To be precise, for every operation, the following overapproximation holds for some unit roundoff value $u$:

        (op_float x y) = (op_real x y) * (1 + e)

for $|e| <= u$.

The essential insight I wish to exploit is that we can find bugs with respect to the _overapproximation_ rather than with respect to the actual IEEE spec.  To construct a bug in such a verification tool, it suffices to furnish an counterexample program, input vector $\vec{x}$ and trace of epsilons added at every single floating point operation that violates the error guarantees provided by the verification tool.

Note that this notion of correctness is with respect to the assumptions and overapproximations used by the floating point verification tool itself, NOT with respect to the IEEE floating point spec. In particular, it is possible for the (faulty) error bound to be accidentally correct with respect to the IEEE floating point spec (as in, no concrete set of inputs to the program will violate the error guarantee under the floating point semantics) but not sound (the error guarantee was constructed in a way that violates the soundness proof and overapproximations assumed by the verification tool).

## MVP
For a program $P$, an MVP would be to randomly sample for valid $\vec{x}$. Then, for each valid set of inputs, interpret $P$ with input $\vec{x}$ under an infinite-precision semantics (to mimic the reals) and under an infinite-precision semantics that adds up to $u error at every single operation. Check if the error observed violates the error guarantees given by the verification tool.

Note: Some verification tools take advantage of special hardcoded facts about floats. For example we know that multiplying by 2 or Sterbenz subtraction do not introduce error. I would need to be careful that we have added all the special hardcoded facts that the verification tool has access to.

## Going Further (with a backwards static analysis)
A simple way to improve on the MVP would be to do a backwards static analysis on the program $P$ to find the appropriate directions to point the epsilons at every single floating point operation.

As an example, consider the following program, whose error terms we wish to maximize:

        (a - b) / (c + d)

To maximize the error of the above program, we wish to maximize the error terms of $(a - b)$ and minimize the error terms of $(c + d)$.  To maximize the error terms of $(a - b)$, we need to maximize the error terms in $a$ and minimize the error terms in $b$.  To maximize the error terms of $(c + d)$, we need to maximize the error terms in $c$ and $d$.  We can continue this analysis until we know exactly how to point our error terms at every single floating point operation such that the error terms in the final output of the program is maximized.

## Going Even Further (with autodiff and smarter fuzzing)
Note that we are still randomly (perhaps uniformly) sampling for good $\vec{x}$. This is a search problem.  Also observe that most benchmark floating point programs $P$ are differentiable. To find good $\vec{x}$, we can hook it up to a fuzzer (that tries to increase the error) and run an automatic differentiation tool on $P$ to greedily / locally optimize $\vec{x}$. Explore and exploit.

## Going Much Further (somewhat speculative)
Wouldn't it be cool if we could prove that the backwards static analysis suggested was indeed worst-case for the program (up to our overapproximations)? Then we might be able to combine this information with other verification techniques to provide new kinds of floating point error bounds.

# Proposed Evaluation
I plan to run this tehnique on all FPBench benchmarks and as many verification tools that take in FPCore programs as input.  Success will be measured in number of bugs found.

# Team members
Just me; open to others joining.

InnovativeInventor:
Closes #510.


content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
++++
+title = "What is a bug? New ways to test verification tools."
+[extra]
+latex = true
+bio = "Max Fan is a first-year CS PhD student at Cornell University."
+[[extra.authors]]
+name = "Max Fan"
+link = "https://max.fan"
```
sampsyo, line 8: Incredibly cool URL.

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+an input _x_ such that **GUARANTEE(**_p(x)_**)** does not hold.
+
+This is a perfectly reasonable definition. If you can violate a guarantee, it
+seems like you have a bug.
+
+But deep down in our hearts, there is _secretly_ a stronger notion of
+correctness that we'd like the verification tool to respect. In particular, it
+would be really bad if _I_ published a paper claiming my tool does **X** and
+prove **X** to be a sound technique when in reality it does **Y**. That sounds
+like a bug!
```
sampsyo: I've kinda lost track of what X and Y are here, and how they relate to VERIFY and GUARANTEE above. Is X like GUARANTEE', a different property of program executions that implies the original GUARANTEE? A tiny bit more intuition might help here.

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+> iff it does not faithfully implement its own theory. 
+
+## Abstract Testing of Floating-Point Software
+
+With a fancy new definition, let's hunt for bugs! Observe that many (but not
+all) verification tools operate over an abstract domain using overapproximate
+simplifying assumptions. 
+
+For example, the "standard model of floating-point error" assumes that each
+floating-point operation $(op_{float} \ x \ y)$ can introduce up to $1*\epsilon$
```
sampsyo: Do you mean 1+\epsilon? (That's what appears in the equation below.)

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+error. To be precise, for every operation, the following overapproximation holds
+for some unit round-off value $u$:
+
+> $(op_{float} \ x \ y) = (op_{real} \ x \ y) * (1 + \epsilon)$
+
+for $|\epsilon| <= u$. Different tools may use varying overapproximations, but the
+principle is the same: to tractably verify you (typically) need to
+overapproximate.
```
sampsyo: Can you briefly mention the relationship between this standard model and actual floating-point reality? Presumably, the relationship is: actual floating point evaluation has error that is bounded by \epsilon, but the actual amount of error might differ between different operations and different inputs.

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+
+But how to test? Here's an idea: one can view an overapproximation as a bigger
+testing budget to smash the verification tool with. For floating-point
+verification, we can simply add $\epsilon$ error at every operation without
+worrying about whether the IEEE floating-point spec actually allows us to. In
+other words, the testing game is to find input $x$ and $\epsilon$-trace
+$\epsilon_1$, $\epsilon_2$, $\epsilon_3$ ... added at run-time such that we can violate
```
sampsyo: What is an "$\epsilon$-trace"? I think I get it, but it would be nice to allocate one additional sentence to define what this new term means. Especially because the rest of this paragraph seems to imply that you'll always add the same \epsilon to every operation…

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+   thing, we realize we need to
+   - minimize the error term in $c$, and,
+   - minimize the error term in $d$.
+
+Once we know which direction to point each $\epsilon$, we can assign the most
+extreme values (either $+u$ or $-u$) to get a pretty good $\epsilon$-trace.
+
+For this program, the $\epsilon$-trace we computed happens to be the
+**worst-case**, which is awesome. This technique generalizes in a pretty
+straightforward manner to a backwards static analysis. [^2]
```
sampsyo, line 181: Does that static analysis _always_ produce the worst-case epsilon-trace? I'm assuming not… if not, can you explain a little bit more about when it is conservative (in the sense that there may exist an even worse/larger epsilon-trace)? Seeing one example of such a program would be instructive!

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Implementation and Preliminary Evaluation
+To implement this technique, I interpret a program twice:
+
+- under an ideal (infinite-precision) semantics with no floating point error,
+and,
+- under an approximate semantics that adds a trace of $\epsilon$s at run-time.
+
+To evaluate, I selected a few benchmarks from the
+[FPBench](https://fpbench.org/benchmarks.html) I had already written a parser
```
sampsyo:
```suggestion
[FPBench](https://fpbench.org/benchmarks.html). I had already written a parser
```

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+and could reuse infrastructure for another project. My weapon of choice was
+OCaml for similar practical considerations. [^3]
+
+
+### Charts and Graphs
+Below is a log-scale violin plot showing the distribution of absolute error
+abstractly witnessed by my prototype tool. [^4] (_Aside: I think more people should use
+violin plots._)
```
sampsyo: Can you say a little bit more about where this distribution came from? It sounded like, from the above description, your analysis would produce a _single_ \epsilon-trace, and therefore there would be a single error value for every benchmark you try. What are you randomly sampling? The input values? If so, from what ranges, and with what distributions?

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+| delta | 5.457664e-13 | 1.657179e-12 |2.197089e-12| 2.346965e-12 |
+| delta4 | 3.126935e-14 |  7.394990e-14 |7.676607e-14| 1.160113e-13|
+
+As you can see, the maximum error sampled can get quite close to the guarantee
+provided by the various tools.
+
+For future work, it would be cool to run this on more tools and benchmarks.
+Additionally, a question not addressed by this tool (and omitted blog post) is
+how to find a good input $x$. Currently, the testing tool uniformly samples from
+the input space. One technique I wrote in my initial proposal (but did not have
```
sampsyo: Ah, maybe this answers my question above. But it seems worth hoisting earlier…

content/blog/2025-05-13-abstract-testing-floating-point.md
```diff
@@ -0,0 +1,201 @@
...
+![Violin plot of "Absolute error (abstractly) witnessed by FPBench
+benchmark".](/blog/sample-violins.svg)
+
+Finally, here is a table showing the minimum and maximum error sampled on a few
+selected benchmarks along with the corresponding guarantees provided by FPTaylor
+and Daisy:
+
+| Name | Min. error sampled | Max. error sampled | FPTaylor Guarantee | Daisy Guarantee |
+|:----:|:-----------------:|:-----------------:|:-:|:-:|
+| rigidBody1 | 0.0 | 3.409494-13 | 2.948752e-13 | 2.948752e-13 |
```
sampsyo: For this case (and a few others), should I be alarmed that the max error is greater than the guarantees from the tools? Does this mean you found a bug??

sampsyo (changes requested):
Nice work here, @InnovativeInventor! This was a clear explanation of the approach and the goal, and the outcome seems more or less expected. You'll see one big question in my inline comments: did you find any bugs? :smiley: Either way, it might be nice to state that outcome in the post somewhere.

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: Yep, fixed -- good catch. Thanks!

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: Thanks for the feedback. I added the following text to clarify the relationship between these two:
```diff
+To be precise, for every operation, the following overapproximation (of
+the IEEE floating-point spec) holds for some unit round-off value $u$:

 > $(op_{float} \ x \ y) = (op_{real} \ x \ y) * (1 + \epsilon)$

 for $|\epsilon| <= u$. Different tools may use varying overapproximations, but the
 the principle is the same: to tractably verify you (typically) need to
+overapproximate. The IEEE floating-point spec, by contrast, is a fully
+executable and deterministic spec.
```
Does this help?

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: Ah `**X**` and `**Y**` are meant to be different metavars tracking different things. I had thought that different casing and formatting would make this clear, but clearly it didn't work. Sorry about the confusion. I updated the text to say:
```diff
-would be really bad if _I_ published a paper claiming my tool does **X** and
-prove **X** to be a sound technique when in reality it does **Y**. That sounds
-like a bug!
+would be really bad if _I_ published a paper claiming my tool does **\$BLAH**
+and prove **\$BLAH** to be a sound technique when in reality it does **\$BLEH**.
+That sounds like a bug!
```

sampsyo: Sure… I guess I'm asking mostly about the overall point you're getting at here. How do X and Y relate to VERIFY and GUARANTEE? One of the pairs sounds like "the stated goal of the tool" and one is maybe "the way the tool is formalized" or something, but what is the relationship between these two? Surely they have _some_ kind of relationship!!

content/blog/2025-05-13-abstract-testing-floating-point.md
sampsyo: It helps some, but would it be correct to say that actual IEEE FP semantics have error that is bounded by the "standard error"? Kind of like my comment above, I want to know the relationship between these two things.

sampsyo:
Hi, @InnovativeInventor! Looks like you've added a few commits here—let me know when you think this is ready and I'll take another look (and publish).

sampsyo:
Just one last reminder about the above, @InnovativeInventor—let me know when it's time to take another look.

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: Added the following to clarify things:
```diff
 ### Charts and Graphs
 Below is a log-scale violin plot showing the distribution of absolute error
-abstractly witnessed by my prototype tool. [^4] (_Aside: I think more people should use
-violin plots._)
+abstractly witnessed by my prototype tool, broken down by benchmark program. The
+distribution comes from uniformly sampling from the program input space and
+using the corresponding $\epsilon$-trace for each program constructed via the
+backwards static analysis described above. [^4] (_Aside: I think more people
+should use violin plots._)
```

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: Done!

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: Yes! I am reluctant to publicly claim that they're a bug until I have identified the precise cause and have a fix accepted by the upstream developers. I added the following line to make this clear but hedge a bit. I don't want to overclaim here.

```diff
+Entries that are **bolded** indicate a potential
+bug, in either the verification tool under test (~thousands LOC) or the testing
+framework itself (a few hundred LOC).
```

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: Thanks for the suggestion! I added the following footnote to avoid breaking up the flow of the text too much:
```diff
+[^2]: A wrinkle occurs when terms can be shared. If a verification tool cannot
+    reason about shared error terms, this algorithm still produces a worst-case
+$\epsilon$-trace. Otherwise, this one might need to do a little searching to
+find the worst-case $\epsilon$-trace. In particular, if error terms can be
+shared, we start to care about the constants in the floating-point expression.
+Concretely, `(x - 2x)` is quite different from `(2x - x)`; in the former we wish
+to minimize the error term in `x`, while in the latter we wish to maximize the
+error term in `x`.
```

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: I added the following sentence to help clarify things:
```diff
+For floating-point verification, we can sample concrete inputs _and_
+opportunistically construct bad $\epsilon$s at every floating-point operation to
+violate the verification tool. 
```

content/blog/2025-05-13-abstract-testing-floating-point.md
InnovativeInventor: Does this help? I also overhauled how I was explaining the tool -- added a new section explaining the tool in terms of abstract interpretation.

```diff
To be precise, for every operation, the following overapproximation (of
+the IEEE floating-point spec) holds for some unit round-off value $u$
+(determined by the float precision and rounding mode used):
```

InnovativeInventor:
I've been thinking about better ways to explain this idea (the old version of the blog post was quite dissatisfying and seemed jumbled). 

To make it easier to understand, I added a new section titled:
```diff
+ ## Interlude: Abstract Interpretation
... [parts of diff removed] ...
+ The key idea we will exploit in this blog post is that there are many different
+ semantics (ingredient no. 4) for which the *same* concrete domain (1), abstract
+ domain (2), programming language syntax (3), and abstract transformer (5) are
+ still sound for!
```

Hopefully it helps!

---

For some context: this idea came out of a side project of mine (with Ben Kushigian and a few others at UW) where we're interested in test completeness: doing (provably sound) program verification via testing. A test suite is complete iff passing the test suite implies the spec is satisfied; equivalently, a test suite is complete if adding another test won't catch more defective implementations. Investigating complete test suites is interesting for a few reasons, but one thing that it addresses is "what is a good spec?" -- if you like every example in a compete test suite, you are _committed_ to liking the spec (or a refinement of it).

More relevantly to this blog post, there is an notion we're developing that is dual to test completeness -- a complete implementation. An implementation is complete if it can (possibly non-deterministically) witness all the valid traces that are allowed under the spec. My initial idea was to do a "complete implementation" of the standard (1 + $\epsilon$) model of floating point error. But this is not necessary: one can quite easily cut down on the search space, which is the approach I ended up going with.

sampsyo:
Nice! This seems great. It seems like this version gets across that underlying motivation a bit more directly. Nice work!

PR #541, opened 2025-05-14
Add Brilooped

### LINKED ISSUE #504: Project Proposal: Brilooped
**What will you do?**
* The idea is to implement the relooper algorithm to convert bril programs into bril programs with structured control flow. 
* The idea is that we'll be able to write bril programs with two extra constructs: if-then-else and while loops. 
* To generate brilooped programs, you'll be able to write them by hand and compile them to bril. 
* Or you'll be able to generate briloop programs from bril files, a translation implemented using Ramsey's relooper algorithm.
* Another team (Michael, Annabel, and Serena) will work on a project that transforms briloop programs into webassembly. We'll coordinate with them on the file format and our end goal is that we can build a fulll pipeline bril -> briloop -> wasm. 

**How will you do it?**
* To implement the compilation from briloop to bril we'll use a classic compiler pipeline that will translate while and if statements into blocks and jump instructions. The main idea of this phase is to grow comfortable with the grammar for brilooped programs and to enable us to test autogenerated brilooped programs. We hope to spend only a small amount of time in this phase. 
* To implement compilation from bril to briloop we'll use the relooper algorithm from Ramsey. We hope to spend a majority of time of the project in this stage. 

* By implementing a full translation loop! bril -> briloop -> bril and test the output back on bril benchmarks. 
* Our goal is for the implementation to be correct (i.e. the output of each program is equivalent), we can also establish a metric of efficiency in terms of how many dynamic instructions are executed for the original bril vs the translated bril program. 
* As a stretch goal, it would be fun to compare how different some programs look if we write briloop by hand vs by compiling bril to brilooper! Would any optimizations over a bril program lead to significant different briloop program?

**Team members:**
@devv64 
@gerardogtn

devv64:
Closes #504 
@gerardogtn


content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+name = "Gerardo Garcia Teruel Noriega"
+[[extra.authors]]
+name = "Dev Patel"
++++
+
+# Brilooped
+
+## Abstract
+
+In this project we introduce Briloop, an extension into the [Bril programming language](https://capra.cs.cornell.edu/bril/) that supports structured control flow, this allows bril programs to use if-then-else, while, and block statements instead of relying on jumps and branches. We also implemented Ramsey's [Beyond Relooper](https://dl.acm.org/doi/10.1145/3547621) to translate bril programs into briloop automatically, having a an average performance penalty of 9% when using the core bril benchmarks. Brilooped not only enables programmers to write bril programs with structured control flow, but also enables a translation from Bril into WebAssembly.
```
sampsyo: Let's try to keep capitalization consistent throughout the post: bril -> Bril and briloop -> Briloop.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+
+## Representing Briloop Programs
+
+Brilooped programs extends the Bril instruction set with the following op codes: while, block, if, break, and continue.
```
sampsyo: Seems like a good idea to mark these opcode names with Markdown code backticks: `while`, `block`, `if`, etc.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+Brilooped extends the Bril instruction set with new control flow operations that provide structured alternatives to jumps and labels. These new op codes allow programmers to write more readable and maintainable code with familiar control flow constructs like loops and conditionals. This structured control flow can also be useful for specific applications such as WASM.
+
+### `while`
+
+The `while` operation introduces a structured loop into Bril programs:
+
+- It takes exactly one argument: a string containing the name of a boolean variable that serves as the loop condition.
+- It includes an additional `children` field that contains an array of arrays of `Instr` objects representing the loop body.
+  - Note: this is an array of arrays to make the language a little more uniform (children field is used in both while loops and if-then-else statements)
```
sampsyo: …and presumably `block` as well?

Maybe it would be a good idea to, before getting into the individual opcodes, introduce this new `children` field _in general_ and say that it will be used in several operations. It's interesting to highlight this up front because it means that Brilloop programs, unlike classic Bril programs, look "nested."

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+      },
+      {
+        "args": ["b", "ten"],
+        "dest": "cond",
+        "op": "le",
+        "type": "bool"
+      }
+    ]
+  ],
+  "op": "while"
```
sampsyo: Consider putting `op` first in the dictionary, for legibility.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+},
```
sampsyo: JSON listings might be a little easier to read if you "wrap" some lines manually. For example, it can be nice to put entire instructions on one line, like this:

```json
{ "args": ["b", "ten"], "dest": "cond", "op": "le", "type": "bool" }
```

Same with, like, `"args": ["cond"]`. It requires a little manual reformatting vs. what you get out of a raw `json.dump`, but it makes a big difference for the reader.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+
+### `if`
+
+The `if` operation implements conditional execution:
+
+- It takes exactly one argument: a string containing the name of a boolean variable that serves as the condition.
+- It includes a `children` field with an array of arrays of `Instr` objects.
+  - This will contain 1 or 2 arrays, belonging to the True and False blocks respectively.
+  - When the condition is true, the first array of `Instrs` objects in the `children` is executed.
+  - When it is false, the second array is executed.
```
sampsyo: (if any)

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+}
+
+```
+
+### The `break` and `continue` Operations
+
+Two additional operations provide finer control within loops:
+
+- `break`: Takes one constant argument and exits `value` loops. This brings execution to the top of this loop.
+- `continue`: Takes one constant argument and exits `value` loops. This continues execution at the following instruction.
```
sampsyo: It's maybe worth adding one more sentence to explain what the integer argument means (the "number of loops" to traverse). It goes by somewhat quickly here, and it would be nice to highlight this somewhat non-obvious language choice.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+
+### New Language Constructs
+
+We expanded Bril's control flow capabilities by introducing five new constructs to the interpreter type system:
+
+- `while` - For creating loops
+- `block` - For grouping instructions
+- `if` - For conditional execution
+- `break` - For exiting loops
+- `continue` - For skipping to the next loop iteration
```
sampsyo: I think you've already covered this above? As in, you've already explained the set of new operations you've added and what they mean, so you don't need this section here. You can just say, in one sentence, that you implemented each of the new constructs in the reference interpreter.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+            result.action === "speculate" ||
+            result.action === "commit" ||
+            result.action === "abort"
+        ) {
+            return result;
+        }
+    }
+}
+
+```
```
sampsyo, line 250: Can I encourage you to avoid the temptation to include big blocks of the implementation code here in the blog post? You can always include a link to the GitHub blob page for people who are really interested! But from the reader's perspective, just one or two sentences that briefly captures the _idea_ of the code can be way more valuable than the full code itself.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+}
+```
+
+The multi-level breaks and continues proved particularly challenging to implement, especially ensuring the correct bubble-up mechanism through nested control structures. Our solution hinged on the `level` counter, which elegantly tracks how many loop boundaries a `break` or `continue` statement needs to cross. During development, we encountered several cases where improper handling of the recursive evaluation caused programs to become trapped in infinite loops. These issues were difficult to debug — we ultimately resorted to execution profiling to trace the interpreter's control flow and identify where it was getting stuck. The resulting implementation not only handles complex control flow patterns correctly but does so with minimal additional overhead on the interpreter.
+
+## Beyond Relooper
+
+The Beyond Relooper algorithm was described by Ramsey as a way to translate Cmm sources into WebAssembly, for this project we adapted the algorith to translate Bril into Briloop. We'll start by providing a high level overview of the algorithm and some of the adaptations we did as the source and target languages are different from what are presented in Ramsey's paper.
```
sampsyo: It might be nice to include a hyperlink or something to explain what Cmm is (I think it's an IR used in GHC?).

content/blog/2025-05-13-brilooped/index.md
sampsyo: > … that supports structured control flow, this allows bril programs …

This is a comma splice. Please replace the `,` with a `;` or a `.` (and a new sentence).

content/blog/2025-05-13-brilooped/index.md
sampsyo: "algorith"

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+
+The Beyond Relooper algorithm takes as input the Dominator Tree of a Control Flow Graph, with the important characteristic that the children in the tree are sorted in descending reverse post order, and produces the Briloop Program as an output. It is implemented as a recursive function over the structure of the dominator tree, where the current node is translated into either a while statement, a list of briloop instructions, or a block statement:
+
+- A basic block is translated into a while statement if it is a loop header, we use Ramsey's definition of a block header being a loop header if it has a predecessor with a larger reverse post order than itself rather than the more standard definition of a loop header being a node that has a predecessor that it dominates.
```
sampsyo: > … if it is a loop header, we use Ramsey's definition …

This is a comma splice; please start a new sentence here.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+- If a basic block is a merge node it is translated into a briloop block statement, to support breaking.
+- Otherwise, the basic block is translated into a flat list of briloop instructions.
+
+Detecting the different kinds of translations allows us to only introduce the minimal number of blocks needed for the translation.
+
+### Merge nodes
+
+Identifying merge nodes is a key part of the briloop algorithm, they are defined as nodes that have at least 2 in edges from nodes with a smaller reverse post order. This means that in the original control flow, there are two basic blocks that jump into it and that thse nodes do not form a loop. Detecting them is important since we need to rely on block statements and break instructions in order to reach them (see Theorem three in Peterson, Kasami, and Tokura for more info).
```
sampsyo: > … a key part of the briloop algorithm, they are defined …

Another comma splice.

sampsyo: It might be nice to hyperlink from those author names to the paper you're referring to.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+        block {
+            print b;
+            continue 1;
+        }
+        print c;
+    }
+}
+```
+
+The quirk from the translation to bril to briloop is that we rely on all while loops to execute forever and rely on break statements to terminate the loop execution. This has the unfortunate consequence of worsening performance for loops (especially nested ones) as we introduce a true variable before every while translation.
```
sampsyo: [non-actionable comment; nothing to change here] As we discussed briefly today, I think it would make sense to (in a version 1.1 of this extension) replace `while` with `loop` for this reason! It keeps the language a little simpler, and it seems unlikely to make the downstream WebAssembly backend any more complicated.

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+
+## Results
+
+### Impact of Brilooped Transformation on Dynamic Instruction Count
+
+Our analysis of the Bril to Brilooped transformation using the relooper algorithm shows significant variations in dynamic instruction count across different benchmarks. As shown in Figure 1, the majority of benchmarks (37 out of 38) experienced an increase in dynamic instruction count after transformation, with only the `collatz` benchmark showing a reduction (-6.51%) due to `br` and `jmp` instructions making up a majority of the functionality. The changes in dynamic instruction count ranged from minimal increases (below 1%) for benchmarks like `sum-check` and `lcm` to substantial increases exceeding 30% for benchmarks such as `binary-fmt` (28.00%), `hanoi` (31.31%), and `pythagorean_triple` (37.60%).
+
+![Figure 1: Relative Change in Dynamic Instructions After Brilooped Transformation](relative_change.png)
+
+The blue dashed line in Figure 1 is the average (9.06%) change in dynamic instructions after the transformation. 11 of the benchmarks were significantly above (>15%) this average while 9 were significantly below (<1%), leading to a large standard deviation (9.75%)
```
sampsyo:
```suggestion
The blue dashed line in Figure 1 is the average (9.06%) change in dynamic instructions after the transformation. 11 of the benchmarks were significantly above (>15%) this average while 9 were significantly below (<1%), leading to a large standard deviation (9.75%).
```

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+![Figure 3: Impact of Brilooped Transformation vs Benchmark Size](size_vs_change.png)
+
+1. The most substantial percentage increases occurred in benchmarks of varying sizes, with `pythagorean_triple` (largest at ~10,000 instructions) showing a ~37% increase, `hanoi` (~100 instructions) showing a ~31% increase, and `binary-fmt` (~100 instructions) showing a ~28% increase.
+
+   - Briloop syntax seems to be causing lots of overhead in the smaller examples
+   - Nested loops are very expensive
+   - Recursion can be very expensive
+
+2. There doesn't seem to be any strong correlation between benchmark size and the magnitude of impact from the Brilooped transformation. Both small and large benchmarks showed varying degrees of change. Benchmarks clustered around the 10^2 instruction count range had the greatest variance, ranging from approximately -6% to +31%.
+   - Note: this could be a product of the benchmark suite.
```
sampsyo: I think you can leave this note off—it's true of basically everything all the time. :)

content/blog/2025-05-13-brilooped/index.md
```diff
@@ -0,0 +1,443 @@
...
+
+These results suggest that while transforming Bril code to Brilooped using the relooper algorithm introduces some overhead in most cases, the structured control flow benefits of Brilooped come with a reasonable performance cost. The high variance in impact across different benchmarks indicates that the transformation's efficiency depends significantly on the specific control flow patterns in the original code. This tradeoff seems acceptable, especially considering that all benchmarks maintained their functional correctness.
+
+### References
+
+Peterson, Kasami, Tokura. "On the capabilities of while, repeat, and exit statements". https://dl.acm.org/doi/10.1145/355609.362337
+
+Ramsey, Norman. "Beyond Relooper: recursive translation of unstructured control flow to structured control flow (functional perl)". https://dl.acm.org/doi/10.1145/3547621
+
+Leroy, Xavier. "The birth of control structures: from 'goto' to structured programming". https://xavierleroy.org/CdF/2023-2024/1.pdf
```
sampsyo: Because this is a webpage and not a PDF, I recommend just linking to these papers from the text instead of having a separate "references" section at the end!

sampsyo (changes requested):
Very very cool; nice work here! This is a substantial chunk of implementation work, and you've done a great job of explaining what you've done. And the empirical evaluation is great as well!!

I would love to merge two things into the main Bril monorepo:

* The specification of the Briloop language extension.
* The "Beyond Relooper" implementation that does the transformation.

Could I have your help opening some PRs for those into the Bril repo so we can get those merged? 😇

content/blog/2025-05-13-brilooped/index.md
gerardogtn: great catch! expanded this section quite a bit and added some examples for break and continue. I agree that it's non-obvious and i tried to include the examples that i used when talking to the wasm team to clarify its behaviour, open to more changes if we can still improve this section

content/blog/2025-05-13-brilooped/index.md
gerardogtn: great catch! guilty as charged of overusing comma splices. All three of them should be fixed now

content/blog/2025-05-13-brilooped/index.md
gerardogtn: I can change whiles to loop operations for the merge into the bril repo. It would definitely make things nicer and it's one of the few things that i wished we had used since the beginning instead of a while loop

gerardogtn:
> Very very cool; nice work here! This is a substantial chunk of implementation work, and you've done a great job of explaining what you've done. And the empirical evaluation is great as well!!
> 
> I would love to merge two things into the main Bril monorepo:
> 
>     * The specification of the Briloop language extension.
> 
>     * The "Beyond Relooper" implementation that does the transformation.
> 
> 
> Could I have your help opening some PRs for those into the Bril repo so we can get those merged? 😇

For the briloop language extension, we tried to keep things separate in the development repo by making a copy of `briltxt.py` and naming it `brilooptxt.py` for example. Would you prefer we merge this into the briltxt.py and keep it as the default behaviour of `bril2json`/`bril2txt`, or would you prefer to keep them as separate commands?

I'll work on sending a PR with the beyond relooper algorithm!

content/blog/2025-05-13-brilooped/index.md
sampsyo: Looks great!

content/blog/2025-05-13-brilooped/index.md
sampsyo: FWIW, I think it would be just fine to merge stuff you already have, and we can try to address the `while` -> `loop` simplification in a future PR! Whatever strategy makes sense to you.

sampsyo:
> For the briloop language extension, we tried to keep things separate in the development repo by making a copy of `briltxt.py` and naming it `brilooptxt.py` for example. Would you prefer we merge this into the briltxt.py and keep it as the default behaviour of `bril2json`/`bril2txt`, or would you prefer to keep them as separate commands?

I think it makes sense to just have one such command! That's the basic idea with all the Bril language extensions: the support exists in such common Bril tools as the parser and printer, but you don't have to use them. In other words, their presence in the parser/printer doesn't interfere with code that doesn't use the constructs.

sampsyo:
Anyway, this all looks great! I'll publish this post now.

PR #542, opened 2025-05-14
Final Project

### LINKED ISSUE #514: Project Proposal: Optimizing code with trigonometric functions
**What will you do?**

My project will be based on [Dr.Jit](https://github.com/mitsuba-renderer/drjit) codebase. Here is the [paper](https://dl.acm.org/doi/10.1145/3528223.3530099) that describes the compiler. In short, Dr.Jit traces the program to compute an AST, then performs some optimizations on this representation, then manually assembles either LLVM IR or PTX code depending on the used backend, and then it is compiled to a kernel. Here is the summary figure from the paper:
Image
Dr.Jit also supports automatic differentiation.

My project will have two unrelated parts. The first one is inspired by this [blog post](https://iquilezles.org/articles/noacos/) that explains that in many cases there is no need for trigonometric or inverse trigonometric functions in graphics code and that they can be replaced with dot and cross products improving precision and performance. My goal for this part of the project will be to optimize the use of trigonometric functions. The first way to improve it is to find patterns like this:

Image

Image

In the first case, the whole computation can be simply replaced with constant 1 since $cos^2 + sin^2 = 1$, in the second case, there is no need to compute $sin^2$ as it can be replaced with $1-cos^2$ if cos is already computed.

The second way is to transform code like that 
`ang = dr.acos(dr.dot(z, d));
co = dr.cos(ang)` into this `co = dr.dot(z, d)`.

The second part of the project is based on the observation that the current version of Dr.Jit builds one AST for the whole computation. That means that the sin function is not represented by a separate tree, but is inlined in the program every time it is called. It may not be a problem for some backends (for cuda, for example) where there is hardware support for the sin function and the call can be represented with one node in the graph (first image). It may be a problem for CPUs because in this case the call to dr.sin is expanded to a large graph (second image).

Image

Image

When the sin function is called multiple times, the AST keeps expanding. I'd like to explore the possibility of placing these functions into separate ASTs. 

**How will you do it?**

I will implement the first part using pattern matching. I expect this implementation to be somewhat similar to the peephole optimizations.

To implement the second part, I will need to modify how ASTs and LLVM code are generated. Additionally, I'll need to figure out how to propagate this change to the autodiff.

I will use the benchmarks mentioned in the [paper](https://dl.acm.org/doi/10.1145/3528223.3530099) (Fig. 6) to evaluate the change in performance. I will also compare the size of the generated code.

mariasoroka:
Closes #514


content/blog/2025-05-13-Final Project.txt
```diff
@@ -0,0 +1,44 @@
...
+[extra]
+bio = """
+  Grace Hopper made the first compiler. [Adrian Sampson](https://www.cs.cornell.edu/~asampson/) is an associate professor of computer science, so that's pretty cool too I guess.
+"""
+[[extra.authors]]
+name = "Adrian Sampson"
+link = "https://www.cs.cornell.edu/~asampson/"  # Links are optional.
+[[extra.authors]]
+name = "Grace Hopper"
++++
```
sampsyo: Please include your own title, author name, and bio.

content/blog/2025-05-13-Final Project.txt
```diff
@@ -0,0 +1,44 @@
...
+
+My project was based on Dr.Jit codebase. Here is the [paper](https://dl.acm.org/doi/10.1145/3528223.3530099) that describes the compiler. In short, Dr.Jit traces the program to compute an AST, performs some optimizations on this representation, then manually assembles either LLVM IR or PTX code depending on the used backend, and finally compiles it into a kernel.
```
sampsyo: What does "manually" mean in this context?

content/blog/2025-05-13-Final Project.txt
```diff
@@ -0,0 +1,44 @@
...
+tmp_3 = tmp_1 - tmp_2
+```
+
+Below is the trace of the following code snippet before and after the modification. Clearly, the optimization worked out and $\sin^2$ is now computed as $1 - \cos^2$. Note that $\sin$ node still appears in the trace because $\sin$ was also created as a separate variable. 
+
+
+<img src="./2025-05-13-Final_Project/graph_old.png" alt="drawing" height="300"/>
+<img src="./2025-05-13-Final_Project/graph_new.png" alt="drawing" height="300"/>
+
+To better test the optimization and evaluate the performance, I planned to render the three scenes shown in Fig. 6 of the Dr.Jit paper. However, I noticed that during rendering, the optimization was never invoked. To address this, I modified the renderer code to make it less efficient, ensuring that there will be nodes to which the optimization can be applied. I rendered all the scenes with and without my optimization five times to get average times. The results are reported below.
```
sampsyo: Can you say a tiny bit more about that change you applied? Did it involve manually going the "opposite direction" from your transformation (i.e., replacing a `cos` expression with a `sin` expression?

content/blog/2025-05-13-Final Project.txt
```diff
@@ -0,0 +1,44 @@
...
+
+Below is the trace of the following code snippet before and after the modification. Clearly, the optimization worked out and $\sin^2$ is now computed as $1 - \cos^2$. Note that $\sin$ node still appears in the trace because $\sin$ was also created as a separate variable. 
+
+
+<img src="./2025-05-13-Final_Project/graph_old.png" alt="drawing" height="300"/>
+<img src="./2025-05-13-Final_Project/graph_new.png" alt="drawing" height="300"/>
+
+To better test the optimization and evaluate the performance, I planned to render the three scenes shown in Fig. 6 of the Dr.Jit paper. However, I noticed that during rendering, the optimization was never invoked. To address this, I modified the renderer code to make it less efficient, ensuring that there will be nodes to which the optimization can be applied. I rendered all the scenes with and without my optimization five times to get average times. The results are reported below. 
+
+<img src="./2025-05-13-Final_Project/evaluation.png" alt="drawing" width="300"/>
```
sampsyo, line 53: Before showing your results, can you briefly say something about your experimental setup (hardware/OS, Dr.Jit version, etc.) and your approach to measurement (how did you measure the execution time, how many replicas did you use, etc.)?

content/blog/2025-05-13-Final Project.txt
```diff
@@ -0,0 +1,44 @@
...
+
+Well, my optimization did not improve the performance, but at least I know that the trace modification was correct since produced images were identical.
+
+The second part of the project was much less straightforward, and I was not able to figure it out. As described in the project proposal, the idea was to trace functions that lack hardware support and cannot be represented by a single node into a separate trace, and then redirect the main AST to that newly created trace. I was unable to find a way to achieve this using the existing tools in the codebase. Implementing this optimization would require introducing a new type of node (e.g., `call`) and writing the corresponding PTX or LLVM IR code to support it.
```
sampsyo: > As described in the project proposal

Maybe it would be a good idea to link to the proposal, so people can read it if they want.

sampsyo (changes requested):
It's too bad that the original plan did not work out! This was, however, a clear explanation of the small change that you did manage to apply. There are a few places where some additional detail would be very useful.

sampsyo:
Looks good; I'll publish this now!

PR #546, opened 2025-05-15
Bril to x86 Blog

### LINKED ISSUE #506: Project Proposal: Bril2x86
**What will you do?**
I want to learn about x86, so I will implement a compiler from core Bril to x86-64. It will be non-optimal.

**How will you do it?**
Incrementally! The strategy is to get something working asap, despite not being optimal. To reduce on scope, I will avoid register allocation and throw every variable onto the stack. Will translate from 1 Bril instruction at a time. 

The steps I currently plan on taking for the baseline are:
1. Handwrite a couple small x86 programs to familiarize myself with x86 and make sure I can get this compiling and running on my computer.
2. Implement just the main function with no control flow
	1. Implement const, print, ret
	2. Implement binop and uop like add, sub, etc
3. Implement control flow
4. Implement function calls

Stretch tasks after baseline in no order are:
- memory extension
- float extension
- basic register allocation

Will mostly be focusing on correctness as opposed to its performance. Will definitely run it against the benchmarks to see if results match. That said, it should probably not be slower than the interpreter, so I will do some timed runs to compare.

tean-lai:
Resolves #506

### DIFF
--- a/content/blog/2025-05-14-bril-to-x86.md
+++ b/content/blog/2025-05-14-bril-to-x86.md
@@ -8,22 +8,22 @@ name = "Tean Lai"
 The goal of [the project](https://github.com/tean-lai/bril-to-x86) was to build a Bril compiler targeting x86_64. The priority of the project was to implement something working, and not necessarily something optimal. So for this project, things like register allocation were out of scope. 
 
 ## Design and Implementation
-The current implementation was done in Python and it supports all the core features of Bril. 
+The current implementation was done in Python and it supports all the core features of Bril. I decided to go with Python despite its slowness because it's fast to iterate with, but I would expect the process of converting it to another language to be extremely smooth.
 
-The overall design of the compiler at a high-level is relatively simple. Because the design of Bril resembles assembly, this compiler just needs to do a linear pass converting each instruction at a time to the relevant code. The program does two passes: first it converts all the Bril instructions into another intermediate representation that is a one-to-one correspondence to an x86 instruction. Then the second pass will actually produce the assembly program and format it. The reason I decided to go with splitting it into two passes as opposed to printing the program directly was because there are multiple different syntaxes for x86, so the intention was if different syntax styles were to be supported, the changes would be more isolated. For the project, I implemented support for Intel syntax.
+The overall design of the compiler at a high-level is relatively simple. The design of Bril resembles assembly, so the compiler just needs to do linera passes converting each instruction at a time to the relevant code. The program does two main passes: first it converts all the Bril instructions into another intermediate representation that is a one-to-one correspondence to an x86 instruction. Then the second pass will actually produce the assembly program and format it. The reason I decided to go with splitting it into two passes as opposed to printing the program directly was because there are multiple different syntaxes for x86, so the intention was if different syntax styles were to be supported, the changes would be more isolated. For the project, I implemented support for Intel syntax.
 
 
 ## The Tricky Parts
-Although at a high level, things were simple, the details were tricky to get right. The first big hurdle was actually getting a compiled program to run. This required support for several Bril instructions, you would need to support constants, and either printing or returning in order to observe that some changes had happened. But it also required support for some calling conventions to be supported, otherwise the program wouldn't terminate. What helped got me over this hurdle proved to be quite useful in general for developing this compiler. I spent a lot of time modeling Bril programs as C programs and comparing my Bril -> x86 results with Clang's C -> x86 results. This helped me find stuff like the following snippet of text, which I would include at the beginning of every file for macs:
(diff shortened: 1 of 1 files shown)


content/blog/2025-05-14-bril-to-x86.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## The Tricky Parts
+Although at a high level, things were simple, the details were tricky to get right. The first big hurdle was actually getting a compiled program to run. This required support for several Bril instructions, you would need to support constants, and either printing or returning in order to observe that some changes had happened. But it also required support for some calling conventions to be supported, otherwise the program wouldn't terminate. What helped got me over this hurdle proved to be quite useful in general for developing this compiler. I spent a lot of time modeling Bril programs as C programs and comparing my Bril -> x86 results with Clang's C -> x86 results. This helped me find stuff like the following snippet of text, which I would include at the beginning of every file for macs:
```
sampsyo: > What helped got me over this hurdle proved to be quite useful in general for developing this compiler.

I don't understand this sentence… what exactly was helpful, and with what?

sampsyo: > for macs

To be more precise: "when generating code to target macOS"

content/blog/2025-05-14-bril-to-x86.md
```diff
@@ -0,0 +1,50 @@
...
+
+```
+.section	__TEXT,__text,regular,pure_instructions
+.build_version macos, 15, 0	sdk_version 15, 2
+```
+
+These directives seemed to vary quite a bit device-to-device; but I found comparing with a reference compiler was a reliable way of finding out what directives are necessary. Online sources often lack the specific things you need for your own device.
+
+It was also quite tricky to get printing right. I implemented Prof. Sampson's suggestion of linking the printing with a helper [rt.c file](https://github.com/sampsyo/bril/blob/main/brilift/rt.c) to avoid printing Booleans by hand. To get linking right, again it helped a lot to model the behavior I would expect in an equivalent c program, and guide my implementation to match that output.
```
sampsyo: c -> C

content/blog/2025-05-14-bril-to-x86.md
```diff
@@ -0,0 +1,50 @@
...
+.loop:
+    one: int = const 1;
+    print x;
+    y: int = add x one;
+    call @main y;
+.done:
+    ret;
+}
+```
+This program takes in a single int as a command line argument, and keeps printing its value and incrementing it until it hits 10. The tricky part comes from the fact that this function also calls itself. So this means the assembly code for this function must somehow handle both command line parsing and support proper calling conventions when it's called as a normal function. More concretely, in x86, it's expected that the main function's first and second parameters are `argc` and `argv` respectively. But we also want the main function's first function to be `x`.
```
sampsyo, line 43: You've stated the problem here, but what was the solution?

FWIW, the solution used in other native compilers is to add a "main wrapper" function that calls the true Bril `@main` function. The "main wrapper" parses the CLI arguments and then immediately calls `@main`.

content/blog/2025-05-14-bril-to-x86.md
```diff
@@ -0,0 +1,50 @@
...
+    y: int = add x one;
+    call @main y;
+.done:
+    ret;
+}
+```
+This program takes in a single int as a command line argument, and keeps printing its value and incrementing it until it hits 10. The tricky part comes from the fact that this function also calls itself. So this means the assembly code for this function must somehow handle both command line parsing and support proper calling conventions when it's called as a normal function. More concretely, in x86, it's expected that the main function's first and second parameters are `argc` and `argv` respectively. But we also want the main function's first function to be `x`. 
+
+## Correctness
+I wanted to automate checking for correctness in comparison to the interpreter, but I didn't get to it in time.
```
sampsyo: It's too bad that you didn't get to test for correctness, but is there a way you could give a brief summary of how complete you _think_ your implementation is? As in, how much of the Bril language do you think you covered, and what do you think is still not working? Just a couple of sentences of theorizing about this could help clarify to readers where you think things stand.

sampsyo (changes requested):
Hi, @tean-lai—I'm glad you got something working! It's too bad that you weren't able to do an empirical evaluation, but I hope you learned something about generating x86 assembly.

content/blog/2025-05-14-bril-to-x86.md
tean-lai: I ended up coming to the same solution, I could include that in the blog post

content/blog/2025-05-14-bril-to-x86.md
tean-lai: Woah, I'm sorry, that section is completely wrong.  I had rephrased it but I guess it didn't save. I didn't have time to compare to other AOT compilers like you suggested in the proposal PR, but I did test it for correctness and the output matches every core benchmark when compared to the interpreter.

sampsyo: Great news!

sampsyo:
I see you've marked this PR as a draft—let me know when it's ready for me to take another look and I'll publish it.

tean-lai:
Hi, I think this is ready for another look

sampsyo:
Looks good; I'll publish this now!

PR #548, opened 2025-05-30
rename directory to enable blog post to render

neel-patel-1:
directory was misnamed causing blog post not to show up on the site


sampsyo:
Good catch; thanks!

PR #551, opened 2025-07-08
Make path to image relative (instead of absolute)

InnovativeInventor:
I realized that the path to the image in my blog post is absolute instead of relative. This causes the image to not load properly when deployed to production behind the `2025sp` prefix (but works in development on my laptop). I changed the path to be relative to fix this issue.


sampsyo:
Good call!

PR #563, opened 2025-08-25
Add Jeremy as lead for tracemokey

jku20:
This paper looks cool and I'd like to give discussion lead a go on it. If anyone is interested in co-leading this, that'd be cool too!


sampsyo:
Great; you've got it!

PR #564, opened 2025-08-26
Fix date and add my paper

tf-mac:
Open to pairing up, but the paper looks interesting so I'd like to at least claim a part in it


sampsyo:
Sounds great; you've got it! And thanks for fixing my stale date!!

PR #565, opened 2025-08-26
Add Jeffrey as lead for Garbage Collection and Rc counting

jeffreyqdd:


sampsyo:
Sounds good; you've got it!

PR #566, opened 2025-08-26
Adding discussion leaders for profiling

Jacqueline-Wen:
adding Jacqueline, Serena, Maggie as discussion leaders of profiling


sampsyo:
Sounds good; you've got it!

PR #568, opened 2025-08-26
Attempt at fixing RSS link

pedropontesgarcia:
It turns out that the RSS link is at `atom.xml` and not at `rss.xml`. Without much knowledge about the website's build system, this is an attempt at fixing this issue.


sampsyo:
Oops; thanks!! I think this default changed in a recent version of Zola…

PR #573, opened 2025-08-29
Request leading the superoptimization paper

VesalBakhtazad:


sampsyo:
Hi @tf-mac—would you be OK with teaming up with @VesalBakhtazad on this one?

PR #584, opened 2025-09-02
Update discussion leaders Cynthia and Jonathan for Dynamic Languages

CynyuS:
- Added partners Cynthia Shao and Jonathan Brown


keikun555:
Awesome, thank you for being flexible!

PR #636, opened 2025-12-16
[Final Project] Principled Exploration of Tiling and Fusion for Tensor Programs

### LINKED ISSUE #609: Project Proposal: Principled exploration of loop transformations for tensor programs
**What will you do?**

We will design strategies for choosing optimal strategies for a number of loop transformations on tensor programs.

The main transformations we are interested in are tiling and fusion (both to reduce memory accesses), but if time and infrastructure permit, we are also interested in pipelining (for HLS targets), unrolling for parallelization, and optimizing memory layouts.

Ideally, we will design an approach that works on arbitrary graphs of MLIR `linalg.generic` ops, but we will start with a few simple kernels (matmul, batched matmul, elementwise, etc).

**How will you do it?**

Our primary tool for representing and transforming programs will be MLIR (specifically the `linalg` dialect and its transforms). Our primary tool for finding which transformations are valid and profitable will be reductions to well-studied combinatorial optimization problems for which optimized solvers already exist (most likely constraint programming, mixed-integer nonlinear programming, SMT, or similar).

As a first test, I already have a CP-SAT formulation for finding optimal tilings for a single matmul.

We will collect real programs or fragments of real programs in `mlir.linalg` (possibly through `torch-mlir` or similar projects), and use MLIR's built-in passes wherever possible to implement our transformations. We will then use some standard MLIR compilation pipeline to lower these programs to code which we can run and compare against various baseline implementations.
This will probably be CPU code generated through the standard MLIR-LLVM pipeline to begin with, although it would be nice to try it on some accelerators as well.

**Team members:**
Nate Young (nty3, @natetyoung) and Amanda Wang (@arw274)

natetyoung:
Closes #609


content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+  Amanda Wang and Nathaniel Young are both CS PhD students at Cornell.
+"""
+[[extra.authors]]
+name = "Nathaniel Young"
+link = "https://natetyoung.github.io"  # Links are optional.
+[[extra.authors]]
+name = "Amanda Wang"
++++
+
+# What was the goal?
```
sampsyo: You don't really need to include the headings from my list of questions in the syllabus! Since this is for external consumption, you can use whatever structure works for your post. In particular, jumping right in with this intro without any kind of heading would work just fine here.

content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+
+Our overall goal is easily understood by comparison: both this project and Orojenesis consider the problem of finding tiling and fusion strategies for minimizing the number of accesses made beyond some cache level, given the size of the cache. While Orojenesis uses exhaustive search with manually-imposed fusion constraints to find a set of strategies which are Pareto-optimal in terms of cache size and accesses, we wanted to come up with a solver-based approach which does not require manually specifying constraints per einsum type; we designed programs to find the best strategy in terms of accesses with a fixed cache size, although essentially the same programs could be used to construct an Orojenesis-style Pareto frontier in a solver which supports that.
```
sampsyo: The second sentence in this paragraph is very long—it might be easier to read if you break it up.

content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+For validity, ensure that for each dimension in the original einsum, the product of its factors in the tiled loop nest is at least its total size. To ensure we don't overflow the cache, the sum of buffer sizes in the cache should be at most a given capacity.
+
+So how can we find a tiled loop nest of optimal performance? As a first attempt at expressing the optimal tiling problem using CP-SAT, let’s try splitting each dimension into a fixed number of tiled loops (say 2) and then let the solver choose:
+1. The order of the factored loops
+2. The factor size in each loop
+3. The blocking level of each tensor
+This works fine; the positions of the loops and blocking levels can be represented using boolean variables. For a 6-loop schedule of matmul, for instance, we will have 36 boolean variables of the form "loop __i__ is assigned to position __j__" and 18 boolean variables of the form "position __j__ is inside the blocking level for tensor A".
```
sampsyo: Hmm… what about item 2 in the list above ("factor size," which I am assuming means the block size)? Those need integer variables, right?

content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+
+Note that this has some redundancy, though. We’re asking the solver to choose the exact position of every loop, but we only care whether each loop is inside or outside each blocking level. To get rid of this, we can try asking the solver to instead choose only a partial order over the loops, from which we can later extract any loop schedule consistent with that partial order. This formulation is available [here](https://github.com/natetyoung/tensor-program-dse/blob/main/po_einsum_cpsat97.py).
```
sampsyo: Could you perhaps add just a sentence or two of intuition here about how you encode the structure of the partial order (and presumably restrict it to be a complete partial order, i.e., connected and acyclic)? I don't think it's entirely obvious.

content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+
+This is still a little redundant, though. We’re not making the solver choose a total order, but it's still _allowed_ to choose one, so it might be spending extra time proving to itself that the exact order doesn't matter. Another issue is that we have a few poorly-justified assumptions in our model. Why 2 factors of each dim specifically? Can we assume that we only have 1 factor for a reduction dimension (this makes the solver considerably faster, but is only justified by empirics).
```
sampsyo: I don't think I fully understand this:

> Can we assume that we only have 1 factor for a reduction dimension (this makes the solver considerably faster, but is only justified by empirics).

Maybe the problem is that I'm not sure what a "reduction dimension" is?

content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+
+We can take our approach further in order to solve both of these problems at once. If all that matters is whether each factor is inside or outside each tiling level, let's _only choose the order of the tiling levels_ and then choose how much of each dim goes into each "bucket" created between them. In the tiled matmul example above, for instance, we have 4 buckets. The outermost contains factors `M/128` and `N/128` of `M` and `N` respectively; the second bucket has all of dim `K`; the third has factor `128` of `M`, and the innermost bucket has factor `128` of `N`. This formulation is available [here](https://github.com/natetyoung/tensor-program-dse/blob/main/cb_einsum_cpsat97.py).
```
sampsyo: I still think this could use some further explanation to get clear about how the encoding works. I understand the reference to the example, but I don't yet see a direct description of which constraints need to be constructed to reflect these "buckets." What does it mean to "only choose the order of the tiling levels"? I think some step-by-step development of some example constraints could help a lot…

content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+
+# Success
+
+We evaluated our project on some real-world tensor operations against -O3 without vectorization or loop unrolling. We follow the evaluation criteria used in Orojenesis by generating some C code with dummy values, which models a chain of 3 einsums (Q_proj, QK, QKV) from an inference pass through a GPT-style block. Notice that the output of each einsum is fed as input to the next, allowing for loop fusion. To vary the sizes of these operations, we adjusted the corresponding "sequence length" and "model dimensions" used to generate the einsums in the test code, and used a spread of different sizes: (`seq`, `dim`) = (`32, 128`), (`64`, `256`), (`128`, `512`), (`1024`, `4096`). We refer to them as `tiny`, `small`, `med`, and `large`, respectively. Note that the dimensions under `large` actually correspond to realistic dimensions used in the [Cerebras GPT-3-6.7B](https://huggingface.co/cerebras/Cerebras-GPT-6.7B) block.
```
sampsyo: > against -O3 without vectorization or loop unrolling

I suppose you're leaving it implicit that the baseline is imperative implementations in C or C++. Can you mention which C compiler you used?

sampsyo: I think you haven't mentioned yet that your compiler can generate C code that implements a given schedule for a given einsum. (Assuming that it can? And that you're not hand-writing each program according to the solver output.) Can you describe that a bit earlier, in your implementation section?

sampsyo: For this example program, it sounds like your baseline is a completely un-tiled, un-fused implementation. Surely there exists an "expert" implementation of this stuff, tuned by someone for a particular target? Is that something that you could compare against?

content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+
+More concretely, we used the "full-tree" approach to decide the optimal tiling and fusion for the input chain, and then generated C code with accordingly tiled loops. For each input size, we ran our optimization with capacities `4*1024`, `8*1024`, `16*1024` (in number of floats) to generate three separate optimized programs. To gauge how well our optimizer fared against code compiled with standard optimizations, we timed our solver-optimized test code and naive test code, both compiled with -O3 with vectorization and loop unrolling disabled. The table below gives the time elapsed for each test size + optimization option, averaged over 5 runs.
```
sampsyo, line 106: It makes sense to disable vectorization and unrolling for a basic comparison, but what happens if you turn these back on? Is one or the other more effectively vectorizable?

content/blog/2025-12-16-tiling-fusion-search.md
```diff
@@ -0,0 +1,135 @@
...
+med  |opt |4096|0.028004
+med  |opt |8192|0.025999
+med  |opt |16384|0.028292400000000002
+large |unopt |N/A|62.1979504
+large |opt |4096|11.396286800000002
+large |opt |8192|17.026373
+large |opt |16384|92.80004540000002
+
+
+Our optimized code compiled with -O3 consistently outperformed the naive code compiled with -O3, which we consider a success. The exception is the large size tiled to capacity `16*1024`, which performed worse than the unoptimized code. We're not sure why this is, but we suspect that this could be approaching the cache size of the architecture we were running on, which for some cores is 64k, corresponding to 16k floats. One thing to note is that while our optimized C program does already tile the loops to fuseable shapes, it doesn't directly fuse them. We also experimented with manually fusing the code, but found that the times didn't change, suggesting that -O3 was probably fusing them automatically, as expected.
```
sampsyo: > suggesting that -O3 was probably fusing them automatically

It would be great to confirm this by looking at the compiler's output.

sampsyo (changes requested):
Cool! It's awesome that this worked out, and that the CP-SAT programs turn out to be so efficiently solvable for this use case. I'd be excited to see how a full-scale evaluation of this tool pans out, and how it compares against other DSE-style approaches for generating optimized loop nests.

I have a few low-level writing suggestions—could you please address those and then use GitHub's "request review" button to let me know when you're ready to publish?

sampsyo:
Just a friendly reminder to please address a few of my writing comments above so we can publish your post.

sampsyo:
Looking great! Thanks!!

Issue #616, opened 2025-11-05
Project Proposal: Modeling Stateful Accelerators in Allo


sunwookim028:
**What will you do?**
We will extend the Allo programming model to represent stateful hardware accelerators. [Allo](https://dl.acm.org/doi/full/10.1145/3656401) is a Python-embedded Domain Specific Language capable of both designing and programming hardware accelerators. Allo has attractive features as a hardware description language, such as productivity and modularity, but currently the hardware it models is stateless, i.e., the driver cannot assume a persistent state across invocations. This critically limits the practicality of Allo, since many accelerators are stateful. For instance, we need the state of partial sum, to compute dot products of large vectors using a smaller, fixed-sized dot product accelerator. Hence, our project will be a critical bridge to model practical ASICs in Allo.

**How will you do it?**
We will implement the notion of state in Allo programming model, by extending Allo types, Allo IR, and the backend codegen targeting Vitis HLS. Concretely, we will add the followings. First, prev(n) field to Allo types to store past states (history) and access one of them. Second, using MLIR features to represent global variables to extend the Allo IR for stateful objects. Lastly, the lowering pass to Vitis HLS, targeting the static keyword in Vitis HLS to implement the state at the actual hardware level.
<img width="1200" height="675" alt="Image" src="https://github.com/user-attachments/assets/3a8d2e05-e172-45a0-b53c-160514b94b9a" />

Our current task is to develop tractable HLS designs, more detailed weekly roadmap to implement all components and division of labor. We plan to deliver the exact code-level targets by EOD Nov 11.

**How will you empirically measure success?**
We will concretely scope our target hardware in 3 milestones: scalar accumulator, vector accumulator, then something more practical (TBD, for instance outer-product GEMM engine or the dot product engine; we'll define a case which involves more past states and achievable with Vitis HLS). We will build target representation in HLS first, verify the design with the target invocation scenario in a simulation or on a FPGA board (to check the design is not absurdly inefficient), before we decide to target that.
We will implement by extending the existing Allo and MLIR infrastructure. We will unit-test the type to IR pass and IR to Vitis HLS passes utilizing the existing infrastructure as well.
We will evaluate our progress in a weekly basis to make adjustments.

**Team members:**
@Fangtangtang 
As we discussed with Prof. Sampson, we are doing this for another course project as well (ece6775). There are 3 other students in that project team, and we'll clarify the individual contributions in both our roadmap due Nov 11 and the final deliverables, as necessary.

[...]

PR #637, opened 2025-12-16
Final project, Stateful Allo

sunwookim028:


sunwookim028:
This closes #616.

content/blog/2025-12-16-stateful-allo.md
```diff
@@ -0,0 +1,115 @@
...
+# Design and Implementation
+<img src="./2025-12-16-stateful-allo/overview.png" alt="allo to mlir to hls" width="310"/>
+This is the overview of our compilation flow.
+We'll use this scalar accumulator example to explain each step.
+
+```python
+# A kernel that accumulates values across invocations
+def stateful_kernel(x: int32) -> int32:
+    # 'acc' retains its value between calls
+    acc: stateful(int32) = 0
```
sampsyo, line 40: It's an interesting choice to use a type qualifier to mark state like this. One thing that's a little weird about it is that this no longer matches Python semantics; if you were to run this code in CPython, the variable would be "fresh" in every invocation of `stateful_kernel`.

A typical Python programmer would accomplish this statefulness like this:

```py
acc = 0
def stateful_kernel(x: int32) -> int32:
    global acc
    acc += x
    return acc
```

I understand that this pattern might not be feasible to recognize in Allo's Python frontend, but it occurred to me to mention as an alternative.

content/blog/2025-12-16-stateful-allo.md
```diff
@@ -0,0 +1,115 @@
...
+    %0 = memref.get_global @__stateful_stateful_kernel_acc_1 : memref<i32>
+    // 3. Load/store operations using the handle
+    %1 = affine.load %0[] {from = "acc"} : memref<i32>
+    // ... arithmetic operations ...
+    affine.store %5, %0[] {to = "acc"} : memref<i32>
+    return %6 : i32
+  }
+}
+```
+They are then translated into global variables in IR, marked with a naming pattern `_stateful`.
```
sampsyo: Why do they need to be marked with a special naming convention? Does it not suffice to have declared this variable with `memref.global`?

content/blog/2025-12-16-stateful-allo.md
```diff
@@ -0,0 +1,115 @@
...
+  int32_t v7 = v6;    // L9
+  __stateful_stateful_kernel_acc_1 = v7;    // L10
+  *v1 = __stateful_stateful_kernel_acc_1;    // L11
+}
+```
+Then they are translated into static variables in the backend.
+HLS codegen emits static keyword for such marked stateful variables.
+
+# Additional Evaluation
+By measuring execution time, we discovered that the overhead of copying the state in host adds ~20% overhead.
```
sampsyo: This needs a lot more detail. Some questions it raises:

* What program did you use?
* What is "overhead"? In what units?
* What toolchain did you use for the evaluation?
* How did you measure that overhead?
* Is this a necessary cost (to implement specific functionality) or something that could be eliminated?

content/blog/2025-12-16-stateful-allo.md
```diff
@@ -0,0 +1,115 @@
...
+
+Also, we demonstrated a programmable accelerator design here.
```
sampsyo: The listing of Python code is not very helpful by itself. What do you want people to take away from looking at this code?

sampsyo (changes requested):
Hi, @sunwookim028—congratulations on getting this to work. It sounds like a pretty straightforward implementation in the Allo toolchain.

Your evaluation needs more detail. It currently doesn't describe much about what you actually did or how you measured it. We also don't get much description of how you checked whether your implementation was correct. Is there anything else you can say about what you evaluated?

When you are ready for me to take a look, please use GitHub's "request review" button and it should send me a notification.

sampsyo:
Just a friendly reminder to please address a few of my writing comments above so we can publish your post.

sunwookim028:
Sorry I missed this and thank you for valuable your comments. I'll address by until tomorrow.

sunwookim028:
Professor Sampson, I've addressed your feedback. My sincere apologies for the long delay. I wish to close this loop properly. Thank you again.

PR #638, opened 2025-12-16
[Final Project] Global Value Numbering

### LINKED ISSUE #613: Project Proposal: Global Value Numbering
**What will you do?**
We will implement global value numbering (GVN) for Bril, following the SSA-based algorithmic structure described by Briggs et al. GVN generalizes local value numbering to handle redundant computations across basic blocks, allowing elimination of common subexpressions, copies, and dead code more effectively. We will build upon our previous transformation infrastructure from earlier assignments.

**How will you do it?**
We plan to implement GVN in Python, leveraging the SSA form already supported in Bril. Our design will center around constructing and traversing the dominator tree, maintaining a scoped or unified hash table of expression equivalences, and applying algebraic simplifications and constant folding. Once redundancy detection is complete, we will rewrite redundant instructions to use existing values and remove unnecessary computations. For correctness, we will run our implementation on all programs in the Bril benchmark suite provided in the repository.

We will evaluate our implementation in two aspects:

* **Correctness:** Verify that all optimized programs produce identical outputs to their originals by running the full Bril benchmark suite and comparing outputs.
* **Performance:** Measure improvement through both static instruction count reduction and dynamic operation count/runtimes on benchmark programs before and after applying GVN.

**Next steps**
After completing the core GVN implementation, we plan to extend the optimizer to include:

* Constant propagation and folding
* Copy propagation
* Algebraic simplification

**Team members:**
@SyphonArch (Jake Hyun)
@tobiwg (Tobi Weinberg)
@adnan-armouti (Adnan Armouti)

SyphonArch:
Closes #613


content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+name = "Jake Hyun"
+[[extra.authors]]
+name = "Tobi Weinberg"
+[[extra.authors]]
+name = "Adnan Armouti"
++++
+
+# Optimizing Bril with Global Value Numbering
+
+*By Jake Hyun, Tobi Weinberg, and Adnan Armouti*
```
sampsyo: You don't need to include the title and authors; those will be displayed on the page automatically.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+## 1. Project Goal
+
+The goal of this project was to implement **Global Value Numbering (GVN)**, a powerful compiler optimization that eliminates redundant computations by identifying expressions that compute the same value. GVN is more powerful than local value numbering because it operates across basic block boundaries using the Static Single Assignment (SSA) form and dominator tree traversal.
+
+Specific objectives included:
+
+1. **Correctness**: The optimized programs must produce identical outputs to the original unoptimized programs for all inputs.
+2. **Effectiveness**: The optimization should reduce both static (code size) and dynamic (runtime) instruction counts.
+3. **Faithfulness to Literature**: Implement GVN using real phi nodes (as described in the original Cooper & Briggs algorithm) rather than simpler alternatives like Bril's get/set encoding.
```
sampsyo: Any particular reason for this goal?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+4. **Minimal Overhead**: The SSA conversion and lowering process should not introduce more instructions than it removes.
```
sampsyo: Including this bullet raises some questions. The main goal here is GVN, but it sounds like you're implying that it also includes into-SSA and out-of-SSA transformations in the same scope? Again, I suppose I'm wondering: any particular reason to do that? Keeping things focused on just the GVN (when we're already in SSA) would keep stuff simpler, and it would help separate two sources of overhead/improvement.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+SSA Lowering (from_ssa.py with fixed copy sequentialization)
+    ↓
+Dead Code Elimination (tdce.py)
+    ↓
+Optimized Program
+```
+
+### 2.1 GVN Algorithm {#21-gvn-algorithm}
+
+#### 2.1.1 Core Data Structures
```
sampsyo: I don't think it's helpful to include the Python listings for the implementations of these data structures. Can you please delete this entire section and replace it with a few sentences that capture what the important data structures _do_ at a high level?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+                key_args = [key_args[1], key_args[0]]
+        return (op, ty, *key_args)
+
+    return None
+```
+
+Key design decisions:
+- **Commutative normalization**: For operations like `add`, `mul`, `and`, we sort arguments by VN so that `a + b` and `b + a` produce the same key.
+- **Relational normalization**: We flip `gt`/`ge` to `lt`/`le` by swapping arguments, reducing the number of distinct expression patterns.
+- **Type inclusion**: Types are included in the key because operations on different types (e.g., `int` vs `float`) are not interchangeable.
```
sampsyo: These do not seem like "key design decisions." They are some miscellaneous observations about how your implementation works.

With apologies if this is too forward, did an LLM write this bulleted list? It's fine to use LLMs in your writing, but stuff like this is characteristic of the "low-content, high-marketing" stuff that an LLM can produce if you're not careful. Just one sentence of your high-level takeaways would be much more valuable than many bullet points like this.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+#### 2.1.4 Conservative Handling
+
+The implementation is conservative in several ways:
+
+1. **Effectful Operations**: Any instruction with side effects (memory operations, function calls, I/O) clears all expression scopes:
+   ```python
+   if not is_pure(instr):
+       ctx.clear_all_scopes()
+   ```
```
sampsyo: Again, I don't think you need the Python code to make your point here.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+2. **Loop Backedges**: Phi nodes at loop headers receive fresh value numbers because their values may change on each iteration:
+   ```python
+   if has_backedge:
+       vn = ctx.new_vn()  # Conservative: fresh VN for loop header phis
+   ```
+
+3. **Division Excluded**: Integer division is excluded from pure operations because speculating a division could trap on divide-by-zero.
```
sampsyo: This seems like it's part of item 1?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+```
+
+Key design decisions:
+- **Commutative normalization**: For operations like `add`, `mul`, `and`, we sort arguments by VN so that `a + b` and `b + a` produce the same key.
+- **Relational normalization**: We flip `gt`/`ge` to `lt`/`le` by swapping arguments, reducing the number of distinct expression patterns.
+- **Type inclusion**: Types are included in the key because operations on different types (e.g., `int` vs `float`) are not interchangeable.
+
+#### 2.1.3 Dominator Tree Walk
+
+The core GVN algorithm performs a depth-first traversal of the dominator tree:
```
sampsyo: Can you please describe your algorithm/implementation with words? You can link to the code if you want to make it easy to browse the actual Python, but the far more valuable thing to do here would be to just discuss the high-level takeaways you want people to pay attention to.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+The implementation is conservative in several ways:
+
+1. **Effectful Operations**: Any instruction with side effects (memory operations, function calls, I/O) clears all expression scopes:
+   ```python
+   if not is_pure(instr):
+       ctx.clear_all_scopes()
+   ```
+
+2. **Loop Backedges**: Phi nodes at loop headers receive fresh value numbers because their values may change on each iteration:
```
sampsyo: I don't understand how this is conservative (nor exactly what the purpose here is). What does it mean to receive "fresh value numbers"? Don't you need this for every phi-node, regardless of whether a backedge is involved?

In what sense is this "conservative" vs. just needed for the basic correctness of the algorithm?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+| Benchmark | Original | After Minimal SSA | Bloat |
+|-----------|----------|-------------------|-------|
+| fizz-buzz.bril | 57 | 149 | +162% |
+| Loop-heavy programs | N | ~2.5N | +150% |
+
+The root cause: minimal SSA creates phi nodes for **every** variable at dominance frontiers, even if the variable is never used after that point.
+
+#### 2.2.2 Pruned SSA Solution
+
+**Pruned SSA** only places phi nodes for variables that are actually **live** at the join point. This requires liveness analysis:
```
sampsyo: Again, just one or two sentences of your high-level insights would be worth more than any code excerpt.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+| Benchmark | Phi Nodes (Minimal) | Phi Nodes (Pruned) | Reduction |
+|-----------|---------------------|--------------------|-----------|
+| fizz-buzz.bril | 73 | 1 | 98.6% |
+| loopfact.bril | 12 | 2 | 83.3% |
+| Average | N | ~0.15N | 85%+ |
+
+This dramatically reduced the SSA overhead, making the optimization effective.
+
+### 2.3 SSA Copy Propagation {#23-ssa-copy-propagation}
```
sampsyo: Is this part of GVN? If not, maybe de-emphasize it…

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+    for instr in instrs:
+        if "args" in instr:
+            instr["args"] = [ultimate_src(a) for a in instr["args"]]
+```
+
+#### 2.3.2 Key Insight: Phi Argument Propagation
+
+A subtle but important point: we **can** propagate copies into phi arguments in SSA form. This is safe because in SSA, every definition dominates all its uses. If `x.1 = id y.1`, then wherever `x.1` appears (including in phi nodes in successor blocks), we can replace it with `y.1`.
+
+This is a key advantage of performing copy propagation **before** SSA lowering.
```
sampsyo: Is this a restatement of the observation "copy propagation is always safe"? I don't think I understand what "propagate copies into phi arguments" means; phi-nodes don't seem very relevant here.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+### 2.4 Dominator-Tree Copy Elimination {#24-dominator-tree-copy-elimination}
+
+#### 2.4.1 Design
+
+After SSA copy propagation, there may still be `id` instructions that weren't eliminated because they define phi destinations or have other uses. The dominator-tree copy elimination pass walks the dominator tree with a **rename environment**:
```
sampsyo: This doesn't really explain the goal… what exactly are we going to recognize and change here?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+**Example:**
+```
+     B0: x = 5
+    /  \
+   B1   B2
+   y=x  z=x
+```
+
+If we're at B0 and establish `x → 5`, both B1 and B2 see this rename. But renames in B1 don't affect B2 because they receive separate copies of the environment.
```
sampsyo: While this detail makes sense, I'm still confused about exactly which `id`s you are hoping to eliminate here… can you focus on that instead of this relatively-straightforward correctness argument?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+#### 2.4.3 Preserving Phi Destinations
+
+We explicitly preserve phi destinations because they are **merge points**. Even if a phi produces `x = phi(y, z)` where `y` and `z` have the same value, we can't eliminate the phi — it defines a new SSA name that may be used later.
```
sampsyo: I don't understand this argument. If `y` and `z` are guaranteed to hold the same value, can't we just use `y` in place of `x` everywhere? Is the problem that `y` might be undefined?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+### 2.5 Fixed Parallel Copy Sequentialization {#25-fixed-parallel-copy-sequentialization}
+
+#### 2.5.1 The Problem
```
sampsyo: This is called the "swap problem," and it is discussed in the 6120 lecture on SSA.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+Naive sequentialization fails:
+```
+x = y   # x gets y's value
+y = x   # y gets... x's new value (wrong!)
+```
+
+#### 2.5.2 The Algorithm
+
+The fixed algorithm handles dependencies correctly:
```
sampsyo: You have described a potential problem, but you haven't described _what actually has this problem_. What are we fixing here?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+```
+# Parallel: (x, y, z) := (y, z, x)  # 3-way cycle
+# Step 1: No safe copy (all dests are sources)
+# Step 2: Save x to tmp: tmp = x
+# Step 3: Update: (x, y, z) := (y, z, tmp)
+# Step 4: Now z := tmp is safe, emit it
+# Step 5: Now y := z is safe, emit it
+# Step 6: Now x := y is safe, emit it
+# Result: tmp=x; z=tmp; y=z; x=y
+```
```
sampsyo: With apologies, I don't understand this at all… the notation is never really defined.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+### 2.6 Real Phi Node Representation {#26-real-phi-node-representation}
+
+#### 2.6.1 Why Real Phi Nodes?
+
+The original GVN paper by Cooper and Briggs uses real phi nodes in SSA form. Bril provides an alternative approach using `get`/`set` operations with labels, which is easier to implement but doesn't faithfully represent SSA.
```
sampsyo: What do you mean by "doesn't faithfully represent SSA"?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+We chose to implement **real phi nodes** for several reasons:
+
+1. **Algorithmic Fidelity**: The GVN algorithm naturally operates on phi nodes as first-class instructions
```
sampsyo: What do you mean by "first-class instructions"? Aren't `get` and `set` just as eligible?

There is a straightforward way to translate one into the other, so I don't see the advantage in terms of similarity to the paper.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+2. **Value Numbering of Phis**: We can assign value numbers to phi nodes and detect redundant phis
```
sampsyo: Why is this not true of `get` and `set`? You can assign value numbers to `get` instructions just as you can to any other instruction.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+3. **Learning Value**: Understanding real SSA conversion is pedagogically important
```
sampsyo: This is unsubstantiated. What makes one more "pedagogically important" than the other?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+```json
+{
+  "op": "phi",
+  "dest": "x.3",
+  "type": "int",
+  "args": ["x.1", "x.2"],
+  "labels": ["B1", "B2"]
+}
+```
```
sampsyo: This seems to be the standard "old SSA" extension? Please try not to give the impression that you invented these instructions…

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+After implementing all previous optimizations, we discovered that **63 out of 114 benchmarks were still worse** in both static and dynamic instruction count. The root cause was **phi lowering overhead**.
+
+When converting out of SSA form, each phi node becomes multiple `id` (copy) instructions:
+
+**Example: triangle.bril**
+```
+Original:    14 instructions, 0 id copies
+After SSA:   12 instructions (2 phis)
+After GVN:   12 instructions (phis intact)
```
sampsyo: I'm confused… GVN didn't optimize at all? That seems like the core problem here, not the SSA conversion itself…

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+    ctx.insert_expr(phi_key, vn)
+    return vn
+```
+
+This allows GVN to:
+- Recognize trivial phis (all args same) and propagate the value
+- Detect redundant phis (same pattern of arg VNs)
+- Be conservative at loop headers
+
+### 2.7 Copy Coalescing {#27-copy-coalescing}
```
sampsyo: This section is a VERY long way to introduce an extremely simple copy-prop pass… can you summarize it in a single sentence?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+#### 2.7.6 Critical Bug Fix: Function Parameters
+
+Our initial implementation had a subtle bug: it would rename function parameters on subsequent coalescing iterations, causing undefined variable errors.
+
+**Example of the bug:**
+```
+Iteration 1: input.1 = id input  →  rename input.1 → input ✓
+Iteration 2: input.2 = id input  →  rename input → input.2 ✗
+```
```
sampsyo: This notation is not defined, making it impossible to follow.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+| Metric | Before Coalescing | After Coalescing | Change |
+|--------|-------------------|------------------|--------|
+| Static Ratio (GM) | 0.9293x | **0.8477x** | +8.8% better |
+| Dynamic Ratio (GM) | 0.9152x | **0.8167x** | +9.6% better |
+| Benchmarks worse (both) | 63/114 | **30/114** | -52% |
+| Benchmarks improved (static) | 38/114 | **50/114** | +32% |
+
+Copy coalescing transformed the optimizer from a net negative to a clear positive on the majority of benchmarks.
+
+### 2.8 Phase 2 Optimizations: SSA Overhead Elimination {#28-phase-2-optimizations-ssa-overhead-elimination}
```
sampsyo: Can you please summarize this section with a few simple sentences? It seems like there are 3 low-level hacks applied to the SSA translation, and the high-level idea about how they work is far more important than the low-level detail about exactly how they are implemented.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+**Example:**
+```bril
+@convolve(arr.3: ptr<int>, ...) {  ← Parameter renamed from arr
+  loc.1: ptr<int> = ptradd arr.3 offset;
+  ...
+}
+```
+
+TDCE didn't know `arr.3` was a parameter and removed the `ptradd` instruction.
```
sampsyo: Why does this have to do with renaming? Wouldn't this have been a problem with the old names too?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+# B3:
+use(x)
+```
+
+If we propagate `x → a` and eliminate the copy, the program breaks because `a` might not be available in B3 (it came from a different predecessor).
+
+**The Solution**: Perform copy propagation **while still in SSA form**, before phi lowering. In SSA, every definition dominates all its uses, so copy propagation is always safe.
+
+### 3.4 Entry Block Creation for Label-First Functions
```
sampsyo: This is a pretty good example: this seems like an extremely small change. Was this really one of the _hardest things you encountered_?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+| Static Ratio (GM) | 0.8474x | **0.8397x** | +0.9% better |
+| Dynamic Ratio (GM) | 0.8165x | **0.8142x** | +0.3% better |
+| Benchmarks worse (both) | 30/114 | **22/114** | -27% |
+| Benchmarks improved (static) | 47/114 | **53/114** | +13% |
+
+These optimizations further reduced the number of regressions while maintaining 100% correctness on all 114 benchmarks.
+
+---
+
+## 3. Hardest Parts to Get Right
```
sampsyo: Sorry to be blunt again, but it is obvious that an LLM wrote this section. None of these seem like very hard things; it's a catalog of random technical choices and bugs, most of which have already been discussed in earlier sections. Can you please just reflect a tiny bit about your personal experience with this project and write a few short sentences about what you found challenging?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+**Result**: This reduced "worse in both" from 63/114 to 30/114 benchmarks, and improved the static ratio from 0.93x to **0.85x**.
+
+---
+
+## 4. Empirical Evaluation
+
+### 4.1 Methodology
+
+**Benchmarks**: We evaluated on 123 Bril benchmarks across 5 categories:
```
sampsyo: Are these all the benchmarks from these categories?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+- `mem`: 31 memory operation benchmarks
+- `mixed`: 4 combined operation benchmarks
+- `long`: 2 large/complex programs
+
+**Correctness**: Every optimized program was verified to produce identical output to the original for all provided inputs.
+
+**Metrics**:
+- **Static Instruction Count**: Total instructions in the program (code size)
+- **Dynamic Instruction Count**: Instructions executed at runtime (performance)
+- **Geometric Mean Ratio**: GM of (optimized/original) across benchmarks
```
sampsyo: Ratio of what?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+| float | 18 | 0.7371 | 0.6948 |
+| mem | 31 | 0.9506 | 0.9591 |
+| mixed | 4 | 0.9728 | 0.9527 |
+| long | 2 | 0.8870 | 0.7373 |
+
+The `float` benchmarks show the best improvement (26.3% static, 30.5% dynamic) due to floating-point computations having many redundant expressions. The `core` benchmarks also show strong improvements (18.5% static, 21.2% dynamic). Memory-heavy benchmarks (`mem`) show minimal improvement because memory operations are effectful and cannot be optimized by GVN.
+
+### 4.5 Comparison with Get/Set SSA Implementation
+
+We compared our real phi node implementation against an alternative approach using Bril's `get`/`set` label-based syntax. While Bril supports `get`/`set` for SSA representation, the Cooper/Briggs paper assumes real phi nodes. We initially implemented GVN using `get`/`set`, then re-implemented it with real phi nodes for comparison.
```
sampsyo: Oh wow, I am surprised to learn at this late point that you actually implemented GVN _twice_ for the two different forms of SSA. What motivated you to do this? It seem like this is a big deal that deserves some emphasis…

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+| Metric | Our Implementation (Real Phis) | Get/Set Implementation | Improvement |
+|--------|--------------------------------|------------------------|-------------|
+| Total Static (62 common tests) | 2,262 | 2,313 | +51 instructions better |
+| Static Reduction | 12.0% | 10.0% | +2.0% better |
+
+The final phi node implementation outperforms the `get`/`set` approach by 51 instructions on the common benchmark set, primarily because real phis expose phi node structure to the GVN algorithm, enabling better redundancy detection and more precise dominator-tree analysis.
```
sampsyo: I don't immediately see why this is the case. Can you present a little bit of an argument to justify your claim?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+The following figures are generated by `generate_figures.py`:
+
+**GVN Optimization Results:**
+1. **static_improvement_bar.png**: Bar chart showing top 20 benchmarks by static instruction reduction
+2. **static_ratio_histogram.png**: Distribution of static instruction ratios across all benchmarks
+3. **dynamic_improvement_bar.png**: Bar chart showing top 15 benchmarks by dynamic instruction reduction
+4. **category_comparison.png**: Grouped bar chart comparing improvements by benchmark category
+5. **static_vs_dynamic_scatter.png**: Scatter plot showing correlation between static and dynamic improvements
+6. **summary_table.png**: Summary statistics table
+7. **waterfall_savings.png**: Waterfall chart showing cumulative instruction savings
```
sampsyo: I don't think we need the filenames… can you just associate the descriptions with the images themselves?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+This section compares our implementation against the [Cooper/Briggs 1997 paper](https://www.cs.tufts.edu/~nr/cs257/archive/keith-cooper/value-numbering.pdf) on value numbering.
+
+### 7.1 Alignment with the Paper
+
+Our implementation addresses all the core requirements of the DVNT (Dominator-based Value Numbering with Tables) algorithm:
+
+| Issue | Status | Evidence |
+|-------|--------|----------|
+| **Real phi nodes** | **FIXED** | Uses `"op": "phi"` instructions. `to_ssa.py` creates real phis, `from_ssa.py` lowers them. |
```
sampsyo: Is this really an "issue"?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+| **DCE after GVN** | **FIXED** | `tdce.trivial_dce_func(func)` runs after GVN |
```
sampsyo: This seems to be about DCE, which is separate from GVN?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+| **Dominator tree traversal** | **MATCHES PAPER** | `gvn_walk` does recursive dom-tree walk |
+| **Scoped expression tables** | **MATCHES PAPER** | `expr2vn_stack` with `push_scope`/`pop_scope` |
+| **Expression canonicalization** | **MATCHES PAPER** | Commutative sorting, relational flipping |
+| **Meaningless phi detection** | **MATCHES PAPER** | Checks if all phi args have same VN |
+| **Redundant phi detection** | **MATCHES PAPER** | Checks if phi pattern already in expression table |
```
sampsyo: I think I'm confused about what this table is trying to say in general… what are we supposed to glean from this in terms of the mapping onto parts of the paper?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+### 7.2 Conservative Design Choices
```
sampsyo: This was already covered?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,1350 @@
...
+
+| Aspect | Paper | Our Implementation | Impact |
+|--------|-------|---------------------|--------|
+| **Memory operations** | Suggests memory SSA or alias analysis | Conservative: clears expression table on effectful ops | Safe but limits optimization on `mem/` benchmarks |
+| **Loop headers** | Can do iterative refinement | Conservative: fresh VN for backedge phis | Safe but may miss some loop-carried redundancies |
+
+### 7.3 Features Not Implemented
```
sampsyo: It seems like this only takes one sentence to say… can you eliminate this whole section by just saying somewhere "we implement DVNT from the paper, not SCC-VN"? This section seems like WAY too much detail for something that you didn't implement…

sampsyo (changes requested):
Hi, folks. It's very impressive that you got end-to-end speedups using GVN, even accounting for SSA conversion overhead. That's a significant success that you should be proud of.

I am concerned about the current state of your writeup here. As you mention in the end of the report, an LLM wrote some of the text here. That is not necessarily a problem by itself, but the LLM has done a few serious disservices to your report:

* It includes lots of irrelevant code snippets instead of just describing how your project works.
* It goes into extreme detail on low-level implementation concerns that are not connected to the main idea of the project.
* It spends the majority of the report talking about tweaks to SSA conversion, and comparatively little time on the GVN optimization itself.
* It is missing much high-level reflection on your personal experience of doing the project.

The result is a post that is very, very long but with a low level of insight-per-word. Would you all be willing to think through the sections that are really necessary and collapse things into a more compact form that might get the "big idea" across more efficiently?

SyphonArch:
Thanks for the detailed feedback. We’ll revise accordingly and post an updated draft soon.

SyphonArch:
We have revised the report based on your feedback. Originally we thought that this was meant to serve as a detailed log of our development - now it should read more like a summary of key ideas. We took out most sections with too much details or divergence.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+"""
+[[extra.authors]]
+name = "Jake Hyun"
+[[extra.authors]]
+name = "Tobi Weinberg"
+[[extra.authors]]
+name = "Adnan Armouti"
++++
+
+# Optimizing Bril with Global Value Numbering
```
sampsyo, line 16: You don't need this heading here; the site will already render your post's title above the text. (You can preview your post with [Zola](https://www.getzola.org) to make sure it renders correctly.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+This project implements **Global Value Numbering (GVN)** for Bril, following the dominator-tree–based, hash-table approach described by Cooper, Briggs, and Simpson. Our goal was to eliminate redundant computations across basic blocks while preserving correctness, and to evaluate whether GVN can produce *net* performance wins in Bril despite the overhead introduced by SSA conversion and lowering.
```
sampsyo, line 18: Could you please link to the paper from this first mention of it?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+## 1. Overview of the GVN Approach
+
+We implement the **dominator-tree value numbering with tables (DVNT)** variant of GVN. The algorithm traverses the dominator tree in depth-first order while maintaining a scoped table that maps canonicalized expressions to value numbers. Expressions computed in a block are visible to its dominated descendants but not to siblings, ensuring that table membership corresponds to dominance-based availability.
+
+Each SSA name is assigned a value number (VN). Two expressions receive the same VN if and only if they are proven equivalent by the algorithm. When a redundant expression is encountered, it is rewritten as a copy from a dominating representative when safe; otherwise, the expression is retained but unified at the VN level so that future redundancies may still be detected.
+
+We restrict value numbering to **pure operations** (arithmetic, logical, and comparison operations). Any effectful instruction—such as memory operations or calls—is treated as a barrier and conservatively clears the expression table. This choice is safe given Bril’s lack of alias or purity information, but it limits optimization opportunities in memory-heavy programs.
```
sampsyo, line 30: I'm not quite sure I understand your treatment of side effects here. To boil it down: why is it necessary to clear the expression table? Why doesn't it suffice to simply leave the side-effecting instructions out of the table?

Since you say that you're not applying VN at all to memory loads & stores, how is it possible for these side-effecting operations (including calls) to affect whether two *unrelated* values should be considered equal? Consider this code:

```
a: int = add b c;
store p q;
d: int = add b c;
```

It sounds like you're saying that, when scanning this code, you'd clear the value-number table when you reach the `store`. This means that `a` and `d` won't be considered equal and subject to redundancy elimination, even though they both hold the value `b+c`. Is that in fact what this paragraph is saying? If so, can you explain a little more about why this is necessary?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+## 2. Expression Canonicalization
+
+To maximize redundancy detection, expressions are canonicalized before lookup:
+
+* Commutative operations (e.g., `add`, `mul`) sort operands by value number.
+* Relational operators are normalized (e.g., `gt(a,b)` is rewritten as `lt(b,a)`), reducing equivalent comparison forms.
+* Types are included in expression keys to prevent cross-type equivalences.
```
sampsyo, line 38: What kinds of "cross-type equivalences" do you have in mind? In Bril, for example, `add` and `fadd` are separate opcodes.

Also, it's a little odd that this appears in a list of about redundancy and canonicalization. This seems like a different issue?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+Constants are treated as expressions keyed by their literal value and type. Copy (`id`) instructions are not treated as expressions; instead, they are handled separately by copy propagation and elimination passes.
+
+## 3. Phi Nodes and Loops
+
+We represent control-flow merges using **explicit SSA phi nodes**, rather than Bril’s alternative `get`/`set`-based encoding. This choice aligns more directly with the presentation in the Cooper–Briggs GVN literature and allows phi nodes to participate naturally in value numbering.
```
sampsyo, line 44: I think I had a comment about this before, but can you say more about why `get` and `set` cannot "participate naturally in value numbering"? I don't quite see why `phi` has an advantage here.

Also in a different comment that has now disappeared, I think I mentioned that this is the "old SSA" Bril extension, which you may want to link to.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+Phi nodes are processed before other instructions in each block, consistent with SSA semantics. Our GVN pass handles three important cases:
+
+1. **Meaningless phis**: If all phi arguments have the same value number, the phi is redundant and the destination is assigned that VN.
+2. **Redundant phis**: If two phis have identical argument VN patterns, they receive the same VN.
+3. **Loop header phis**: If a block has an incoming backedge, we conservatively assign fresh VNs to its phi nodes. This avoids unsound reasoning about loop-carried dependencies, at the cost of missing some equalities that partition-based algorithms could discover.
```
sampsyo, line 50: I also don't understand why this is necessary (why a back-edge-having phi is any different from any other phi, and under what circumstances this could lead to "unsound reasoning about loop-carried dependencies"), but maybe that is just too detailed to get into?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+Using real phi nodes simplifies reasoning about dominance and equivalence, compared to the `get`/`set` framework, which obscures merge semantics behind memory-like operations and makes value numbering significantly harder to express cleanly.
```
sampsyo, line 52: Here's a place where some unsubstantiated claims remain in this version of your writeup. There is no evidence here to support your "harder to express cleanly" claim. I recommend either substantiating or deleting this and any other statement that does not come with justification.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+This design choice mirrors the limitations of single-pass, hash-based GVN in cyclic SSA regions and reflects a deliberate tradeoff between simplicity and optimization power.
```
sampsyo, line 54: I don't quite get what you're saying here about "mirror[ing] the limitations." Isn't this an *implementation* of single-pass, hash-based GVN? So in what sense are we "mirroring" anything?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+## 4. SSA Construction and Its Costs
+
+GVN operates most naturally on SSA form, but converting Bril programs to SSA and lowering them back introduces overhead—primarily in the form of phi nodes and copy instructions. A naïve (minimal) SSA construction placed phi nodes at all dominance frontiers, leading to extreme code bloat and negating GVN’s benefits.
+
+To address this, we implemented **pruned SSA**, which places phi nodes only for variables that are live at join points. This required a standard backward liveness analysis over the control-flow graph. In practice, pruned SSA reduced the number of phi nodes by over 80% on average, making SSA-based optimization viable.
+
+## 5. Copy Propagation and Elimination in SSA
+
+After GVN, many redundancies are rewritten as copy instructions. While in SSA form, copy propagation is always safe: each variable has exactly one definition, and every definition dominates all of its uses. We therefore aggressively propagate copies—including into phi arguments—before lowering SSA.
```
sampsyo, line 64: Can you say a little more about what it looks like to "propagate copies […] into phi arguments"? What phase of the algorithm are we talking about? I guess I just don't quite understand the "into phi arguments" shorthand here.

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+Fixing this logic was essential for correctness and avoiding infinite loops during lowering.
+
+## 7. Copy Coalescing and SSA Overhead Recovery
+
+Even with careful SSA construction and copy elimination, phi lowering introduces many copies—especially at loop entry and backedges. These copies were responsible for many benchmarks becoming worse after “optimization.”
+
+To address this, we implemented **copy coalescing**:
+
+* A simple local pass handles cases where a value is copied immediately after its definition and never used again.
```
sampsyo, line 86: Are you talking about code like this, within a single basic block?

```
x: int = ...;
...
y: int = id y;
[x is dead]
```

If so, why wasn't this caught by your earlier copy elimination pass, described under section 5?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+* A global pass uses instruction-level liveness analysis to detect non-interfering live ranges and safely merge variable names.
```
sampsyo, line 87: Because the goal is to "merge variable names," it sounds like this is doing something akin to register allocation: i.e., using fewer variables when possible because two different variables have disjoint live ranges. I don't see how that relates to copies—can you explain a tiny bit more what the idea is here?

content/blog/2025-12-16-gvn-final-report.md
```diff
@@ -0,0 +1,212 @@
...
+
+We do **not** implement the paper’s partition-based (AWZ-style) algorithm, which can reason more precisely about cyclic SSA regions. We also do not implement AVAIL-based removal or PRE-style code motion. Instead, we rely on copy propagation and coalescing to realize most eliminations.
+
+Compared to the paper, our design is more conservative in two respects: we assign fresh value numbers to loop-header phis, and we treat all effectful operations as global barriers. These choices simplify the implementation and ensure correctness in Bril, at the cost of reduced optimization power in some cases.
+
+## 10. Reflections
+
+Several lessons stood out in this project:
+
+* **GVN is easy; making it pay off is hard.** The core algorithm is compact, but SSA construction and lowering dominate engineering effort.
```
sampsyo, line 160: Hmm… is it really "easy"? I guess I wouldn't have concluded that from your experience, but maybe I'm wrong and you actually did not have any bugs when writing this part.

sampsyo (changes requested):
Thanks for rethinking the writeup here. This version is a lot clearer and I understand your project much better now.

As a consequence, I still have a few technical questions about some of your design decisions. Can you add a little more detail in these places?

PR #640, opened 2025-12-21
[Final Project] Rust Aliasing

### LINKED ISSUE #617: Project Proposal: Evaluating Alias Analysis in Rust
# Project Proposal: Evaluating Alias Analysis in Rust
**What**:
Aliasing is constrained in Rust more so than other languages. This suggests it should be possible to get more information than typical and perform aggressive optimizations based on aliasing rules. This is already the case, with proposed alias models, for example Stacked Borrows or the current state of the art Tree Borrows, permitting reasoning about how references change across function calls using an intraprocedural analysis which would be entirely impossible in a language with weaker conditions on aliasing.

However, due to `unsafe` code, static pointer analyses still have to be fairly conservative. I want to answer three questions:
1. How conservative must current pointer analyses be?
2. Does more precise aliasing information unlock further optimizations?
3. If more precise aliasing information *is* helpful for optimizations, is it possible to define a pointer analysis tailored to Rust's alias model letting it perform better than the state of the art?

**Measuring Effectiveness**:
It's a hard to separate the procedure from the evaluation for (1) and (2) as they are effectively a quantitative evaluation of current technologies. Therefore the sections are grouped together.

I will quantify (1) by rephrasing it to "how many aliases do current alias analyses attribute to a given pointer which can never happen." As stated, this is still likely not a tractable problem because finding all aliases likely will not terminate (though I might as well try it). Therefore I will approximate an answer. To do so, I will create a suite of Rust code taken from the Rust standard library (due to it's liberal use of `unsafe` code) and popular Rust crates. I will then run a current alias analyses on the code, a pass built in the compiler if one exists (I'm unsure if one does) or instead use [rupta](https://github.com/rustanlys/rupta). Using the test suites from the given projects, I will use [miri](https://github.com/rust-lang/miri) to record dynamically what references alias each other. This will likely require hacking miri in some way, however as miri supports checking MIR against the Stacked Borrows model, which requires dynamically recording what pointers aliases what, this new feature should be possible. I can then take the difference between the dynamically recorded number of aliases and the statically approximated number of alias pairs. Reporting these measurements constitutes a success.

(2) is hard to quantify. First, I will have to make a shortlist of very aggressive optimizations which require more precise alias information than found in (1). To create this, it will likely help to categorize the types of aliasing which were missed in (1). In part, this is the purpose of (1). A possible answer to this question is that the missed aliasing information does not unlock any optimizations.

If (2) finds some optimizations might be helpful, one way I propose to improve alias analysis is an implementation of either Stacked Borrows or Tree Borrows which is approximated statically using the results of an alias analysis. The result from this can then be used to perform the optimizations found in (2). These can then be measured for regressions or speedups based on the tests used in (1). It also might be possible to simply use [rustc-perf](https://github.com/rust-lang/rustc-perf). If execution time improvements on these tests are found without major regressions, success!

**How Much of This Do I Expect To Complete?**:
Completing (1)-(3) is *very* ambitious. I expect to complete (1) and start (2) for this course project. I believe this should still be reasonable scope. If I am lucky and make more progress, great! I don't see any ways (1) and (2) can fail terribly, save unexpected technical hurdles, however, they may produce the boring answer "the current state of the art is pretty good" and (3) would not be a useful thing to do. I think (3) is over scoped for the course project, especially as it would have to be done after (1) and (2).

**Team Members**:
Jeremy Ku-Benjet (jk2582)

jku20:
closes #617 
As a final note, most of this project felt like rediscovering stuff already known by the Rust language team. The big idea of comparing pointers ended up feeling pretty meaningless. However reading a lot about Rust aliasing semantics did reveal how limited aliasing turned out to be. It also suggested that a next step could be looking at doing specific analyses which are aware that Rust's `Cell`s (and friends) are the main way safe Rust gets mutable aliasing. At the other side of that reading right now, I'm not sure what that new tech could be, if it even possibly exists at all.

I don't like how much of the post is background, but I kind of wanted it to be approachable without a background using Rust or that much of a compilers background, though I feel like I partially failed at both, especially near the end where I kind of gave up because giving a full explanation of the async/await pin issue seemed like it would take fairly long while for not much payoff, and this is already really really really embarrassingly late.


content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+title = "Musings on Aliasing in Rust for Optimization"
+[extra]
+bio = """
+  Jeremy Ku-Benjet is a MS student at Cornell University.
+"""
+[[extra.authors]]
+name = "Jeremy Ku-Benjet"
++++
+
+# Musings on Aliasing in Rust for Optimization
```
sampsyo: You don't actually need this heading; the template renders your post title for you.

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+## Experience Report
+**Goal**: I planned to evaluate how well static analysis can do in finding aliasing pointers in the context of performance optimization by comparing the results of [Rupta's](https://github.com/rustanlys/rupta) static analysis to logs of Miri running on various Rust projects. Specifically I wanted to compare the number of aliasing pointers found by Rupta at compile time to the number of aliasing pointers found by miri when run on various Rust project's tests. If these numbers were similar, it would suggest Rust is particularly well suited for alias analysis. If they are significantly different, it would suggest either alias analysis is hard in Rust, more similar to C for example, or more effort could be spent to leverage Rust's aliasing guarentees.
```
sampsyo: Please do a little `s/miri/Miri/g` for style consistency.

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+
+**Success?**: Being upfront, this project failed. After seeing the discrepency in these metrics, I found their comparison was not meaningful. At best, they measured the same thing, serving as proxies for the amount of aliasing in a given Rust project. However, I don't know how to make a convicing argument for this as it's possible for a project with many short immutable lived aliases pairs (the kind not breaking optimizations) look similar by these metrics to a project with long lived aliases between many pointers. A proper metric for measuring aliasing in the context of optimization would require some way to know if the aliases detected were important towards optimizations. That is, it would require using the aliases found by Rupta to fuel an optimization and comparing that to a similar optimization fueled by a profiling run of Miri.
```
sampsyo: I think something syntactic may have gone wrong in the phrase "short immutable lived aliases pairs"

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+let _ = &shared_ref1; // Some valid reads
+let _ = &shared_ref2;
+let _ = &shared_ref1;
+*shared_ref1 = 2;     // Rejected as shared reference `shared_ref1` cannot be mutated.
+
+// Mutable references
+let x = 1;
+let mut_ref1 = &mut x;
+let _mut_ref2 = &mut x;
+let _ *mut_ref1;      // Rejected as `x` was borrowed a second time making _mut_ref2.
```
sampsyo:
```suggestion
let _ = *mut_ref1;      // Rejected as `x` was borrowed a second time making _mut_ref2.
```

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+define i32 @foo(ptr noalias %0, ptr noalias %1) #0 {
+  store i32 4120, ptr %0, align 4
+  store i32 6120, ptr %1, align 4
+  ret i32 4120
+}
+```
+This is the same LLVM output that the C code compiles to when using `restrict` (with an added `noalias` on `y` because it is also an `&mut`). It turns out this reasoning can be extrapolated, letting [most reference be marked as](https://github.com/rust-lang/rust/blob/cb79c42008b970269f6a06b257e5f04b93f24d03/compiler/rustc_ty_utils/src/abi.rs#L273) `noalias`. Though, a better way of saying this is the langauge's aliasing rules make sure not to preclude letting these optimizations occur.
+
+## When Mutable Aliasing Occurs
+Despite it being nice to not have to worry about mutable aliases, there are cases in which these guarentees cannot be easily applied. The most simple example is Rust's raw pointers. Raw pointers must exist, for example when doing FFI with C. To work raw pointers have to be C-like pointers, able to alias and be untracked by the compiler as `rustc` has no way to track what goes on in `C` code:
```
sampsyo:
```suggestion
Despite it being nice to not have to worry about mutable aliases, there are cases in which these guarentees cannot be easily applied. The most simple example is Rust's raw pointers. Raw pointers must exist, for example when doing FFI with C. To work raw pointers have to be C-like pointers, able to alias and be untracked by the compiler as `rustc` has no way to track what goes on in C code:
```

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+```
+The compiler does not reject the above even though it is effectively the same thing as what happened using just mutable references as there is no guarentee on `*mut` not aliasing other `*mut`. One notable thing which pops up in this example is an [unsafe block](https://doc.rust-lang.org/nomicon/what-unsafe-does.html). This makes sense, as using raw pointers it becomes very easy to invoke undefined behavior. Consider the following code passing aliasing mutable references into the `foo` from above:
+```rust
+let x = 1;
+let r1 = &mut x as *mut i32;
+unsafe {
+  foo(&mut *r1, &mut x);
+}
+```
+`r1` and `&mut x` alias so making mutable references from them creates aliasing mutable references. This leads to undefined behavior, manifesting in an incorrect result when when calling `foo` as it gets optimized to `foo_opt` because it's args are `noalias`. Another way too look at this is `rustc` cannot attach `noalias` to the args of  `bar(x: *mut i32, y: *mut i32)` in the emitted llvm.
```
sampsyo:
```suggestion
`r1` and `&mut x` alias so making mutable references from them creates aliasing mutable references. This leads to undefined behavior, manifesting in an incorrect result when when calling `foo` as it gets optimized to `foo_opt` because it's args are `noalias`. Another way too look at this is `rustc` cannot attach `noalias` to the args of  `bar(x: *mut i32, y: *mut i32)` in the emitted LLVM.
```

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+```rust
+let x: &UnsafeCell<i32> = &4120.into();
+let r1 = x.get();
+let r2 = x.get();
+// r1 and r2 are aliasing *mut i32s
+```
+The super power of `UnsafeCell` in the above is it is perfectly defined behavior to call `x.get()` and use the mutable aliases to shared memory it returns. However, with mutable alias to its memory, `&UnsafeCell` looses the ability to be treated like a normal shared reference and the compiler builds in special support for it. As with the above raw pointers, one way this manefest is when `&UnsafeCell` (or any of it's derivatives like `Cell` or `RefCell`) is used as a function arg, it cannot be annotated with `noalias` when compiled to LLVM.
```
sampsyo:
```suggestion
The super power of `UnsafeCell` in the above is it is perfectly defined behavior to call `x.get()` and use the mutable aliases to shared memory it returns. However, with mutable alias to its memory, `&UnsafeCell` loses the ability to be treated like a normal shared reference and the compiler builds in special support for it. As with the above raw pointers, one way this manifests is when `&UnsafeCell` (or any of its derivatives like `Cell` or `RefCell`) is used as a function arg, it cannot be annotated with `noalias` when compiled to LLVM.
```

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+
+The interesting thing about `Cell`s and especially `RefCell`s is they end up forcing the compiler to treat them as possibly having mutable aliases despite them being safe types. Looking at [Rust's abi code (same link as above)](https://github.com/rust-lang/rust/blob/cb79c42008b970269f6a06b257e5f04b93f24d03/compiler/rustc_ty_utils/src/abi.rs#L273), there are only two other cases in which `noalias` doesn't annotate function args:
```
sampsyo:
```suggestion
The interesting thing about `Cell`s and especially `RefCell`s is they end up forcing the compiler to treat them as possibly having mutable aliases despite them being safe types. Looking at [Rust's ABI code (same link as above)](https://github.com/rust-lang/rust/blob/cb79c42008b970269f6a06b257e5f04b93f24d03/compiler/rustc_ty_utils/src/abi.rs#L273), there are only two other cases in which `noalias` doesn't annotate function args:
```

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+
+Looking at these three case, there are limited ways mutable aliases can make their way into safe Rust code. Considering it's (in my experience so far) rare to change `Box`s allocator and the interaction with async/await seems [more like a bug being worked on](https://github.com/rust-lang/rust/issues/125735), these rules lets the programmer limit their view of mutable aliases to thinking about `Cell`s. Their effectively "choke points" where mutable aliases hide. They `Cell`s exist, but anecdotally, I find them to be a vast minority of pointers. So, I find it cool to see the reasoning and potential performance penalty constained so much.
```
sampsyo:
```suggestion
Looking at these three case, there are limited ways mutable aliases can make their way into safe Rust code. Considering it's (in my experience so far) rare to change `Box`'s allocator and the interaction with async/await seems [more like a bug being worked on](https://github.com/rust-lang/rust/issues/125735), these rules lets the programmer limit their view of mutable aliases to thinking about `Cell`s. They're effectively "choke points" where mutable aliases hide. They `Cell`s exist, but anecdotally, I find them to be a vast minority of pointers. So, I find it cool to see the reasoning and potential performance penalty constained so much. 
```

sampsyo: I think "tiny minority" would sound more normal than "vast minority"?

content/blog/2025-12-19-musings-on-aliasing-in-rust.md
```diff
@@ -0,0 +1,150 @@
...
+
+[^1]: I'm taking this example from the [Stacked Borrows paper](https://dl.acm.org/doi/10.1145/3371109) which uses it to show a similar thing, though with different exposition.
+[^2]: An interesting corrolary of this is that at compile time it's possible for pointers of the same type to have to be treated differently depending on how they were created. In other words, pointers have some extra data attached to them the compiler has to keep track of. This is sometimes called *provenance*. Ralf Jung has [two good](https://www.ralfj.de/blog/2018/07/24/pointers-and-bytes.html) [articles on this](https://www.ralfj.de/blog/2020/12/14/provenance.html) arguing for it's existance. In some languages, for example Rust, provenance is in this interesting position where [it is not yet fully specified, but it still has to be reasoned about](https://doc.rust-lang.org/std/ptr/index.html#provenance).
```
sampsyo:
```suggestion
[^2]: An interesting corrolary of this is that at compile time it's possible for pointers of the same type to have to be treated differently depending on how they were created. In other words, pointers have some extra data attached to them the compiler has to keep track of. This is sometimes called *provenance*. Ralf Jung has [two good](https://www.ralfj.de/blog/2018/07/24/pointers-and-bytes.html) [articles on this](https://www.ralfj.de/blog/2020/12/14/provenance.html) arguing for its existance. In some languages, for example Rust, provenance is in this interesting position where [it is not yet fully specified, but it still has to be reasoned about](https://doc.rust-lang.org/std/ptr/index.html#provenance).
```

sampsyo (changes requested):
Nice work overall here, @jku20. Even if this turned out to be a "negative result" in the end, you've done a great job assembling a guide to issues re. aliasing in Rust and the corner cases where they can arise—this does clarify the original set of questions we had about whether there's any opportunity for fancier aliasing information in rustc.

I have a few minor writing suggestions. When you have a chance to address them, can you please "request a review" in GitHub's interface so I know to take a look and merge?

jku20:
Cool, will do when I get the time and computer access (probably some time in January).

As another question, would you mind if I eventually post an edited and possibly slightly extend version of this to my personal blog?

sampsyo:
Great!

Yes, 100% totally cool to mirror this on your own blog (or do anything else with it).

jku20:
I made some changes based on the comments and fixed some more awkward sentences and typos. The zola build is currently failing on the CI, but I'm not sure it has to do with the edits I made. It seems to be an issue with the formatting `config.toml` in the project's root.

